PR #578, opened 2026-04-13
fix(web): disable resume and show agent selector when profile is deleted

luancm:
Users couldn't recover from sessions that used a deleted agent profile — Resume would silently fail and the New Session dialog would submit with the deleted profile ID. Resume is now disabled with a tooltip, and the dialog falls back to a valid profile with the selector forced visible.

## Important Changes

- `FailedSessionBanner`: reads `profileExists` from store; disables Resume + shows tooltip when profile is gone
- `NewSessionDialog`: separates `sessionProfileId` (original, for missing-check) from `effectiveDefaultProfileId` (valid fallback, for state init); adds `initialProfileId` prop to `NewSessionForm`
- `NewSessionForm`: shows `AgentSelector` when `isDefaultProfileMissing`; shows inline message + disables submit when no profiles remain

## Validation

```
cd apps && pnpm --filter @kandev/web e2e -- tests/session/new-session-deleted-profile.spec.ts
```
2 tests passing.

## Screenshots

**Resume button — before / after profile deleted**

Before
before_resume

After
after_resume

**New Agent dialog — before / after profile deleted**

Before
before_new_agent

After
after_new_agent

## Checklist

### DIFF
--- a/apps/backend/internal/agent/settings/store/sqlite.go
+++ b/apps/backend/internal/agent/settings/store/sqlite.go
@@ -114,7 +114,7 @@ func (r *sqliteRepository) initSchema() error {
 }
 
 // migrateDropModelCheckConstraint recreates agent_profiles without the legacy
-// CHECK(model != '') constraint. Existing databases created before the ACP-first
+// CHECK(model != ”) constraint. Existing databases created before the ACP-first
 // migration carry this constraint, which prevents empty model values. New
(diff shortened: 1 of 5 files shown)


luancm:
Sorry for the force pushes, had some issues with different identities in different machines

carlosflorencio:
Nice! Triggered the CI.

PR #583, opened 2026-04-14
fix(web): stop vertical scroll on mobile column tabs

xsu1010:
On mobile, the column tabs allow for a vertical scroll when there's only a horizontal scroll.
This PR prevents the vertical scroll from happening.

## Screenshots

(Scrolling down so that the effect is visible)

| Before: | After: |
|--------|-------|
|image | image |

## Checklist


zeval:
Thank you for your contribution! 👍

PR #614, opened 2026-04-17
fix(web): validate activeSessionId belongs to activeTaskId before use

jcfs:
When navigating to a task without sessions (created with `start_agent=false`), the chat panel incorrectly displayed messages from the previously viewed task's session. This occurred because `activeSessionId` was used as a fallback without validating it belonged to the current `activeTaskId`.

## Important Changes
- Added validation in `useSessionState` hook to check `activeSessionId.task_id === activeTaskId` before using it
- Added validation in `useTaskPageData` when computing `effectiveSessionId` for the same reason

## Validation
- `make fmt` — passed
- `pnpm e2e --grep "Session isolation"` — 2 tests passed
- Full E2E suite — 302 passed (5 failures are pre-existing, unrelated flaky tests)

## Checklist

- [ ] Self-reviewed
- [ ] Tests pass locally
- [ ] No unrelated changes

### DIFF
--- a/apps/web/components/task/task-page-content.tsx
+++ b/apps/web/components/task/task-page-content.tsx
@@ -536,12 +536,20 @@ function useTaskPageData(
   const setActiveSession = useAppStore((state) => state.setActiveSession);
   const setActiveTask = useAppStore((state) => state.setActiveTask);
 
+  // Validate that activeSessionId belongs to activeTaskId to prevent showing
+  // messages from an unrelated session when navigating to a task without sessions.
+  const validatedActiveSessionId = useAppStore((state) => {
+    if (!activeSessionId || !activeTaskId) return null;
+    const session = state.taskSessions.items[activeSessionId];
+    return session?.task_id === activeTaskId ? activeSessionId : null;
+  });
+
   const { task } = useTaskDetails(activeTaskId, initialTask);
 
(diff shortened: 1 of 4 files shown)


jcfs:
Thanks for the thorough review! Here's how I addressed each point:

## Blocker 1: Missing unit test ✅ Fixed
Added `use-session-state.test.ts` with 8 test cases covering:
- `sessionId` supplied directly → uses it without validation
- `sessionId = null`, matching task → uses `activeSessionId`
- `sessionId = null`, mismatched task → returns `null`
- `sessionId = null`, session not yet in store → returns `null`
- Session flag derivation (isStarting, isAgentBusy, isFailed)

## Suggestion 1: Duplicated validation selector pattern — Skipped
While the duplication exists, it's only ~3 lines in each place. Extracting a shared selector adds indirection for a pattern that currently exists in exactly 2 places. If this pattern spreads further, I agree it would be worth extracting.

## Suggestion 2: Magic string in E2E tests ✅ Fixed
Extracted to `const SIMPLE_MOCK_RESPONSE` at module scope.

## Suggestion 3: Three separate useAppStore subscriptions — Skipped
The overhead of 3 subscriptions vs 1 is negligible in practice (Zustand batches updates). The current structure is more readable with each selector doing one thing.

PR #640, opened 2026-04-18
fix(web): use UUID fallback for attachments on non-secure contexts

zeval:
## Summary

Self-hosted instances served over plain HTTP (e.g. Tailscale IPs) crashed when pasting or picking chat attachments because `crypto.randomUUID()` is only exposed in secure contexts. Route the 3 remaining callsites through the existing `generateUUID()` helper so the UI degrades gracefully on HTTP.

## Validation

- `pnpm --filter @kandev/web test` — 302 pass, incl. 4 new `generateUUID` cases and 2 new `processFile` insecure-context cases
- `pnpm exec eslint` on changed files — clean
- `prettier` — clean

## Checklist

- [ ] Tests added/updated
- [ ] Docs updated (if needed)
- [ ] Verified locally

### DIFF
--- a/apps/web/components/task/chat/file-attachment.test.ts
+++ b/apps/web/components/task/chat/file-attachment.test.ts
@@ -1,0 +1,61 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { processFile } from "./file-attachment";
+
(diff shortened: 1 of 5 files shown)


zeval:
Thanks for the review, @coderabbitai.

Skipping the `FakeImage` `naturalWidth`/`naturalHeight` nitpick — the stub is file-local, used only by `processFile`, which doesn't read those fields. Adding defaults now would be speculative: `processImageFile` in `image-attachment-preview.tsx` has its own testing surface, and if this file ever imports that function the test author can extend the stub then. Prefer to keep the stub minimal to what the code under test actually touches.

PR #667, opened 2026-04-20
fix: remove duplicate WebSocket event subscriptions

jcfs:
subscribeEventBusHandlers() registered handlers for message.added, message.updated, task_session.state_changed, and github.task_pr.updated -- but TaskEventBroadcaster already subscribes to all four. This caused every message event to produce two WebSocket broadcasts, and the legacy handlers also stripped fields (turn_id, raw_content, updated_at) that the frontend expects.

## Important Changes

- Removed subscribeEventBusHandlers() call from main.go and the function itself
- Removed 4 legacy handler factory functions from helpers.go: newMessageAddedHandler, newMessageUpdatedHandler, newSessionStateChangedHandler, newGitHubTaskPRUpdatedHandler
- Added TestTaskEventBroadcaster_NoDuplicateSubscriptions to guard against re-introducing duplicate subscriptions
- Added TestTaskEventBroadcaster_PreservesAllFields to verify fields like turn_id and raw_content are forwarded

## Validation

- go build ./cmd/kandev/ -- clean
- go test ./internal/gateway/websocket/ ./cmd/kandev/ -- all pass
- go vet ./cmd/kandev/... ./internal/gateway/websocket/... -- clean
- make -C apps/backend fmt -- clean

## Possible Improvements

One behavioral difference: the old newSessionStateChangedHandler used BroadcastToSession (session-scoped) while TaskEventBroadcaster uses Broadcast (global) -- this is intentional per the existing comment in task_notifications.go to support sidebar state tracking across all sessions.

## Checklist

- [ ] Self-review of changed files
- [ ] Tests added or updated
- [ ] Documentation updated (if applicable)
- [ ] No unrelated changes included


jcfs:
Fixed — rewrote the test with two layers of defense:

1. Subscription count guard: len(b.subscriptions) == 46 catches any duplicate b.subscribe() call within RegisterTaskNotifications itself.
2. Per-event sub-tests: each of the 4 previously-duplicated events gets a fresh event bus + broadcaster, publishes once, and verifies the counter fires exactly once.

Also addressed the other comments: removed unused all(), recorder, and simplified broadcastRecorder to a plain counter.

PR #704, opened 2026-04-25
fix(web): show skeleton in file tree header while workspace path loads

carlosflorencio:
The file tree header sometimes rendered with no workspace folder name — only the copy button — because `displayPath` is derived from `session.worktree_path`/`repository.local_path` while the header itself is gated on tree-load state, so the two could be out of sync. Now an empty `displayPath` renders a small skeleton so the layout stays stable and the loading state is visible.

## Validation

- `make -C apps/backend fmt test lint` (clean)
- `pnpm format` / `pnpm exec tsc --noEmit` / `pnpm exec eslint --max-warnings 0` in `apps/web` (clean)

## Possible Improvements

Low risk — purely a UX patch. The underlying hydration mismatch (tree loaded before session/repository slices populate) still exists; a follow-up could investigate why `useRepository` lags behind the tree.

## Checklist

### DIFF
--- a/apps/web/components/task/file-browser-toolbar.tsx
+++ b/apps/web/components/task/file-browser-toolbar.tsx
@@ -11,6 +11,7 @@ import {
   IconPlus,
 } from "@tabler/icons-react";
 import { Tooltip, TooltipContent, TooltipTrigger } from "@kandev/ui/tooltip";
+import { Skeleton } from "@kandev/ui/skeleton";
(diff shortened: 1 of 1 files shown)


carlosflorencio:
Thanks @claude — applied the suggestion: narrowed the skeleton to `w-24` (96 px) to better match typical short paths and reduce layout pop on resolve.

PR #714, opened 2026-04-27
fix(web): scope settings workflow list to current workspace

carlosflorencio:
Workflows from a previously-visited workspace leaked into a new workspace's settings page because the global Zustand store was never filtered by workspace, and "deleting" a leaked entry issued a real DELETE against the original workspace's workflow — wiping it from both. The settings hook now scopes store reads to the current workspace, so each workspace shows only its own workflows.

## Validation

- `pnpm --filter @kandev/web test --run` (622/622 passing, includes new `use-workflow-settings.test.ts`)
- `pnpm --filter @kandev/web lint` (0 warnings)
- `tsc --noEmit` (clean)

## Possible Improvements

Low risk. The hook still falls back to the unscoped store when no `workspaceId` is provided, preserving behaviour for any other (non-settings) caller. A follow-up could clear `workflows.items` on workspace switch to harden other consumers.

## Checklist

### DIFF
--- a/apps/web/app/settings/workspace/workspace-workflows-client.tsx
+++ b/apps/web/app/settings/workspace/workspace-workflows-client.tsx
@@ -504,7 +504,7 @@ function useWorkspaceWorkflowsPage(
   const router = useRouter();
   const { toast } = useToast();
   const { workflowItems, setWorkflowItems, setSavedWorkflowItems, isWorkflowDirty } =
-    useWorkflowSettings(workflows);
+    useWorkflowSettings(workflows, workspace?.id);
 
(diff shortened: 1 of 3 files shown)


carlosflorencio:
Thanks for the reviews — both suggestions addressed in efc5e039:

- **@claude**: added a test covering the `sw.name !== w.name` branch through `scopedStoreWorkflows` (rename within current workspace).
- **@greptile-apps**: added a test covering the `undefined → defined` workspaceId transition (unscoped fallback first, then scoped after rerender drops the foreign workflow).

Test count is now 624 (+2).

PR #729, opened 2026-04-28
fix(web): self-recover dockview layout from corrupt persisted state

carlosflorencio:
Switching to a task with no active session left the dockview in a corrupted state — panel widths collapsed and the broken layout was persisted to sessionStorage, so the only escape was clearing browser storage. The layout is now defensively validated on every restore path and the sessionless task switch reliably resets to a healthy default.

## Important Changes

- New `isLayoutShapeHealthy` shared validator rejects serialized layouts with non-positive grid/leaf sizes or empty leaf views, applied on both the initial-mount restore and the session-switch slow path.
- New `releaseLayoutToDefault` store action and `finalizeNoSessionSelect` helper handle the sessionless transition: release the outgoing session's portals, rebuild the default layout, then update active task and URL.
- `handleSelectTask` now finalizes a sessionless selection when `prepareAndSwitchTask` returns no session, instead of leaving the dockview wired to the previous session.

## Validation

- `cd apps/web && pnpm exec tsc --noEmit`
- `cd apps/web && pnpm lint`
- `cd apps/web && pnpm test` — 648/648 pass
- `cd apps && pnpm --filter @kandev/web exec playwright test e2e/tests/task/sessionless-task-switch.spec.ts` — 2/2 pass
- Prettier check on all touched files

## Possible Improvements

Low risk — defensive validation is conservative (drops corrupt blobs, builds default). If the validator is too strict it would just rebuild a default layout, never crash.

## Checklist

### DIFF
--- a/apps/web/components/task/dockview-layout-restore.test.ts
+++ b/apps/web/components/task/dockview-layout-restore.test.ts
@@ -1,0 +1,140 @@
+import { describe, it, expect } from "vitest";
+import { sanitizeLayout } from "./dockview-layout-restore";
+
+const VALID_COMPONENTS = new Set<string>(["chat", "files", "shell", "git"]);
+
+/**
+ * Build a minimal valid SerializedDockview-shaped object — matches what
+ * dockview's api.toJSON() produces for a 3-column layout (sidebar | center | right).
+ */
+function buildLayout(opts?: { centerSize?: number; sidebarSize?: number; rightSize?: number }) {
+  return {
+    grid: {
+      root: {
+        type: "branch" as const,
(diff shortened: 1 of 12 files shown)


carlosflorencio:
Addressed Claude's suggestion in the next commit: replaced the unhealthy `{ grid: {} }` mock in the "saved layout matches" test with `makeHealthyLayoutWith({})` so `getHealthySessionLayout` returns truthy and the `savedLayoutMatchesLive` fast-path branch is actually exercised again.

Greptile had no actionable findings (5/5). CodeRabbit was rate-limited and didn't review.

PR #791, opened 2026-05-03
refactor(web): drop dead taskPRs loading state and unused removeTaskPR

jcfs:
The `taskPRs.loaded`/`loading` fields on the github slice were written but never read — `useWorkspacePRs` uses its own `fetchedRef` for fetch-once gating, so the slice flags were vestigial. The unused `removeTaskPR` action was also a footgun for multi-repo tasks: it deleted the entire per-task PR array, which would have wiped every repo's PR state at once.

## Validation

- `pnpm --filter @kandev/web lint` — passes
- `pnpm --filter @kandev/web test` — 1007 tests pass
- `pnpm exec tsc --noEmit` in `apps/web` — clean (only pre-existing generated-file import noise)

## Checklist


jcfs:
Good catch — applied the fix. Switched `taskPRs` to plain `deepMerge` since there's no longer any loading state to preserve.

PR #816, opened 2026-05-04
fix(changes-panel): scope commits/diff to live branch divergence; flatten single-repo

jcfs:
The COMMITS panel was surfacing commits and file changes that were already on main — sometimes 100+ on a branch with one real commit — because both the commit list and the cumulative diff anchored to stale local refs (or a frozen session-start SHA). Now both paths anchor to `merge-base HEAD origin/<base_branch>` so the panel reflects live divergence with the upstream, matching GitHub's three-dot PR diff semantics.

## Important Changes

- agentctl computes merge-base against `origin/<target_branch>` first, with a local-ref fallback for unfetched / locally-only branches; threaded through both the commit log and the cumulative-diff endpoint + WS handler.
- `git log` now passes `--first-parent` so commits brought in via merges from main don't surface as branch work.
- COMMITS panel: single-repo workspaces drop the redundant per-repo sub-header (Commits / Unstaged / Staged); per-repo Push / PR / Stage / Commit buttons lift into the section header. Multi-repo unchanged.

## Validation

- \`make -C apps/backend fmt lint\`
- \`go test ./...\` (full backend suite, including new tests below)
- \`pnpm --filter @kandev/web exec vitest run\` (1050 tests pass)
- \`pnpm --filter @kandev/web exec eslint --max-warnings=0 .\`
- New tests:
  - \`TestGetLog_FirstParentSkipsMergedInCommits\` — verifies merged-in commits are excluded
  - \`TestGetLog_StaleLocalBranchScenario\` — reproduces the stale-local-main shape at the GitOperator level
  - \`TestComputeMergeBase_PrefersOriginOverStaleLocalBranch\` / \`…FallsBackToLocalWhenRemoteMissing\` — API layer
  - \`TestRunGitCumulativeDiffForRepo_PrefersOriginRef\` — same coverage for the file-diff path

## Possible Improvements

Low risk. The \`origin/<base>\` ref relies on the worktree having been fetched at some point — the local-ref fallback covers the truly-unfetched case, but the result will be slightly behind real-time origin/main between agent ops. Acceptable trade-off; matches GitHub PR behaviour.

## Checklist

### DIFF
--- a/apps/backend/cmd/kandev/adapters.go
+++ b/apps/backend/cmd/kandev/adapters.go
@@ -406,7 +406,10 @@ func (a *lifecycleAdapter) GetCumulativeDiff(ctx context.Context, sessionID, bas
 	if agentClient == nil {
 		return nil, nil
 	}
-	return agentClient.GetCumulativeDiff(ctx, baseCommit)
+	// Orchestrator-side cumulative diffs (snapshot tracking) anchor to the
+	// caller-provided base SHA — no dynamic merge-base recomputation. The
+	// live panel uses the WS-handler path, which passes targetBranch.
+	return agentClient.GetCumulativeDiff(ctx, baseCommit, "")
 }
 
 // GetGitStatus retrieves the current git status for a session.
--- a/apps/backend/internal/agent/handlers/git_handlers.go
+++ b/apps/backend/internal/agent/handlers/git_handlers.go
@@ -782,10 +782,15 @@ func (h *GitHandlers) wsCumulativeDiff(ctx context.Context, msg *ws.Message) (*w
 		})
 	}
 
-	// Look up base commit SHA from the session metadata
-	var baseCommit string
+	// Look up base commit SHA and target branch from the session metadata.
+	// targetBranch lets agentctl recompute the base via merge-base against
+	// origin/<branch> for live divergence — same anchoring as the COMMITS
+	// panel, so the file diff doesn't include changes that came in via merges
+	// from main after the session was started.
+	var baseCommit, targetBranch string
 	if h.sessionReader != nil {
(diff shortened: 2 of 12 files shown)


jcfs:
Addressed review feedback in db35dca9:

- **Greptile P1 (`--first-parent` global)**: scoped to the `baseCommit != ""` branch in `GetLog`. Open-ended history calls now keep the full graph; only the divergence-range case follows first-parent. Added `TestGetLog_NoBaseReturnsFullGraph` to guard the regression.
- **Greptile P2 (silent fallback)**: split the conditional in `runGitCumulativeDiffForRepo` so the empty-merge-base case logs a warning before falling back to the stored base.
- **Claude (misleading test name)**: renamed `TestRunGitCumulativeDiffForRepo_PrefersOriginRef` → `TestComputeMergeBase_CorrectAnchorForCumulativeDiff` and updated the doc-comment to reflect what it actually proves (the shared helper, not the function name in the old title).

Skipped Greptile's note on the two slightly-different `isSingleRepo` expressions — they're equivalent in practice (`groupByRepositoryName` collapses falsy names into a single empty-key group), and the file-list path checks `files` directly rather than re-grouping. Not worth churning.

jcfs:
Round 2 addressed in 5bc06a1f:

- **File size**: dropped the dead `ActionButtonsSection` (defined and exported, no consumers anywhere — confirmed via repo-wide grep). File goes from 640 → 516 lines. Cleaner than extracting since the code was unreachable.
- **Three `isSingleRepo` expressions**: extracted `isSingleRepoGroup` into `lib/group-by-repo.ts` and use it from all three sites (`CommitsSection`, `FileListBody`, `FileListSection`). `FileListSection` now also computes `groups` via the helper rather than checking files directly.

PR #819, opened 2026-05-04
feat: attribute cross-task agent messages with a sender badge

zeval:
## Summary

When an agent calls `message_task_kandev`, the receiving task previously saw the prompt as if a human had typed it — no way to tell it came from another agent, and no way to jump back to the source. This PR threads sender attribution end-to-end:

- The MCP server (inside agentctl) injects `sender_task_id` and `sender_session_id` from its own context.
- The backend handler validates the sender (rejects missing IDs, self-messages, unknown tasks), wraps the prompt in a `<kandev-system>` attribution block (so the receiving agent identifies the peer live and on ACP resume), and persists structured metadata on the recorded user message.
- The web UI renders a clickable purple badge above the message bubble with the sender task title; clicking navigates to the source task. Title live-resolves from the kanban store and falls back to the snapshot when the task isn't loaded.

Reuses the existing `<kandev-system>` strip pipeline, so the UI bubble shows only the original prompt body — no new strip code needed.

Plan: `docs/specs/...` → grilled to 14 locked decisions before writing any code.

### Agent 1
image

### Agent 2
image

## Test plan
- [x] Backend: `go test ./...` (full suite, 0 failures); new tests cover `WithSenderTask`, `wrapAgentMessage` (incl. nested `<kandev-system>` collision), and `handleMessageTask` validation paths
- [x] Backend: `make lint` clean
- [x] Web: `pnpm --filter @kandev/web test` (1056 tests pass, 6 new badge tests)
- [x] Web: `pnpm --filter @kandev/web lint` clean
- [x] E2E: 8/8 Playwright tests pass — running/idle target paths, self-message rejection, unknown-target error, badge link target, sender rename live-resolution, `<kandev-system>` wrapper preservation in `raw_content`, and prompt bodies containing literal `<kandev-system>` tags

### DIFF
--- a/apps/backend/internal/mcp/handlers/get_task_conversation_test.go
+++ b/apps/backend/internal/mcp/handlers/get_task_conversation_test.go
@@ -22,7 +22,7 @@ func TestHandleGetTaskConversation_MissingTaskID(t *testing.T) {
 
 func TestHandleGetTaskConversation_UsesPrimarySession(t *testing.T) {
 	svc, repo := newTestTaskService(t)
-	task, sess := seedTaskWithSession(t, svc, repo, models.TaskSessionStateWaitingForInput)
+	_, task, sess := seedTaskWithSession(t, svc, repo, models.TaskSessionStateWaitingForInput)
 
 	_, err := svc.CreateMessage(context.Background(), &service.CreateMessageRequest{
 		TaskSessionID: sess.ID,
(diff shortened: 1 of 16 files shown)


zeval:
Addressed both P1 findings in 0dbe54ea:

**CodeRabbit — `requeueMessage` drops Metadata** (`event_handlers_agent.go`)
Now calls `QueueMessageWithMetadata` and forwards `queuedMsg.Metadata`, so sender attribution survives transient retries.

**Greptile — orphaned attributed message in CREATED branch** (`handlers.go`)
Reordered `dispatchTaskMessage`'s CREATED case so `StartCreatedSession` runs *before* `recordUserMessage`. If the launch fails, no row is written.

Plus two nits Greptile flagged:
- Collapsed identical `displayTitle`/`fullTitle` locals in `SenderTaskBadge`.
- Extracted a `waitForTargetIdle` poll helper in the e2e suite, replacing the 1.5s raw sleep.

Backend tests + e2e suite still green (8/8).

zeval:
Addressed Claude's two suggestions in 28e23ea4:

**1. Title injection** (`wrap_agent_message.go:22`)
Strip `</kandev-system>` from the embedded sender title before formatting; the strip regex's non-greedy match would otherwise short-circuit on a malicious title and leak the attribution tail into the chat bubble. Metadata snapshot retains the original title for badge display. Added a regression test (`TestWrapAgentMessage_TitleContainingClosingTag`).

**2. Cross-workspace isolation** (`handlers.go:649`)
Cross-workspace messaging is intentional per the feature plan (Q13: "allow cross-workspace, badge URL handles nav"). Scoping it would be a product-level decision about cross-workspace auth/visibility. Added a comment on the lookup so future readers don't add a scope check by reflex.

PR #849, opened 2026-05-08
fix(web): rank longest/quickest tasks by active duration

jcfs:
The Longest/Quickest Tasks lists on the stats page were sorting by elapsed wall-clock span (first turn → last turn), so a task that was idle for days outranked one that actually consumed more agent time. Switched the sort key to active duration (sum of turn durations) so the ranking reflects real work, matching the primary number already displayed on each row.

## Validation
- `pnpm lint` (apps/web)
- Verified visually in stats page: subtitles now read "Ranked by active duration"; ordering matches the displayed primary duration

## Checklist
- [ ] Tests added/updated
- [ ] Docs updated
- [ ] Self-reviewed


jcfs:
@coderabbitai applied the simplification — `sorted` alias removed, sort runs in place on `filtered`. Thanks!

PR #860, opened 2026-05-10
fix(web): mobile task switcher sheet skeleton while snapshot loads

zeval:
## Summary
- Mobile task page sheet showed "No tasks yet." instead of a loading state when the workflow snapshot hadn't hydrated `state.kanban` (and stayed that way silently on fetch failure).
- `useWorkflowSnapshot` now flips `kanban.isLoading=true` while a fetch for an un-hydrated workflow is in flight and back to `false` on settle (cancel-safe). Errors surface via `console.warn` instead of the prior swallow.
- `useTasks` returns `{ tasks, isLoading }`; sheet's `useSheetData` exposes `tasksLoading` and feeds it into the existing `TaskSwitcher` skeleton.

## Test plan
- [x] `pnpm vitest run hooks/use-tasks.test.ts hooks/use-workflow-snapshot.test.ts` — 11 new tests, all pass
- [x] Full `pnpm --filter @kandev/web test` — 1242 / 1242 pass
- [x] `tsc --noEmit` clean
- [x] `pnpm --filter @kandev/web lint` clean
- [ ] Manual mobile-viewport repro: open `/t/[taskId]` directly, tap hamburger → sheet should skeleton then list, not "No tasks yet."
- [ ] Manual mobile-viewport repro with snapshot endpoint stubbed to 5xx → empty-state copy + console warning, no infinite skeleton

### DIFF
--- a/apps/web/components/task/mobile/session-task-switcher-sheet-hooks.ts
+++ b/apps/web/components/task/mobile/session-task-switcher-sheet-hooks.ts
@@ -87,7 +87,7 @@ export function useSheetData(workspaceId: string | null, workflowId: string | nu
   const gitStatusByEnvId = useAppStore((state) => state.gitStatus.byEnvironmentId);
   const envIdBySessionId = useAppStore((state) => state.environmentIdBySessionId);
   const messagesBySession = useAppStore((state) => state.messages.bySession);
-  const { tasks } = useTasks(workflowId);
+  const { tasks, isLoading: tasksLoading } = useTasks(workflowId);
   const steps = useAppStore((state) => state.kanban.steps);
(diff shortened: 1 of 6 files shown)


zeval:
Addressed Greptile + Claude review feedback in b473839d:

- guarded `useWorkflowSnapshot` finally block with `cancelled` so an old fetch settling after a workflowId switch can't collapse the new effect's skeleton.
- dropped the redundant `kanbanIsLoading` subscription in `useSheetData` and now forward `tasksLoading` directly.
- fixed the `setState` mock in `use-tasks.test.ts` to actually mutate state (matches `use-workflow-snapshot.test.ts`), and added a regression test for the workflowId-mid-fetch race.

PR #886, opened 2026-05-12
fix(web): allow delete/archive of kanban cards in All Workflows view

carlosflorencio:
Deleting or archiving a kanban card from the All Workflows swimlane silently no-opped because `useTaskCRUD` early-returned when `kanban.workflowId` was null, leaving the dialog closing but the task still on the board. Removing the workflow guard (these operations are task-scoped, not workflow-scoped) restores the expected behaviour.

## Validation

- TDD: added two regression e2e tests in `apps/web/e2e/tests/kanban/card-menu-delete-archive.spec.ts` that fail without the fix and pass with it.
- `pnpm --filter @kandev/web e2e -- tests/kanban/card-menu-delete-archive.spec.ts` (5/5 pass).
- `pnpm --filter @kandev/web typecheck`, `pnpm --filter @kandev/web lint`, full web unit suite (1315 tests pass).
- `make -C apps/backend test lint`.

## Checklist

### DIFF
--- a/apps/web/e2e/tests/kanban/card-menu-delete-archive.spec.ts
+++ b/apps/web/e2e/tests/kanban/card-menu-delete-archive.spec.ts
@@ -1,6 +1,8 @@
 import { test, expect } from "../../fixtures/test-base";
 import { KanbanPage } from "../../pages/kanban-page";
 
+const TASK_VISIBLE_TIMEOUT = 10_000;
+
 // Regression: clicking Delete / Archive must show the confirm dialog, not navigate to the task.
 test.describe("Kanban card actions menu — delete/archive does not navigate", () => {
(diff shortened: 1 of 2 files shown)


carlosflorencio:
Thanks for the review.

1. **Unit test for `useTaskCRUD`**: skipping for this PR. The hook never had a colocated test (pre-existing gap, as noted) and the two new e2e tests fully exercise the regression path. Adding a unit test purely for this two-line guard removal would broaden the scope beyond the bug fix; happy to revisit if the hook gets more logic later.

2. **Inline reason on the eslint-disable**: applied. Moved the explanation onto the same line as the `eslint-disable-next-line` directive so a future reader can't strip `testPage` from the destructure without seeing why it's there.

PR #932, opened 2026-05-18
chore: suppress pnpm modules-purge prompt in non-interactive installs

carlosflorencio:
Running `make start-debug` (or any target that calls `install-web`) would prompt for confirmation when pnpm needed to purge and reinstall node_modules, blocking non-interactive shells and piped runs like `make start-debug | tee logs.txt`. Setting `confirm-modules-purge=false` in `apps/.npmrc` silences this globally without affecting install behaviour.

Also fixes a flaky backend test (`TestFeatures_DefaultOff`) that failed when `KANDEV_DEBUG_DEV_MODE=true` was set in the host environment, causing `DetectEnvironment()` to return `EnvDev` and seed wrong defaults.

**Validation**
- `make -C apps/backend test lint`
- `cd apps && pnpm --filter @kandev/web lint`

**Checklist**
- [ ] Tests added/updated
- [ ] Docs updated if needed

### DIFF
--- a/apps/.npmrc
+++ b/apps/.npmrc
@@ -1,3 +1,5 @@
 # Hoist @lezer/highlight for codemirror theme packages that expect it as a peer dependency
 public-hoist-pattern[]=@lezer/*
 
+confirm-modules-purge=false
+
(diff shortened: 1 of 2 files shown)


carlosflorencio:
Acknowledged - `t.Setenv` is idiomatic here (auto-restores via `t.Cleanup`); the pattern difference from the profiles package helper is fine since these are separate packages.

PR #994, opened 2026-05-21
fix(cli): include ~/.local/bin in service unit PATH for user-mode installs

nlenepveu:
## Summary

`kandev service install` generates a systemd user unit (or launchd user agent) with a hard-coded PATH that omits `~/.local/bin`. That's exactly where agent CLIs land when installed via `npm i -g` with a user prefix, `pipx`, `cargo install`, or fnm/nvm — including the `kandev` binary itself (the unit's `ExecStart` references `~/.local/bin/kandev`!). The daemon can find itself but not its sibling CLIs, so `claude`, `codex`, `opencode`, etc. show up as `not_installed` in the agents list even though they're on the user's interactive PATH.

User-mode units now prepend the loading user's `~/.local/bin`:
- systemd: uses the `%h` specifier (expands to the user's home for `--user` units).
- launchd: interpolates `os.homedir()` at render time (no `%h` equivalent).

System-mode units keep the original system-only PATH — `~/.local/bin` is meaningless for root / `systemUser`.

Repro: install kandev as a user service, install `claude` via `npm i -g` with a user prefix, then call `list_agents` — claude-acp shows `capability_status: "not_installed"`. After this fix, regenerating the unit (`kandev service install`) and restarting picks it up.

## Test plan
- [x] `vitest run src/service/templates.test.ts` — 16 tests pass, including new positive coverage for user-vs-system PATH on both renderers.
- [x] Manual: re-run `kandev service install`, verify `~/.config/systemd/user/kandev.service` contains `Environment=PATH=%h/.local/bin:…`, restart, confirm previously-not-installed agents flip to OK.

### DIFF
--- a/apps/cli/src/service/templates.test.ts
+++ b/apps/cli/src/service/templates.test.ts
@@ -43,6 +43,30 @@ describe("renderSystemdUnit", () => {
     expect(unit).not.toContain("KANDEV_SERVER_PORT");
   });
 
+  it("prepends %h/.local/bin to PATH for user-mode units", () => {
+    const unit = renderSystemdUnit({
+      launcher: NPM_LAUNCHER,
+      homeDir: "/home/alice/.kandev",
+      logDir: "/home/alice/.kandev/logs",
+      mode: "user",
+    });
+    expect(unit).toMatch(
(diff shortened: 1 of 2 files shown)


jcfs (commented):
Thanks for the fix — the user-vs-system PATH split is clean and the tests cover the right cases.

jcfs:
Hey @nlenepveu — this PR has merge conflicts with main (specifically with #1000, which also touches the service unit PATH in `apps/cli/src/service/templates.ts` and its tests). Could you rebase onto latest `main` and resolve the conflicts? Once that's done we can get it merged. Thanks!

PR #1007, opened 2026-05-22
fix(release): build kandev binary with -tags fts5

### LINKED ISSUE #1006: [Bug]: npx kandev@latest fails with 'no such module: fts5' after running make start from local repo
## Summary

Running `make start` from the local kandev repo migrates the production database at `~/.kandev/data/kandev.db` to a newer schema version. After that, `npx kandev@latest` fails on every subsequent run with `failed to initialize schema: no such module: fts5`.

## Affected area

CLI

## Kandev version

npx: `v0.50.0` (internal version string: `dev`)
Local repo build: `v0.50.0-26-g3c7b6c3d`

## Install mode

Source build

## Environment

- OS: Windows 11 + WSL2
- Architecture: amd64
- Kernel: 6.6.87.2-microsoft-standard-WSL2
- Node: v25.2.1

## Kandev configuration

- Install mode: `npx kandev@latest` (npm release)
- Relevant env vars: none set

## Task/session context

N/A

## Steps to reproduce

1. Have `npx kandev@latest` working normally (DB at `~/.kandev/data/kandev.db` with version `dev`)
2. Clone the kandev repo locally
3. Run `make start` from the repo root
4. Stop the local server
5. Run `npx kandev@latest` again
6. Backend exits immediately with code 1

## Reproducibility

Always

## Clean-state check

Does not reproduce on clean state (deleting `~/.kandev/data/kandev.db` and letting npx recreate it works)

## Expected behavior

`npx kandev@latest` should start normally regardless of whether a local dev build was previously run against the same database.

## Actual behavior

Backend exits with code 1. Full backend log:

```text
[kandev] release: (npm)
[kandev] url: http://localhost:38429
[kandev] mcp: http://localhost:38429/mcp
[kandev] db: ~/.kandev/data/kandev.db
[kandev] log level: warn
[kandev] starting backend...
[kandev] backend exited (code=1, signal=null)
[kandev] shutting down (backend exit)...
```

Running the backend binary directly reveals the root cause:

```text
INFO  Starting Kandev (unified mode)...
INFO  Initializing Task Service...
INFO  pre-migration backup taken  {"from_version": "v0.50.0-26-g3c7b6c3d", "to_version": "dev", ...}
INFO  Database initialized (single-writer pool)  {"db_path": "~/.kandev/data/kandev.db", "db_driver": "sqlite"}
ERROR Failed to initialize repositories  {"error": "failed to initialize schema: no such module: fts5"}
```

## Root cause (analysis)

`make start` uses `apps/cli/src/start.ts` which defaults the DB path to `~/.kandev/data/kandev.db` — the same production database used by `npx kandev@latest`. The locally built binary (`v0.50.0-26-g3c7b6c3d`) writes its version to `kandev_meta` and runs schema migrations including FTS5 virtual tables.

When `npx kandev@latest` (internal version: `dev`) starts next, it detects a version mismatch (`v0.50.0-26-g3c7b6c3d` ≠ `dev`), attempts to re-initialize the schema, and fails because the SQLite bundled in the npm release binary does not have the FTS5 module compiled in.

Note: `make dev` correctly isolates state under `<repo>/.kandev-dev/` via `KANDEV_HOME_DIR` and does NOT have this problem. Only `make start` is affected.

## Workaround

Manually reset the `kandev_version` key in the meta table back to `dev`:

```bash
python3 -c "
import sqlite3
conn = sqlite3.connect('/home/<user>/.kandev/data/kandev.db')
conn.execute(\"UPDATE kandev_meta SET value='dev' WHERE key='kandev_version'\")
conn.commit()
conn.close()
"
```

Or restore from the auto-backup kandev creates before migrating (found in `~/.kandev/data/backups/`).

## Last known good version

`v0.50.0` (npx) — worked before `make start` was run from the local repo.

## Extra context

Two related issues:
1. `make start` should either warn users it will mutate the production DB, or default to an isolated path (like `make dev` does with `.kandev-dev/`).
2. The FTS5 module appears to be missing from the SQLite bundled in the npm release binary (`@kdlbs/runtime-linux-x64`), but required by the schema init path triggered on version mismatch.

## Before submitting

- [x] I can reproduce this using the steps above.
- [x] I removed secrets, tokens, private repository URLs, private prompts, and customer data.
- [x] I searched existing issues and did not find a duplicate.

jcfs:
## Summary

- v0.50.0 introduced the `tasks_fts` FTS5 virtual table + sync triggers (#914) and added `-tags fts5` to `apps/backend/Makefile`, but **not** to `.github/workflows/release.yml`. The published binary therefore ships without the SQLite fts5 module.
- Fresh installs are unaffected — `migrateTaskFTS` swallows the "module not available" error and search falls back to LIKE. But any DB that an FTS5-enabled build has already touched contains `tasks_fts` + triggers that reference it, and the non-FTS5 release binary fails schema init on the next DML against `tasks` with `no such module: fts5`.
- This patch adds `-tags fts5` to the release-workflow `kandev` build and propagates the standard version ldflags from the Makefile so `kandev --version` reports the released tag instead of the build-time default `dev`.

Fixes #1006.

## Test plan

- [ ] CI builds the release artifacts cleanly on linux/darwin/windows
- [ ] After release, reproduce the bug scenario from #1006: run a local FTS5-enabled `make start`, then `npx kandev@latest start` against the same `~/.kandev/data/kandev.db` — boot should succeed
- [ ] `kandev --version` on the released binary reports the tag (not `dev`)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -360,15 +360,28 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y gcc-mingw-w64-x86-64
 
+      # The `fts5` build tag is REQUIRED on the kandev binary. The office
+      # task search migrations create a `tasks_fts` virtual table using
+      # `USING fts5(...)`; without the tag, mattn/go-sqlite3 lacks the
+      # module and any subsequent boot against a DB that already has the
+      # virtual table fails with "no such module: fts5" during schema init.
+      # See issue #1006 for the regression this prevents.
+      #
+      # Version ldflags mirror apps/backend/Makefile so `kandev --version`
+      # reports the released tag instead of the build-time default "dev".
       - name: Build backend binaries
         env:
           CC: ${{ matrix.cc || '' }}
+          KANDEV_VERSION: ${{ needs.prepare.outputs.tag }}
         run: |
           cd apps/backend
           mkdir -p bin
           EXT=${{ matrix.goos == 'windows' && '.exe' || '' }}
-          go build -o bin/kandev${EXT} ./cmd/kandev
-          go build -o bin/agentctl${EXT} ./cmd/agentctl
+          COMMIT=$(git rev-parse --short HEAD)
+          BUILD_TIME=$(date -u '+%Y-%m-%dT%H:%M:%SZ')
+          LDFLAGS="-s -w -X main.Version=${KANDEV_VERSION} -X main.Commit=${COMMIT} -X main.BuildTime=${BUILD_TIME}"
+          go build -tags fts5 -ldflags "${LDFLAGS}" -o bin/kandev${EXT} ./cmd/kandev
+          go build -ldflags "${LDFLAGS}" -o bin/agentctl${EXT} ./cmd/agentctl
 
       # Cross-compile a linux/amd64 agentctl regardless of the host platform.
       # Remote executors (local Docker, remote Docker, Sprites) bind-mount this
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -380,6 +380,9 @@ jobs:
           COMMIT=$(git rev-parse --short HEAD)
           BUILD_TIME=$(date -u '+%Y-%m-%dT%H:%M:%SZ')
           LDFLAGS="-s -w -X main.Version=${KANDEV_VERSION} -X main.Commit=${COMMIT} -X main.BuildTime=${BUILD_TIME}"
+          # Export to subsequent steps so the cross-compiled agentctl helper
+          # below gets the same version metadata (matches Makefile parity).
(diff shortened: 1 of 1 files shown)


jcfs:
@greptileai — good catch, fixed in 0f32efd. Exported `LDFLAGS` via `$GITHUB_ENV` from the "Build backend binaries" step so the cross-compiled `agentctl-linux-amd64` helper picks up the same `-X main.Version/Commit/BuildTime` metadata, matching `apps/backend/Makefile`'s `build-agentctl-linux` target.

@claude — agreed the `-X` flags on `agentctl` are currently no-ops because `cmd/agentctl/main.go` doesn't declare those variables. Keeping them anyway for Makefile parity and so it just works if/when `agentctl` grows a `--version` flag of its own.

Issue #1032, opened 2026-05-23
CLI passthrough: three load-bearing gaps requiring per-deployment workarounds (cold-launch loop, unreliable submit, no MCP injection)


dbrown99c:
We've been driving an autonomous workflow on top of Kandev cli_passthrough sessions for ~a week and hit three distinct gaps that all require local workarounds. Filing as a single issue so maintainers can see the systemic pattern; happy to split into individual issues if preferred.

Context: passthrough Claude Code TUI on Windows, claude-acp profile with `cli_passthrough: true`, custom scoping/development workflows.

Related: #1031 (mobile UI passthrough render bug), PR #923 (auto-inject task description on idle).

---

### Symptom
A passthrough session created with `start_agent: false` and no prior `EnsureSession` call lands on a workflow column. The on_enter `auto_start_agent` event fires, the launcher spawns `claude.exe` bare (no positional prompt, no `-c`, no `--resume`). Claude defaults to "continue last conversation" semantics, finds none, exits, gets auto-restarted by `handlePassthroughExit`, loops forever.

### Repro
1. POST a task with `start_agent: false` directly into a column whose step has `on_enter: [{ type: auto_start_agent }]`
2. Don't call `session.ensure`
3. The lifecycle eventually spawns the PTY via `processOnEnter` → `autoStartPassthroughPrompt` → `WritePassthroughStdin` against an execution that hasn't started yet → falls into restart path
4. Process listing shows `claude.exe --verbose --model default --dangerously-skip-permissions` with NO positional arg — that's the broken case

### Root cause
`apps/backend/internal/agent/runtime/lifecycle/manager_passthrough.go`'s `freshPassthroughCommand` deliberately builds a Command with no SessionID / no Resume / no Prompt:

```go
cmd := resolved.agent.BuildPassthroughCommand(agents.PassthroughOptions{
    Model:            profileModel(resolved.profile),
    PermissionValues: profilePermissionValues(resolved.profile),
    CLIFlagTokens:    m.profileCLIFlagTokens(resolved.profile),
})
```

When that's the relaunch after a fast-fail, claude exits immediately because there's nothing to resume and no prompt to process. `attemptResumeFallback` writes the "[No prior conversation to resume]" banner and starts the same fresh command again. Infinite loop.

### Workaround we ship
[...]

PR #1033, opened 2026-05-23
fix(mobile): passthrough TUI rendering + touch-scroll (closes #1031)

### LINKED ISSUE #1031: Mobile UI: passthrough TUI silently disappears on re-mount — Chat tab falls back to TaskChatPanel
## Summary

On mobile viewport, the Chat tab for cli_passthrough sessions sometimes renders the `TaskChatPanel` (description bubble + status indicator) instead of the live `<PassthroughTerminal>` xterm, even though the backend PTY process is alive and healthy.

## Repro

Reliably reproducible in Chrome devtools:

1. Create a task with a `cli_passthrough: true` agent profile (e.g. claude-acp default)
2. Open the task in desktop viewport — Chat tab shows the live TUI ✓
3. Toggle Chrome devtools mobile viewport (iPhone/Pixel preset)
4. The Chat tab swaps to the chat-bubble view (no TUI)
5. Toggle back to desktop viewport — TUI returns

Also reproduces on real mobile devices: navigate away from the task panel and back, TUI is gone.

## Root cause

`apps/web/components/task/mobile/session-mobile-layout.tsx:68`:

```jsx
{isPassthroughMode ? (
  <PassthroughTerminal sessionId={effectiveSessionId} mode="agent" />
) : (
  <TaskChatPanel sessionId={effectiveSessionId} onOpenFile={onOpenFile} />
)}
```

`isPassthroughMode` is derived in `apps/web/components/task/task-center-panel.tsx:55`:

```js
const isPassthroughMode = useMemo(() => {
  if (!activeSession?.agent_profile_snapshot) return false;
  const snapshot = activeSession.agent_profile_snapshot as any;
  return snapshot?.cli_passthrough === true;
}, [activeSession?.agent_profile_snapshot]);
```

The memo returns `false` when `agent_profile_snapshot` is missing from the Redux store's session record. The DB has `cli_passthrough: true` populated correctly in the JSON blob, AND `task_sessions.is_passthrough = 1` on the column — only the frontend's cached snapshot is sometimes undefined on re-mount.

## Why desktop isn't affected

`apps/web/components/task/dockview-desktop-layout.tsx:259` renders `<PassthroughTerminal>` **unconditionally** for the passthrough panel slot — no `isPassthroughMode` ternary. So desktop never falls back even when the snapshot is stale.

## Proposed fix

Either:

**A.** Make `isPassthroughMode` accept the `is_passthrough` boolean on the session as authoritative:

```js
const isPassthroughMode = useMemo(() => {
  if (activeSession?.is_passthrough === true) return true;
  if (!activeSession?.agent_profile_snapshot) return false;
  return (activeSession.agent_profile_snapshot as any)?.cli_passthrough === true;
}, [activeSession?.is_passthrough, activeSession?.agent_profile_snapshot]);
```

This matches the backend's source-of-truth logic: `StartAgentProcess` (`apps/backend/internal/agent/runtime/lifecycle/manager_startup.go`) uses `execution.IsPassthrough` (sourced from `TaskSession.IsPassthrough`) as the routing authority, with a comment that says exactly this: *"the session-creation snapshot is the source of truth: a profile that toggles CLIPassthrough after the session was created must not strand existing sessions in the wrong launch path."*

**B.** Or: ensure the session fetch path that hydrates the Redux store on mobile re-mount always includes the `agent_profile_snapshot` field.

A is cheaper and consistent with the existing backend invariant.

## Environment

- Kandev: native Windows binary (NSSM service), commit ~current main as of 2026-05-22
- Browser: Chrome 130 desktop with mobile viewport
- Also reproed on iOS Safari mobile (same task across network changes)

dbrown99c:
## Summary

Two surgical fixes for the mobile cli_passthrough UX issues reported in #1031.

### 1. `task-center-panel.tsx` — passthrough TUI silently disappears on re-mount

`isPassthroughMode` previously only consulted `agent_profile_snapshot.cli_passthrough`. When the Redux store re-hydrated a session without the full snapshot (the leaner session-list responses populate the row but omit the JSON snapshot), the memo returned `false` and the mobile chat panel rendered `TaskChatPanel` (description bubble + status banner) instead of the live `PassthroughTerminal`.

Desktop wasn't affected because `dockview-desktop-layout.tsx:250` (and `dockview-header-actions.tsx:68`, `task-page-content.tsx:133`) already reads `session.is_passthrough` directly as the source of truth — that matches what `manager_startup.go`'s `StartAgentProcess` routes on. Mobile was the only place still using only the snapshot, so the bug was mobile-only by accident, not by design.

Fix: short-circuit on `activeSession.is_passthrough === true`, fall back to the snapshot check only when that field isn't populated (e.g. legacy sessions before the column was added).

### 2. `passthrough-terminal.tsx` — terminal scrollback unreachable on mobile

xterm.js handles mouse wheel scrolling natively but doesn't wire any touch handlers. Swiping inside the terminal on mobile bubbled up the gesture and scrolled the page — meaning anything that scrolled off the top of the terminal viewport (which is most of a long-running agent session) was unreachable. The user could only see whatever fit in the current screen.

New `useMobileTouchScroll` hook attaches `touchstart` / `touchmove` / `touchend` to the terminal div, maps vertical drag delta to `Terminal.scrollLines(...)`, and only activates on `(pointer: coarse)` so desktop xterm behaviour stays identical. Row-height is sourced from xterm's internal `_core._renderService.dimensions.css.cell.height` with a measured-from-DOM fallback and a 16px constant as a last resort.

## Test plan

- [ ] Manual: open a `cli_passthrough: true` task on mobile (or Chrome devtools mobile viewport), verify TUI renders in Chat tab
- [ ] Manual: navigate to a different mobile panel (Plan / Changes / Files / Terminal) and back to Chat — TUI re-mounts, doesn't fall back to `TaskChatPanel`
- [ ] Manual: in mobile Chat tab with a TUI that has scrolled output, swipe vertically inside the terminal — scrollback scrolls, page stays put
- [ ] Manual: desktop unchanged — mouse wheel still works, text selection still works (the touch hook bails on `(pointer: fine)`)
- [ ] Automated: no existing tests for `task-center-panel.tsx` or `passthrough-terminal.tsx` — happy to add Vitest unit tests for `isPassthroughMode` and/or a Playwright spec for the mobile re-mount scenario if maintainers prefer that bar before merge

## Related

- Closes #1031
- Filed alongside #1032 (broader cli_passthrough first-class gaps); these two fixes don't depend on that work.
- Out of scope for this PR but related (planned follow-up): expose the `MobileTerminalKeybar` in the Chat tab for passthrough sessions, not just the Terminal tab. Currently the keybar's `visible` prop in `session-mobile-layout.tsx:394` is gated on `currentMobilePanel === "terminal"` only, so passthrough agents on mobile have no way to send Ctrl-C / Ctrl-D / Tab / arrows. Will file separately once this lands.

### DIFF
--- a/apps/web/components/task/passthrough-terminal.tsx
+++ b/apps/web/components/task/passthrough-terminal.tsx
@@ -249,6 +249,7 @@ export function PassthroughTerminal(props: PassthroughTerminalProps) {
   });
 
   usePendingCommand(pendingCommand, isConnected, wsRef, onCommandSent);
+  useMobileTouchScroll(refs.terminalRef, refs.xtermRef, isTerminalReady);
 
   return (
     <div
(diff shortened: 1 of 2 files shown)


dbrown99c:
Closing while we iterate further on this fix locally. The diff missed one of two  memos (only fixed , not , which is what the mobile layout actually consumes), and pull-to-refresh on iOS Safari still triggers a re-mount race we want to harden against before a fresh PR. Will reopen once validated end-to-end on real device.

PR #1034, opened 2026-05-23
fix(mobile): isPassthroughMode falls back to TaskChatPanel when snapshot is missing (closes #1031)

### LINKED ISSUE #1031: Mobile UI: passthrough TUI silently disappears on re-mount — Chat tab falls back to TaskChatPanel

dbrown99c:
## Summary

Mobile chat panel for a `cli_passthrough` task could silently render `<TaskChatPanel>` (chat-bubble + "Loading messages…" status) instead of the live `<PassthroughTerminal>` xterm, and stay stuck there until a layout swap forced a remount (portrait→landscape uses the desktop layout, which renders `PassthroughTerminal` unconditionally).

## Root cause

`apps/web/hooks/use-session-layout-state.ts`'s `isPassthroughMode` memo only consulted `agent_profile_snapshot.cli_passthrough`. The session row has a dedicated `is_passthrough` column that `dockview-desktop-layout.tsx`, `dockview-header-actions.tsx`, `task-page-content.tsx`, etc. already use as the authoritative source. The WS handler at `apps/web/lib/ws/handlers/agent-session.ts:100-102` writes the snapshot under a truthy guard but `is_passthrough` under `!== undefined` — partial `state_changed` events can land the latter without the former, leaving the snapshot absent and the old memo returning `false`.

## Fix

Honor any defined `is_passthrough` value (including `false`) as authoritative; fall back to the snapshot for legacy session rows that pre-date the column.

```diff
   const isPassthroughMode = useMemo(() => {
+    if (activeSession?.is_passthrough !== undefined) return activeSession.is_passthrough;
     if (!activeSession?.agent_profile_snapshot) return false;
     const snapshot = activeSession.agent_profile_snapshot as any;
     return snapshot?.cli_passthrough === true;
-  }, [activeSession?.agent_profile_snapshot]);
+  }, [activeSession?.is_passthrough, activeSession?.agent_profile_snapshot]);
```

This brings mobile in line with the rest of the codebase's pattern and with `manager_startup.go`'s backend invariant (`StartAgentProcess` routes on `execution.IsPassthrough` which is sourced from `TaskSession.IsPassthrough`).

## Test plan

- [x] Manual on iOS Safari, mobile passthrough task: open Chat tab → TUI renders
- [x] Navigate to Plan/Changes/Files panel and back → TUI re-renders, no fallback to chat-bubble
- [x] Trigger an iOS pull-to-refresh on the chat tab → page reloads → TUI renders on first paint instead of stuck "Loading messages…"
- [x] Desktop browser unchanged — already used `is_passthrough` direct check
- [x] `pnpm typecheck` clean
- [x] `pnpm test` clean (2152 tests)

## Related

- Closes #1031
- Mobile terminal touch-scroll (the other half of #1031's repro) split out — filing as a separate issue. Several approaches tried (custom touch→`scrollLines` mapping, with and without inertia, fractional-row accumulation, direction-reversal handling) all surfaced iOS Safari pull-to-refresh leaks and direction-reversal edge cases that deserve a focused upstream discussion rather than shipping a workaround alongside this rendering fix.


carlosflorencio:
Thanks, I've approved the ci to run. Ideally we would also add an e2e test to prevent future regressions but i can add one later.

PR #1040, opened 2026-05-23
feat: add per-agent-profile environment variables

Foprta:
## Summary
- Add `env_vars` to agent profiles (DB, API, launch merge into agent subprocess env)
- Settings UI on agent profile page (reuse `EnvVarsCard`, secret-backed values)
- Profile env fills missing keys only; office/executor env wins on collisions

## Test plan
- [x] Save env vars on Claude profile, start task, verify subprocess receives them
- [ ] Secret-backed env var resolves at launch
- [ ] Existing executor profile env still works alongside profile env

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/backend/internal/agent/runtime/lifecycle/manager_launch.go
+++ b/apps/backend/internal/agent/runtime/lifecycle/manager_launch.go
@@ -313,7 +313,6 @@ func (m *Manager) launchPrepareRequest(req *LaunchRequest, profileInfo *AgentPro
 		if profileInfo.AutoApprove {
 			reqWithWorktree.Env["AGENTCTL_AUTO_APPROVE_PERMISSIONS"] = "true"
 		}
-		m.mergeAgentProfileEnvFromInfo(context.Background(), profileInfo, reqWithWorktree.Env)
 	}
 	mergeRouteOverrideEnv(&reqWithWorktree)
 	return reqWithWorktree, executionID
@@ -419,7 +418,7 @@ func (m *Manager) launchBuildExecutorRequest(ctx context.Context, executionID st
 		return nil, nil, nil, fmt.Errorf("no runtime configured: %w", err)
 	}
 
-	env := m.buildEnvForExecution(executionID, reqWithWorktree, agentConfig)
+	env := m.buildEnvForExecution(ctx, executionID, reqWithWorktree, agentConfig)
 
 	acpMcpServers, err := m.resolveMcpServersWithParams(ctx, reqWithWorktree.AgentProfileID, reqWithWorktree.Metadata, agentConfig)
 	if err != nil {
@@ -929,7 +928,7 @@ func (m *Manager) SetExecutionDescription(_ context.Context, executionID string,
 }
 
 // SetExecutionEnv stores per-run environment variables for the next agent subprocess start.
-func (m *Manager) SetExecutionEnv(ctx context.Context, executionID string, env map[string]string) error {
+func (m *Manager) SetExecutionEnv(_ context.Context, executionID string, env map[string]string) error {
 	execution, exists := m.executionStore.Get(executionID)
 	if !exists {
 		return fmt.Errorf("execution %q not found", executionID)
@@ -937,12 +936,7 @@ func (m *Manager) SetExecutionEnv(ctx context.Context, executionID string, env m
 	if execution.Metadata == nil {
 		execution.Metadata = make(map[string]interface{})
 	}
-	merged := cloneStringMap(env)
-	if merged == nil {
-		merged = make(map[string]string)
-	}
-	m.mergeAgentProfileEnv(ctx, execution.AgentProfileID, merged)
-	execution.Metadata["runtime_env"] = merged
+	execution.Metadata["runtime_env"] = cloneStringMap(env)
 	return nil
 }
 
@@ -1059,6 +1053,7 @@ func (m *Manager) configureAndStartAgent(ctx context.Context, execution *AgentEx
 	if taskDescription != "" {
 		env["TASK_DESCRIPTION"] = taskDescription
 	}
+	m.mergeAgentProfileEnv(ctx, execution.AgentProfileID, env)
 
 	if err := execution.agentctl.ConfigureAgent(ctx, execution.AgentCommand, env, approvalPolicy, execution.ContinueCommand); err != nil {
 		return "", fmt.Errorf("failed to configure agent: %w", err)
--- a/apps/backend/internal/agent/runtime/lifecycle/manager_launch.go
+++ b/apps/backend/internal/agent/runtime/lifecycle/manager_launch.go
@@ -412,13 +412,13 @@ func (r *prepareProgressRecorder) recordStep(step PrepareStep, index int) {
 
 // launchBuildExecutorRequest resolves MCP servers, builds the ExecutorCreateRequest,
 // and creates the runtime instance.
-func (m *Manager) launchBuildExecutorRequest(ctx context.Context, executionID string, reqWithWorktree *LaunchRequest, agentConfig agents.Agent, mainRepoGitDir, worktreeID, worktreeBranch string, onProgress PrepareProgressCallback) (*ExecutorCreateRequest, *ExecutorInstance, ExecutorBackend, error) {
+func (m *Manager) launchBuildExecutorRequest(ctx context.Context, executionID string, reqWithWorktree *LaunchRequest, agentConfig agents.Agent, profileInfo *AgentProfileInfo, mainRepoGitDir, worktreeID, worktreeBranch string, onProgress PrepareProgressCallback) (*ExecutorCreateRequest, *ExecutorInstance, ExecutorBackend, error) {
 	rt, err := m.getExecutorBackend(reqWithWorktree.ExecutorType)
 	if err != nil {
 		return nil, nil, nil, fmt.Errorf("no runtime configured: %w", err)
 	}
 
-	env := m.buildEnvForExecution(ctx, executionID, reqWithWorktree, agentConfig)
+	env := m.buildEnvForExecution(ctx, executionID, reqWithWorktree, agentConfig, profileInfo)
 
 	acpMcpServers, err := m.resolveMcpServersWithParams(ctx, reqWithWorktree.AgentProfileID, reqWithWorktree.Metadata, agentConfig)
 	if err != nil {
@@ -808,7 +808,7 @@ func (m *Manager) launchInternal(ctx context.Context, req *LaunchRequest) (*Agen
 	if req.ACPSessionID == "" {
 		runtimeProgress = progressRecorder.Callback(progressRecorder.Len())
 	}
-	execReq, execInstance, rt, err := m.launchBuildExecutorRequest(ctx, executionID, &reqWithWorktree, agentConfig, mainRepoGitDir, worktreeID, worktreeBranch, runtimeProgress)
+	execReq, execInstance, rt, err := m.launchBuildExecutorRequest(ctx, executionID, &reqWithWorktree, agentConfig, profileInfo, mainRepoGitDir, worktreeID, worktreeBranch, runtimeProgress)
 	if err != nil {
 		m.publishLaunchPrepareCompleted(req, prepResult, progressRecorder, workspacePath, false, err)
 		return nil, err
@@ -833,6 +833,9 @@ func (m *Manager) launchInternal(ctx context.Context, req *LaunchRequest) (*Agen
 	// Build the in-memory AgentExecution from the runtime instance. Extracted
 	// to keep launchInternal under the cyclomatic-complexity budget.
 	execution := m.buildExecutionFromInstance(req, execReq, execInstance, rt, profileInfo, agentConfig, prepResult)
+	if profileInfo != nil && len(profileInfo.EnvVars) > 0 {
+		m.cacheResolvedProfileEnv(execution, m.resolveAgentProfileEnvVars(ctx, profileInfo.EnvVars))
+	}
 
 	// Track + persist + publish. Returns the rollback error if Add lost a race.
 	if err := m.registerAndPublishExecution(ctx, execution, rt, execInstance, req.SessionID); err != nil {
@@ -1053,7 +1056,7 @@ func (m *Manager) configureAndStartAgent(ctx context.Context, execution *AgentEx
 	if taskDescription != "" {
 		env["TASK_DESCRIPTION"] = taskDescription
 	}
-	m.mergeAgentProfileEnv(ctx, execution.AgentProfileID, env)
+	m.mergeAgentProfileEnvForExecution(ctx, execution, env)
 
 	if err := execution.agentctl.ConfigureAgent(ctx, execution.AgentCommand, env, approvalPolicy, execution.ContinueCommand); err != nil {
 		return "", fmt.Errorf("failed to configure agent: %w", err)
--- a/apps/backend/internal/agent/runtime/lifecycle/manager_launch.go
+++ b/apps/backend/internal/agent/runtime/lifecycle/manager_launch.go
@@ -313,6 +313,7 @@ func (m *Manager) launchPrepareRequest(req *LaunchRequest, profileInfo *AgentPro
 		if profileInfo.AutoApprove {
 			reqWithWorktree.Env["AGENTCTL_AUTO_APPROVE_PERMISSIONS"] = "true"
 		}
+		m.mergeAgentProfileEnvFromInfo(context.Background(), profileInfo, reqWithWorktree.Env)
 	}
 	mergeRouteOverrideEnv(&reqWithWorktree)
 	return reqWithWorktree, executionID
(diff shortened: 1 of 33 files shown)


apps/web/lib/api/domains/agent-profile-normalize.test.ts
```diff
@@ -13,6 +13,7 @@ describe("normalizeAgentProfile", () => {
       mode: "acp",
       allow_indexing: true,
       cli_flags: [{ flag: "--verbose", description: "v", enabled: true }],
+      env_vars: [{ key: "ANTHROPIC_BASE_URL", value: "https://api.example" }],
```
carlosflorencio: [blocker] CI fails here. `--max-warnings 0` plus the `sonarjs/no-duplicate-string` rule (threshold 4) trips on `"ANTHROPIC_BASE_URL"` appearing four times in this file. Hoist it to a const at the top of the file (e.g. `const ANTHROPIC_BASE_URL_KEY = "ANTHROPIC_BASE_URL"`) and reuse it. Confirmed locally with `pnpm --filter @kandev/web lint`.

apps/backend/internal/agent/settings/handlers/handlers.go
```diff
@@ -380,6 +383,7 @@ func (h *Handlers) httpCreateProfile(c *gin.Context) {
 		AllowIndexing:  body.AllowIndexing,
 		CLIPassthrough: body.CLIPassthrough,
 		CLIFlags:       body.CLIFlags,
+		EnvVars:        body.EnvVars,
```
carlosflorencio, line 386: [suggestion] Validation errors from `validateProfileEnvVarDTOs` (duplicate key, value+secret_id set together, key contains `=` or null byte) bubble up through `CreateProfile`/`UpdateProfile` and get mapped to a generic 500 with "failed to create/update profile". The user sees a server error instead of the actionable message. Either define an `ErrInvalidProfileInput` sentinel in the controller and map it to 400 with the underlying message, or use `errors.As` on a typed validation error. Same pattern this file already uses for `ErrAgentProfileNotFound`. Applies to both the create handler here and the update handler around line 438.

apps/backend/internal/agent/settings/controller/profile_crud.go
```diff
@@ -301,6 +315,53 @@ func cliFlagsFromDTO(in []dto.CLIFlagDTO) []models.CLIFlag {
...
+	for _, ev := range in {
+		if strings.TrimSpace(ev.Key) == "" {
+			continue
+		}
+		out = append(out, models.ProfileEnvVar{
+			Key:      strings.TrimSpace(ev.Key),
+			Value:    ev.Value,
+			SecretID: ev.SecretID,
+		})
+	}
```
carlosflorencio, line 340: [suggestion] `validateProfileEnvVarDTOs` checks key uniqueness, '=', and null bytes, but there's no cap on entry count, key length, or value length. The JSON blob is persisted to a SQLite TEXT column with no DB-side limit and then materialized into every agent subprocess environ. A user pasting a multi-MB blob would bloat the row and slow every launch for that profile. Add something generous but bounded (e.g. <=100 entries, key <=256 chars, value <=8 KiB). One pass in this function plus a UI hint covers it.

apps/backend/internal/agent/settings/controller/profile_crud.go
```diff
@@ -301,6 +315,53 @@ func cliFlagsFromDTO(in []dto.CLIFlagDTO) []models.CLIFlag {
...
+	return out
+}
+
```
carlosflorencio, line 343: [suggestion] Profile env can supply `KANDEV_*` keys. Today three call sites (`buildEnvForExecution`, `buildPassthroughEnv`, `configureAndStartAgent`) all happen to either stamp `KANDEV_*` before the fill-missing merge or read from a `runtime_env` snapshot that already contains them, so the user can't actually impersonate `KANDEV_TASK_ID` / `KANDEV_SESSION_ID` / `KANDEV_INSTANCE_ID` today. But that's a cross-file invariant nobody enforces. Cheaper to reject keys starting with `KANDEV_` (and maybe `TASK_DESCRIPTION`) right here in validation and stop worrying about call order downstream.

apps/web/components/settings/agent-profile-page.tsx
```diff
@@ -159,6 +166,64 @@ function ProfileSettingsCard({
...
+  onChange: (patch: Partial<AgentProfile>) => void;
+};
+
+function ProfileEnvVarsSection({ savedProfile, onChange }: ProfileEnvVarsSectionProps) {
+  const { items: secrets } = useSecrets();
+  const handleChange = useCallback((envVars: ProfileEnvVar[]) => onChange({ envVars }), [onChange]);
+
+  return (
+    <ProfileEnvVarsEditor
+      key={savedProfile.updatedAt}
```
carlosflorencio: [suggestion] `ProfileEnvVarsSection` reads `savedProfile.envVars` and uses `key={savedProfile.updatedAt}` to force a remount on save, while every other field reads from `draft`. The render-loop guards in `updateDraft` (around line 462) and the effect at lines 191-195 only exist because of this mismatch (the two follow-up commits "address profile env review findings" and "prevent profile env var render loop" were both patching around it). Make `ProfileEnvVarsEditor` fully controlled (take `value` + `onChange`, drop the internal `useEnvVarRows` state, drop the `useEffect`, drop the `key={savedProfile.updatedAt}`) and pass `draft.envVars` like every other field. `useEnvVarRows` is already exported so the executor-profile page that depends on it keeps working.

apps/backend/internal/agent/runtime/lifecycle/manager_launch.go
```diff
@@ -1041,6 +1041,7 @@ func (m *Manager) configureAndStartAgent(ctx context.Context, execution *AgentEx
 	if taskDescription != "" {
 		env["TASK_DESCRIPTION"] = taskDescription
 	}
+	m.mergeAgentProfileEnv(ctx, execution.AgentProfileID, env)
```
carlosflorencio: [perf] `mergeAgentProfileEnv` runs twice per launch in the Docker/Standalone path (once via `buildEnvForExecution` in `manager_startup.go:234`, again here). Each call hits `profileResolver.ResolveProfile` (DB) and iterates `secretStore.Reveal` for every secret-backed entry. Secret reveals can be slow depending on the backend. Either resolve once during `buildEnvForExecution` and stash the resolved map on `AgentExecution.Metadata`, or skip the merge here when `runtime_env` is already populated.

apps/backend/internal/agent/runtime/lifecycle/manager_launch_test.go
```diff
@@ -132,6 +132,37 @@ func TestBuildAgentCommand_CLIFlagsAppended(t *testing.T) {
...
+
+	runtimeEnv, ok := execution.Metadata["runtime_env"].(map[string]string)
+	if !ok {
+		t.Fatalf("runtime_env missing or wrong type: %#v", execution.Metadata["runtime_env"])
+	}
+	if runtimeEnv["EXECUTOR_ONLY"] != "executor" {
+		t.Fatalf("executor env missing: %+v", runtimeEnv)
+	}
+	if _, exists := runtimeEnv["PROFILE_ONLY"]; exists {
+		t.Fatalf("profile env vars must not be snapshotted into runtime_env: %+v", runtimeEnv)
```
carlosflorencio, line 190: [suggestion] The PR checklist still has "Secret-backed env var resolves at launch" unchecked. Current tests cover the resolver in isolation (`TestResolveAgentProfileEnvVars_SecretAndValue`) and the persistence layer (`TestAgentProfileEnvVars_Roundtrip`), but nothing exercises the full launch path with a real secret store. Add a test next to this one that goes through `buildEnvForExecution` with a populated profile + in-memory secret store and asserts the resolved value ends up in the returned env map.

apps/backend/internal/agent/settings/store/sqlite.go
```diff
@@ -458,6 +466,10 @@ func (r *sqliteRepository) CreateAgentProfile(ctx context.Context, profile *mode
 	if err != nil {
```
carlosflorencio, line 466: [suggestion] Literal `value` is persisted as plaintext JSON in a TEXT column. The Secret mode exists for sensitive values but nothing in the UI nudges users toward it, and the DB file gets copied around (backups, dev syncing, etc). At minimum add inline guidance in the editor card ("Use Secret for tokens and API keys"). Optionally show a soft warning in the UI when a key matches `*_TOKEN` / `*_API_KEY` / `*_SECRET` and the user picked Value mode. Hard reject is probably too aggressive for a self-hosted tool.

apps/backend/internal/agent/settings/models/models.go
```diff
@@ -4,8 +4,12 @@ import (
 	"time"
 
 	agentusage "github.com/kandev/kandev/internal/agent/usage"
+	taskmodels "github.com/kandev/kandev/internal/task/models"
 )
 
+// ProfileEnvVar is an environment variable entry on an agent profile.
+type ProfileEnvVar = taskmodels.ProfileEnvVar
```
carlosflorencio, line 11: [nit] `ProfileEnvVar = taskmodels.ProfileEnvVar` imports `internal/task/models` for a 3-field struct that has nothing to do with tasks. It's a convenience alias but it flips the dependency arrow (settings now depends on task models) and the canonical type is misnamed once it's shared. Consider moving the canonical struct into a neutral package (e.g. here in `agent/settings/models`, or a tiny `internal/envvars` package) and have the task package alias from there.

carlosflorencio (commented):
Thanks for the contribution! Mostly hardening + cleanup notes inline. The lint warning on the test file is the only one that will block CI.

apps/backend/internal/agent/runtime/lifecycle/manager_launch.go
```diff
@@ -313,6 +313,7 @@ func (m *Manager) launchPrepareRequest(req *LaunchRequest, profileInfo *AgentPro
 		if profileInfo.AutoApprove {
 			reqWithWorktree.Env["AGENTCTL_AUTO_APPROVE_PERMISSIONS"] = "true"
 		}
+		m.mergeAgentProfileEnvFromInfo(context.Background(), profileInfo, reqWithWorktree.Env)
```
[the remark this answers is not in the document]
Foprta: Fixed in 13e206f+ — `launchPrepareRequest` no longer merges profile env; `configureAndStartAgent` uses caller `ctx`.

apps/web/lib/api/domains/agent-profile-normalize.test.ts
Foprta: Fixed in 1875b333 — hoisted to `sampleEnvVar` const.

apps/backend/internal/agent/settings/handlers/handlers.go
Foprta: Fixed in 1875b333 — `ErrInvalidProfileEnvVars` → 400 with message.

apps/backend/internal/agent/settings/controller/profile_crud.go
Foprta: Fixed in 1875b333 — ≤100 entries, key ≤256, value ≤8KiB.

apps/backend/internal/agent/settings/controller/profile_crud.go
Foprta: Fixed in 1875b333 — reject `KANDEV_*` and `TASK_DESCRIPTION` in validation.

apps/web/components/settings/agent-profile-page.tsx
Foprta: Fixed in 1875b333 — editor reads `draft.envVars`, no remount/effect loop.

apps/backend/internal/agent/runtime/lifecycle/manager_launch.go
Foprta: Fixed in 1875b333 — cache resolved profile env on execution; reuse in `configureAndStartAgent`.

apps/backend/internal/agent/runtime/lifecycle/manager_launch_test.go
Foprta: Fixed in 1875b333 — `TestBuildEnvForExecution_ResolvesSecretBackedProfileEnv`.

apps/backend/internal/agent/settings/store/sqlite.go
Foprta: Fixed in 1875b333 — EnvVarsCard copy nudges Secret mode for tokens/API keys.

apps/backend/internal/agent/settings/models/models.go
Foprta: Deferred — alias is convenience only; no behavior change this PR.

carlosflorencio:
Added some improvements, hope you don't mind.

image

Foprta:
> Added some improvements, hope you don't mind.

Sure! Thats much better. I'm not in project style yet.

apps/backend/internal/agent/runtime/lifecycle/profile_env.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (m *Manager) cacheResolvedProfileEnv(execution *AgentExecution, resolved map[string]string) {
+	if execution == nil || len(resolved) == 0 {
+		return
+	}
+	if execution.Metadata == nil {
+		execution.Metadata = make(map[string]interface{})
+	}
+	execution.Metadata[metadataKeyProfileEnvResolved] = cloneStringMap(resolved)
+}
```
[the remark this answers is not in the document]
carlosflorencio: Acknowledged, but the concern doesn't apply here.  is intentionally not listed in  (see ), so  strips it before any  write. The resolved values exist only in the in-memory  map for the lifetime of a single launch — they are never serialised to the DB or published through any event.

apps/web/components/settings/cli-flags-field.tsx
```diff
@@ -3,11 +3,51 @@
...
+    description: f.description,
+    enabled: f.enabled,
+  };
+}
+
+function rowToFlag(r: CustomFlagRow): CLIFlag {
+  const flagText = r.flag.trim();
+  const valueText = r.value.trim();
+  return {
+    flag: valueText ? `${flagText} ${valueText}` : flagText,
```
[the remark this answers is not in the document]
carlosflorencio: Fixed. The trim in flagToRow was eating trailing spaces during inline editing (flagToRow is called on every keystroke). stripOuterQuotes now strips the quotes added by shellQuoteValue for display, and rowToFlag no longer trims the value - it calls shellQuoteValue which wraps values containing whitespace in POSIX single quotes so the backend tokeniser produces a single argv token. The add-form commit() trims explicitly at submit time.

apps/web/components/settings/agent-profile-page.tsx
```diff
@@ -352,8 +466,20 @@ function ProfileEditor({
...
+    (patch: Partial<AgentProfile>) => {
+      setDraft((current) => {
+        if (patch.envVars !== undefined && areEnvVarsEqual(patch.envVars, current.envVars)) {
+          return current;
+        }
+        return { ...current, ...patch };
+      });
+    },
+    [setDraft],
+  );
```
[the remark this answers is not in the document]
carlosflorencio: Fixed. The early return was returning current even when patch contained non-envVars fields alongside equal envVars. Now we strip envVars from the patch when they are equal, apply the rest of the patch (if any), and return current only when the patch is effectively empty.

apps/web/components/settings/cli-flags-field.tsx
```diff
@@ -3,11 +3,51 @@
...
+    description: r.description,
+    enabled: r.enabled,
+  };
+}
```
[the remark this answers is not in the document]
carlosflorencio: Fixed. rowToFlag now calls shellQuoteValue which wraps any value containing whitespace in POSIX single quotes before concatenating with the flag name. flagToRow reverses this via stripOuterQuotes so the input always shows the raw (unquoted) value. The backend shell tokeniser therefore receives a single argv token even for values with spaces.

apps/backend/internal/agent/settings/controller/profile_crud.go
```diff
@@ -301,6 +315,78 @@ func cliFlagsFromDTO(in []dto.CLIFlagDTO) []models.CLIFlag {
...
+		if strings.HasPrefix(key, reservedProfileEnvVarPrefix) || key == reservedProfileEnvVarKey {
+			return fmt.Errorf("%w: env_vars[%d].key %q is reserved", ErrInvalidProfileEnvVars, i, key)
+		}
+		if first, exists := seen[key]; exists {
+			return fmt.Errorf("%w: env_vars[%d].key duplicates env_vars[%d].key", ErrInvalidProfileEnvVars, i, first)
+		}
+		seen[key] = i
+		if ev.SecretID != "" && ev.Value != "" {
+			return fmt.Errorf("%w: env_vars[%d]: set value or secret_id, not both", ErrInvalidProfileEnvVars, i)
+		}
```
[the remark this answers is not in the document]
carlosflorencio: Fixed. Backend now rejects entries where both value and secret_id are empty (validateProfileEnvVarDTOs). Frontend add form now shows a secret selector when mode=secret (via DraftValueInput component) and disables Add when mode=secret and no secret is selected.

PR #1063, opened 2026-05-24
fix(office): bound idle-timeout task lookup with caller's context

vimzh:
The idle timeout manager's isTaskTerminal helper queried the office repository with context.Background(), so a stalled DB call could block the calling goroutine indefinitely. The methods are not wired into a request-scoped path in production today, but once OnRunFinished gets attached to an event subscriber or websocket handler, that would translate directly into a stuck request.

Propagate a context.Context through OnRunFinished into isTaskTerminal, and bound the underlying repository call with a 5s timeout derived from the caller's context. The new TestIdleTimeout_CancelledContextDoesNotStartTimer exercises the propagation by passing an already-cancelled context and confirming no idle timer is scheduled.

Existing tests are updated to thread t.Context() through OnRunFinished. Happy-path behavior is unchanged.

There is a near-duplicate IdleTimeoutManager in internal/office/scheduler/idle_timeout.go with the same bug. Left out of this PR to keep it single-concern; happy to follow up with a matching change if you'd prefer it bundled.


carlosflorencio:
Thanks!

PR #1064, opened 2026-05-24
fix(office): stop the backoff timer in channel relay when context is cancelled

vimzh:
sendWithRetry used time.After for its exponential backoff sleep, so a cancelled context returned promptly but left the underlying timer pinned in the Go runtime until it fired. Under repeated retry-then-cancel churn that is a slow goroutine drip and a wasted timer wheel slot per cancellation.

Switch to time.NewTimer with a Stop() on the cancellation branch so the runtime can collect the timer immediately. Behavior on the happy path and on full-retry-exhaustion is unchanged.

Adds TestRelayComment_CancelDuringBackoff to verify the function still returns promptly with context.Canceled and does not sit through the full backoff window.

### DIFF
--- a/apps/backend/internal/office/channels/relay.go
+++ b/apps/backend/internal/office/channels/relay.go
@@ -112,7 +112,13 @@ func (r *ChannelRelay) sendWithRetry(
 			select {
 			case <-timer.C:
 			case <-ctx.Done():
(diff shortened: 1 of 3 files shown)


carlosflorencio:
Thanks for the contributions and small PRs 💪.

PR #1085, opened 2026-05-26
chore: remove orphan helpers, setters, and deprecated wrappers

jcfs:
## Summary
- Remove 12 small dead-code units across the backend: 2 whole files (`internal/agentctl/tracing/executor.go`, `internal/orchestrator/executor/script_template.go`) and 10 orphan methods/functions in otherwise-alive files.
- Includes: 7 `TraceExecutor*` tracing helpers never called from any binary; a deprecated `ResolveScript` wrapper and its now-orphaned `scriptengine.ResolveScript` convenience function; an unwired `handleListExecutors` MCP handler (the sibling profile-handlers ARE wired); five never-called `SetInterval`/`SetIssueTickHook`/`SetMatchHook`/`SetPollInterval` setters across `healthpoll`/`jira`/`linear`/`slack`/`updates`; the redundant `FetchLatestRelease` shim above the alive `FetchLatestReleaseFrom`; the never-called `NewClaudeUsageClient`/`NewCodexUsageClient` shims above the alive `*WithPath` variants; and the never-called `database.Service.Path()` accessor.
- Verified by `golang.org/x/tools/cmd/deadcode -test` and by manual `grep` for each symbol.

## Implementation notes
- Pure deletion; no behavior change.
- Two-step order on the scriptengine pair: `executor.ResolveScript` (caller) deleted before `scriptengine.ResolveScript` (callee), so the callee's transitive deadness becomes visible.
- The `FetchLatestRelease` godoc on `ErrGitHubRateLimited` was updated to point at `FetchLatestReleaseFrom`.

## Test plan
- [x] `make build` succeeds
- [x] `make test` passes (no test count delta expected — deadcode -test already confirmed none of these reach test binaries)
- [x] `make lint` clean
- [x] `grep` for each deleted symbol returns zero hits

## Risks & rollback
- Risk: very low. Each item is grep-verified to have no callers; deletions are mechanical.
- Rollback: `git revert` the deletion commits.


jcfs:
Addressed the Claude review suggestion in 783bdda9: removed `issueTickHook`/`matchHook` fields and their `fireIssueTickHook`/`fireHook` methods from jira, linear, and slack. These were unexported fields that could never be set after the public setters were deleted in the original commits, making the fire methods permanent no-ops.

PR #1093, opened 2026-05-26
fix(backend): honor base_branch for same-repo subtasks via MCP

zeval:
## Summary

- `create_task_kandev` dropped `base_branch` silently when the caller did not also pass `repository_id` / `local_path` / `repository_url`. Same-repo subtasks therefore inherited the parent's repos verbatim, so calls intended to branch off a specific feature branch landed on the parent's branch (e.g. `main`).
- Forward `base_branch` at the top level of the WS payload regardless of repo identifiers, and apply it as an override on the resolved repo list when no per-repo `repositories[]` entries were supplied.
- Explicit per-repo `BaseBranch` still wins — cross-repo and multi-repo callers are untouched.

## Test plan

- [x] `go test ./internal/mcp/... ./internal/task/...` — 766 pass
- [x] `make vet`, `make lint` — clean
- [x] New tests: subtask `base_branch` override path, explicit-repo authoritative path, MCP server top-level forwarding

### DIFF
--- a/apps/backend/internal/mcp/handlers/handlers.go
+++ b/apps/backend/internal/mcp/handlers/handlers.go
@@ -366,6 +366,7 @@ func (h *Handlers) handleCreateTask(ctx context.Context, msg *ws.Message) (*ws.M
 		ExecutorProfileID      string               `json:"executor_profile_id"`
 		StartAgent             *bool                `json:"start_agent"`               // nil means default to true for backward compatibility
 		Repositories           []mcpRepositoryInput `json:"repositories"`              // explicit repositories for top-level tasks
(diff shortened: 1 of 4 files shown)


zeval:
Pushed 133e681 — reworded the `BaseBranch` field doc to match the implemented precedence (top-level value is a fallback applied only when no per-repo `repositories[]` entries are supplied; explicit per-repo `BaseBranch` is authoritative). Thanks @coderabbitai and @greptileai for the catch.

zeval:
Applied @claude's suggestion in 7d741ce — top-level `base_branch` is now only written when no per-repo identifier is supplied, mirroring the backend's override gate (`len(req.Repositories) == 0`). Tests still pass.

Issue #1021, opened 2026-05-22
Refactor: extract WatcherDispatchCoordinator (IoC) for issue-watcher fan-out


nlenepveu:
# Watcher Dispatch Coordinator (IoC refactor)

## Motivation / Why

`internal/orchestrator/event_handlers_linear.go` and `event_handlers_jira.go` are intentional, file-level mirrors. Each one independently does the same pipeline:

```
bus event → reserve dedup → build task request → CreateIssueTask
          → AttachTaskID → maybe StartTask → on failure: ReleaseReservation
```

The duplication is documented in code (`//nolint:dupl // intentional symmetry with createJiraIssueTask`). It hurts us in three concrete ways:

1. **Cross-cutting changes cost 2×.** Throttling, observability, retry, fairness, new auto-start policies — every one has to be applied to both files identically. We already have a follow-up planned to throttle fan-out (broad filters / backlog imports create dozens of concurrent tasks instantly, overloading executors and the user); without this refactor that gate has to land in two places.
2. **New integrations start as another mirror.** GitHub-issues watchers, webhook-driven sources, future Slack/Discord triggers — each currently means copy-pasting `createXIssueTask` and its helpers.
3. **The pipeline shape is invisible.** What's the canonical sequence? You have to read both files and diff them to be sure.

This refactor extracts the shared pipeline into a single `WatcherDispatchCoordinator` driven by integration-specific `WatcherSource` implementations. Linear and Jira keep only the bits that genuinely differ between them; the pipeline shape lives in one place.

This is **scoped to the refactor only** — no behaviour change, no schema change, no new metrics. Throttling, back-drain, and any other capability changes are explicitly out of scope and tracked separately.

## What

Introduce two new pieces in `internal/orchestrator/`:

```go
// WatcherSource encapsulates everything integration-specific about turning a
// freshly-observed external issue into a Kandev task. Linear and Jira each
// implement this; future watchers (GitHub issues, webhooks) implement it too.
type WatcherSource interface {
    // Name returns a stable identifier ("linear", "jira", ...). Used for
    // metrics labels and log fields.
    Name() string

    // Reserve atomically claims the dedup slot for this event. Returns
    // false (without error) when another concurrent reservation already won
    // the race — the coordinator treats that as "nothing to do".
    Reserve(ctx context.Context, evt any) (bool, error)

    // Release rolls back a reservation when task creation fails downstream.
    // Best-effort; errors are logged but not surfaced.
[...]

PR #1094, opened 2026-05-26
fix(watcher): self-heal orphaned watchers when agent profile is soft-deleted

nlenepveu:
## Summary

When the reconciler soft-deletes an agent profile because its agent type left the registry (CLI uninstalled, feature flag toggled off, host-utility probe failed), watchers that reference that profile loop forever on `"profile not found: sql: no rows in result set"`. Each poll creates a dead-on-arrival task; the watcher never disables itself; the user has to manually clean up.

This PR makes the dispatch pipeline self-healing.

### Behaviour change

- **Coordinator pre-flight** (`WatcherDispatchCoordinator.Dispatch`) checks the watcher's bound profile before `Reserve`. If the profile is soft-deleted, the source's `SelfHeal` runs (sets `enabled=0` + stamps `last_error` / `last_error_at`) and the dispatch short-circuits — no task created, no dedup reservation taken. Linear and Jira flow through this seam.
- **GitHub** issue/review watchers bypass the coordinator (legacy `createIssueTask` / `createReviewTask`). The same check fires at the top of each function via a shared helper.
- **Resolver disambiguation:** `ProfileResolver.ResolveProfile` now returns a typed `*DeletedProfileError` (wrapping `store.ErrAgentProfileDeleted`) when the row is found but soft-deleted, distinguishing it from a genuinely missing row. Backed by a new `GetAgentProfileIncludingDeleted` method on the settings store.
- **Settings UI:** `ErrProfileInUseDetail` carries a new `Watchers []WatcherReference` field. The 409 response now serializes it alongside `active_sessions` so the delete confirmation dialog can say "this will also disable N watchers — continue?". `force=true` bypasses both checks; the dispatch pre-flight handles the cleanup on the next poll.

### Architecture

```
Bus event (NewLinearIssueEvent | NewJiraIssueEvent)
  → WatcherDispatchCoordinator.Dispatch
       ├─ preflightDeletedProfile ── NEW ──
       │    profileLookup.LookupProfile(src.AgentProfileID(evt))
       │    ├─ deleted? → src.SelfHeal → store.Disable*WithError → STOP
       │    ├─ lookup error → log + fall through (fail-open on transient DB hiccups)
       │    └─ healthy / empty-id → fall through
       └─ Reserve → BuildTaskRequest → CreateTask → AttachTaskID → maybe Start
```

GitHub bypasses the coordinator but uses the same `ProfileLookup` via `Service.preflightDeletedProfileForGitHub*`.

### Schema

Four watcher tables (`linear_issue_watches`, `jira_issue_watches`, `github_issue_watches`, `github_review_watches`) gained `last_error TEXT NOT NULL DEFAULT ''` and `last_error_at DATETIME` columns via idempotent `ALTER TABLE ... ADD COLUMN` migrations. No FK changes.

## Test plan

- [x] `make fmt` clean
- [x] `go vet ./...` clean
- [x] `go test -race -count=1 ./internal/...` passes
- [x] `golangci-lint run ./...` — 0 issues
- [x] Resolver-level test: `*DeletedProfileError` returned on soft-deleted lookup; `ErrAgentProfileDeleted` matches via `errors.Is`
- [x] Per-store tests: `Disable*WithError` flips `enabled`, stamps `last_error` + `last_error_at`
- [x] Coordinator-level tests (4 cases): deleted profile self-heals, healthy passes preflight, lookup error fails open, empty profile id skips lookup
- [x] GitHub path tests: `createIssueTask` and `createReviewTask` self-heal on deleted profile and skip preflight for legacy empty-id events
- [x] Delete-path tests: `ErrProfileInUseDetail` blocks on watchers, `force=true` bypasses, no-watchers preserves happy path

## Out of scope / follow-ups

- Adjacent surfaces (`automations`, `office_routines`, `office_channels`, `office_agent_memory`) also store `agent_profile_id` and could be orphaned by the same reconciler. Their dispatch paths differ enough to warrant separate design — filing as a follow-up.
- Frontend dialog rendering of the new `watchers` array in the 409 response. The backend wire format is in place; the React component update is a separate PR.

### DIFF
--- a/apps/backend/internal/agent/settings/controller/profile_crud.go
+++ b/apps/backend/internal/agent/settings/controller/profile_crud.go
@@ -213,26 +213,70 @@ func (c *Controller) DeleteProfile(ctx context.Context, id string, force bool) (
 	return &result, nil
 }
 
-// prepareProfileDeletion checks for active sessions and cleans up ephemeral tasks before deletion.
+// prepareProfileDeletion checks for active sessions and referencing watchers,
+// then cleans up ephemeral tasks. Returns *ErrProfileInUseDetail when force
+// is false and any active session OR referencing watcher exists — the UI uses
+// the breakdown to render a confirmation dialog. force=true skips both checks
+// and proactively disables any referencing watchers; the dispatch coordinator's
+// preflight stays as the safety net for profiles deleted by other paths
+// (reconciler cleanup of disabled agent types).
 func (c *Controller) prepareProfileDeletion(ctx context.Context, profileID string, force bool) error {
 	if c.sessionChecker == nil {
 		return nil
 	}
-	// Check for active non-ephemeral sessions (unless force is true)
-	// This allows the UI to show a confirmation dialog with affected tasks
 	if !force {
 		activeTasks, err := c.sessionChecker.GetActiveTaskInfoByAgentProfile(ctx, profileID)
 		if err != nil {
 			return err
 		}
-		if len(activeTasks) > 0 {
-			return &ErrProfileInUseDetail{ActiveSessions: activeTasks}
+		var watcherRefs []WatcherReference
+		if c.watcherDeps != nil {
+			refs, err := c.watcherDeps.ListWatchersByAgentProfile(ctx, profileID)
+			if err != nil {
+				c.logger.Warn("watcher deps lookup failed; proceeding without watcher info",
+					zap.String("profile_id", profileID), zap.Error(err))
+			} else {
+				watcherRefs = refs
+			}
 		}
+		if len(activeTasks) > 0 || len(watcherRefs) > 0 {
+			return &ErrProfileInUseDetail{ActiveSessions: activeTasks, Watchers: watcherRefs}
+		}
+	} else {
+		c.disableReferencingWatchers(ctx, profileID)
 	}
-	// Clean up ephemeral tasks (quick chat, config chat) using this profile
-	// Done after the force check since these don't need user confirmation
+	// Clean up ephemeral tasks (quick chat, config chat) using this profile.
+	// Done after the force check since these don't need user confirmation.
 	c.cleanupEphemeralTasks(ctx, profileID)
 	return nil
+}
+
+// disableReferencingWatchers stamps the deletion cause onto every watcher
+// row that referenced this profile so the UI shows "disabled because the
+// agent profile was deleted" the moment the request returns. Without this
+// eager disable, watchers whose filter no longer matches anything after the
+// profile is gone would stay enabled-but-orphaned indefinitely — the
+// dispatch coordinator's preflight only runs when a new external event
+// fires the watcher.
+//
+// Best-effort: a failure is logged and ignored so the delete still proceeds.
+// The preflight remains as a safety net for reconciler-driven deletes that
+// don't pass through this path.
+func (c *Controller) disableReferencingWatchers(ctx context.Context, profileID string) {
+	if c.watcherDeps == nil {
+		return
+	}
+	cause := fmt.Sprintf("agent profile %s was deleted", profileID)
+	disabled, err := c.watcherDeps.DisableWatchersByAgentProfile(ctx, profileID, cause)
+	if err != nil {
+		c.logger.Warn("failed to disable referencing watchers on force-delete",
+			zap.String("profile_id", profileID), zap.Error(err))
+		return
+	}
+	if len(disabled) > 0 {
+		c.logger.Info("disabled referencing watchers on profile force-delete",
+			zap.String("profile_id", profileID), zap.Int("count", len(disabled)))
+	}
 }
 
 // cleanupEphemeralTasks removes ephemeral tasks (quick chat, config chat) associated with a profile.
--- a/apps/backend/internal/agent/settings/controller/profile_crud.go
+++ b/apps/backend/internal/agent/settings/controller/profile_crud.go
@@ -209,6 +209,13 @@ func (c *Controller) DeleteProfile(ctx context.Context, id string, force bool) (
 		}
 		return nil, err
 	}
+	// Eagerly disable referencing watchers only AFTER the row is gone, so a
+	// failed delete never strands watchers disabled against a still-live
+	// profile. If this disable itself fails, the dispatch coordinator's
+	// preflight self-heals the watchers on their next poll.
+	if force {
+		c.disableReferencingWatchers(ctx, id, profile.Name)
+	}
 	result := toProfileDTO(profile)
 	return &result, nil
 }
@@ -216,10 +223,11 @@ func (c *Controller) DeleteProfile(ctx context.Context, id string, force bool) (
 // prepareProfileDeletion checks for active sessions and referencing watchers,
 // then cleans up ephemeral tasks. Returns *ErrProfileInUseDetail when force
 // is false and any active session OR referencing watcher exists — the UI uses
-// the breakdown to render a confirmation dialog. force=true skips both checks
-// and proactively disables any referencing watchers; the dispatch coordinator's
-// preflight stays as the safety net for profiles deleted by other paths
-// (reconciler cleanup of disabled agent types).
+// the breakdown to render a confirmation dialog. force=true skips both checks;
+// the eager disable of referencing watchers runs in DeleteProfile after the
+// row is actually gone (the dispatch coordinator's preflight stays as the
+// safety net for profiles deleted by other paths, e.g. reconciler cleanup of
+// disabled agent types).
 func (c *Controller) prepareProfileDeletion(ctx context.Context, profileID string, force bool) error {
 	if c.sessionChecker == nil {
 		return nil
@@ -242,8 +250,6 @@ func (c *Controller) prepareProfileDeletion(ctx context.Context, profileID strin
 		if len(activeTasks) > 0 || len(watcherRefs) > 0 {
 			return &ErrProfileInUseDetail{ActiveSessions: activeTasks, Watchers: watcherRefs}
 		}
-	} else {
-		c.disableReferencingWatchers(ctx, profileID)
 	}
 	// Clean up ephemeral tasks (quick chat, config chat) using this profile.
 	// Done after the force check since these don't need user confirmation.
@@ -262,11 +268,11 @@ func (c *Controller) prepareProfileDeletion(ctx context.Context, profileID strin
 // Best-effort: a failure is logged and ignored so the delete still proceeds.
 // The preflight remains as a safety net for reconciler-driven deletes that
 // don't pass through this path.
-func (c *Controller) disableReferencingWatchers(ctx context.Context, profileID string) {
+func (c *Controller) disableReferencingWatchers(ctx context.Context, profileID, profileName string) {
 	if c.watcherDeps == nil {
 		return
 	}
-	cause := fmt.Sprintf("agent profile %s was deleted", profileID)
+	cause := formatDeletedProfileCause(profileID, profileName)
 	disabled, err := c.watcherDeps.DisableWatchersByAgentProfile(ctx, profileID, cause)
 	if err != nil {
 		c.logger.Warn("failed to disable referencing watchers on force-delete",
@@ -277,6 +283,27 @@ func (c *Controller) disableReferencingWatchers(ctx context.Context, profileID s
 		c.logger.Info("disabled referencing watchers on profile force-delete",
 			zap.String("profile_id", profileID), zap.Int("count", len(disabled)))
 	}
+}
+
+// profileNameCauseMaxLen caps the rendered profile name in the deletion
+// cause. Mirrors the orchestrator preflight's cap (80 runes); both strings
+// land in the same settings-page watcher banner, and the name is user-typed
+// with no DB-level length constraint.
+const profileNameCauseMaxLen = 80
+
+// formatDeletedProfileCause renders the human-readable string stamped onto a
+// watcher's last_error when its profile is force-deleted. Includes the profile
+// name (truncated) so the settings banner shows "Kilo Profile" rather than a
+// bare UUID — matching the shape of the orchestrator preflight's cause.
+func formatDeletedProfileCause(profileID, profileName string) string {
+	name := profileName
+	if runes := []rune(name); len(runes) > profileNameCauseMaxLen {
+		name = string(runes[:profileNameCauseMaxLen-1]) + "…"
+	}
+	if name != "" {
+		return fmt.Sprintf("agent profile %q (%s) was deleted", name, profileID)
+	}
+	return fmt.Sprintf("agent profile %s was deleted", profileID)
 }
 
 // cleanupEphemeralTasks removes ephemeral tasks (quick chat, config chat) associated with a profile.
--- a/apps/backend/internal/agent/settings/controller/reconciler_test.go
+++ b/apps/backend/internal/agent/settings/controller/reconciler_test.go
@@ -106,8 +106,22 @@ func (f *fakeStore) DeleteAgentProfile(_ context.Context, id string) error {
 	return nil
 }
 
-func (f *fakeStore) GetAgentProfile(context.Context, string) (*models.AgentProfile, error) {
-	return nil, nil
+func (f *fakeStore) GetAgentProfile(_ context.Context, id string) (*models.AgentProfile, error) {
+	for _, list := range f.profiles {
+		for _, p := range list {
+			if p.ID == id {
+				return p, nil
+			}
+		}
+	}
+	return nil, sql.ErrNoRows
+}
+
+func (f *fakeStore) GetAgentProfileIncludingDeleted(ctx context.Context, id string) (*models.AgentProfile, error) {
+	// Delegate to the deleted_at-aware lookup; the fake does not model
+	// soft-delete state, so both methods see the same rows. This keeps
+	// nil dereferences out of any test that calls the new method.
+	return f.GetAgentProfile(ctx, id)
 }
 
 func (f *fakeStore) ListAgentProfiles(_ context.Context, agentID string) ([]*models.AgentProfile, error) {
--- a/apps/backend/internal/agent/settings/handlers/handlers.go
+++ b/apps/backend/internal/agent/settings/handlers/handlers.go
@@ -466,8 +466,9 @@ func (h *Handlers) httpDeleteProfile(c *gin.Context) {
 		var inUseErr *controller.ErrProfileInUseDetail
 		if errors.As(err, &inUseErr) {
 			c.JSON(http.StatusConflict, gin.H{
-				"error":           "agent profile is used by active session(s)",
+				"error":           "agent profile is used by active session(s) or watcher(s)",
 				"active_sessions": inUseErr.ActiveSessions,
+				"watchers":        inUseErr.Watchers,
 			})
 			return
 		}
--- a/apps/backend/internal/agent/settings/store/sqlite.go
+++ b/apps/backend/internal/agent/settings/store/sqlite.go
@@ -729,25 +729,47 @@ func (r *sqliteRepository) DeleteAgentProfile(ctx context.Context, id string) er
 	return nil
 }
 
+// agentProfileSelectColumns is the SELECT projection used by every
+// AgentProfile read path. Extracted once so the next column added to
+// agent_profiles only has to land here — duplicating the list across
+// GetAgentProfile / GetAgentProfileIncludingDeleted / ListAgentProfiles
+// risks the soft-delete path silently scanning zero values for a freshly
+// added column, which is the same shape of bug this package fixes in
+// another layer (orphaned watchers vs. stale projection drift).
+const agentProfileSelectColumns = `
+	SELECT id, agent_id, name, agent_display_name, model, mode, migrated_from,
+		auto_approve, dangerously_skip_permissions, allow_indexing,
+		cli_passthrough, user_modified, plan, cli_flags,
+		COALESCE(env_vars, '[]'),
+		created_at, updated_at, deleted_at,
+		COALESCE(workspace_id, ''), COALESCE(role, ''), COALESCE(icon, ''),
+		COALESCE(reports_to, ''), COALESCE(skill_ids, '[]'),
+		COALESCE(desired_skills, '[]'), COALESCE(custom_prompt, ''),
+		COALESCE(status, 'idle'), COALESCE(pause_reason, ''),
+		last_run_finished_at,
+		COALESCE(max_concurrent_sessions, 1), COALESCE(cooldown_sec, 0),
+		COALESCE(skip_idle_runs, 0), COALESCE(consecutive_failures, 0),
+		COALESCE(failure_threshold, 3), COALESCE(executor_preference, ''),
+		COALESCE(budget_monthly_cents, 0),
+		COALESCE(settings, '{}'), COALESCE(permissions, '{}')
+	FROM agent_profiles`
+
 func (r *sqliteRepository) GetAgentProfile(ctx context.Context, id string) (*models.AgentProfile, error) {
-	row := r.ro.QueryRowContext(ctx, r.ro.Rebind(`
-		SELECT id, agent_id, name, agent_display_name, model, mode, migrated_from,
-			auto_approve, dangerously_skip_permissions, allow_indexing,
-			cli_passthrough, user_modified, plan, cli_flags,
-			COALESCE(env_vars, '[]'),
-			created_at, updated_at, deleted_at,
-			COALESCE(workspace_id, ''), COALESCE(role, ''), COALESCE(icon, ''),
-			COALESCE(reports_to, ''), COALESCE(skill_ids, '[]'),
-			COALESCE(desired_skills, '[]'), COALESCE(custom_prompt, ''),
-			COALESCE(status, 'idle'), COALESCE(pause_reason, ''),
-			last_run_finished_at,
-			COALESCE(max_concurrent_sessions, 1), COALESCE(cooldown_sec, 0),
-			COALESCE(skip_idle_runs, 0), COALESCE(consecutive_failures, 0),
-			COALESCE(failure_threshold, 3), COALESCE(executor_preference, ''),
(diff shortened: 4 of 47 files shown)


apps/backend/internal/github/store.go
```diff
@@ -182,6 +186,14 @@ func (s *Store) initSchema() error {
...
 	_, _ = s.db.Exec(`ALTER TABLE github_review_watches ADD COLUMN cleanup_policy TEXT NOT NULL DEFAULT 'auto'`)
 	_, _ = s.db.Exec(`ALTER TABLE github_issue_watches ADD COLUMN cleanup_policy TEXT NOT NULL DEFAULT 'auto'`)
+	// Watcher self-heal columns: when the dispatch pipeline detects an
+	// orphaned watcher (e.g. its agent profile has been soft-deleted), it
+	// disables the row and stamps a human-readable cause + timestamp here
+	// for the settings page to surface.
+	_, _ = s.db.Exec(`ALTER TABLE github_review_watches ADD COLUMN last_error TEXT NOT NULL DEFAULT ''`)
+	_, _ = s.db.Exec(`ALTER TABLE github_review_watches ADD COLUMN last_error_at DATETIME`)
+	_, _ = s.db.Exec(`ALTER TABLE github_issue_watches ADD COLUMN last_error TEXT NOT NULL DEFAULT ''`)
+	_, _ = s.db.Exec(`ALTER TABLE github_issue_watches ADD COLUMN last_error_at DATETIME`)
```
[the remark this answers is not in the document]
nlenepveu: @greptile-apps This is a false positive. **Reason**: The silent `_, _ = s.db.Exec(...)` pattern for idempotent ALTER TABLE migrations is the established project contract — the backend AGENTS.md explicitly documents it: *'anything else logs Warn but never returns an error (preserving the existing swallow-error contract)'*. The Jira/Linear paths use `tableColumns()` + explicit errors only for their newer schema migrations; both patterns are intentional and consistent within their respective stores. No change needed.

apps/backend/internal/agent/settings/controller/reconciler_test.go
```diff
@@ -106,7 +106,18 @@ func (f *fakeStore) DeleteAgentProfile(_ context.Context, id string) error {
...
+			if p.ID == id {
+				return p, nil
+			}
+		}
+	}
+	return nil, sql.ErrNoRows
+}
+
+func (f *fakeStore) GetAgentProfileIncludingDeleted(context.Context, string) (*models.AgentProfile, error) {
 	return nil, nil
```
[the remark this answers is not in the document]
nlenepveu: @coderabbitai This was addressed in commit 02f5474f. The fake now delegates to GetAgentProfile which returns a valid *models.AgentProfile for known IDs and sql.ErrNoRows for unknown IDs — both paths are exercised. No further change needed.

apps/backend/internal/linear/store_issue_watch_disable_test.go
```diff
@@ -0,0 +1,55 @@
...
+	if err := store.CreateIssueWatch(ctx, watch); err != nil {
+		t.Fatalf("create: %v", err)
+	}
+
+	const cause = `agent profile "Removed Kilo" (deleted-profile) was removed`
+	before := time.Now().UTC()
+	if err := store.DisableIssueWatchWithError(ctx, watch.ID, cause); err != nil {
+		t.Fatalf("disable: %v", err)
+	}
+	after := time.Now().UTC()
```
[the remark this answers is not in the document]
nlenepveu: @coderabbitai This was addressed in commit 02f5474f. Both the linear and github tests now use `time.Now().UTC().Add(-time.Second)` and `Add(time.Second)` on the window boundaries to absorb SQLite's second-precision truncation.

apps/backend/internal/github/store.go
carlosflorencio: These four `ALTER TABLE` calls drop every error on the floor with `_, _ =`. I know the surrounding `initSchema` already does this, but the new columns are the only ones where the reader side panics if they end up missing: `models.IssueWatch.LastError` / `LastErrorAt` get scanned unconditionally, so a driver-level failure here (lock contention, FS permissions, disk pressure, etc.) turns into a confusing scan panic on the next poll instead of a clear boot error.

The sibling stores already solve this. `apps/backend/internal/jira/store.go:116-132` and `apps/backend/internal/linear/store.go:107` use a `tableColumns()` precheck so a fresh install skips the ALTER and only real failures bubble up. Worth doing the same here:

```suggestion
	if err := s.addWatchSelfHealColumns(); err != nil {
		return err
	}
```

with a helper alongside the other migration helpers in this file:

```go
func (s *Store) addWatchSelfHealColumns() error {
	for _, table := range []string{"github_review_watches", "github_issue_watches"} {
		cols, err := s.tableColumns(table)
		if err != nil {
			return fmt.Errorf("read %s columns: %w", table, err)
		}
		if _, ok := cols["last_error"]; !ok {
			if _, err := s.db.Exec("ALTER TABLE " + table + " ADD COLUMN last_error TEXT NOT NULL DEFAULT ''"); err != nil {
				return fmt.Errorf("add %s.last_error: %w", table, err)
			}
		}
		if _, ok := cols["last_error_at"]; !ok {
			if _, err := s.db.Exec("ALTER TABLE " + table + " ADD COLUMN last_error_at DATETIME"); err != nil {
				return fmt.Errorf("add %s.last_error_at: %w", table, err)
			}
		}
	}
	return nil
}
```

`github.Store` doesn't have a `tableColumns` helper yet, so you'd need to copy the small one from `jira/store.go:264` over. Cheap, and it pays off the moment the next ALTER lands.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/backend/internal/agent/settings/store/sqlite.go
```diff
@@ -755,6 +755,35 @@ func (r *sqliteRepository) GetAgentProfile(ctx context.Context, id string) (*mod
...
+			COALESCE(budget_monthly_cents, 0),
+			COALESCE(settings, '{}'), COALESCE(permissions, '{}')
+		FROM agent_profiles WHERE id = ?
+	`), id)
+	profile, err := scanAgentProfile(row)
+	if err != nil {
+		return nil, err
+	}
+	return r.applyLegacyBackfill(ctx, profile), nil
+}
```
carlosflorencio, line 778: This SELECT projection is byte-for-byte identical to `GetAgentProfile` at lines 732-756 except for the trailing `AND deleted_at IS NULL`. The next column added to `agent_profiles` will need to land in both spots, and if it only makes it into one, the soft-delete path will silently scan zero values for that field, which is exactly the kind of bug this PR is trying to prevent in a different layer.

The `ListAgentProfiles` query right below at 787 has the same shape too. A `const` extracted once would cover all three:

```go
const agentProfileSelectColumns = `
	SELECT id, agent_id, name, agent_display_name, model, mode, migrated_from,
		auto_approve, dangerously_skip_permissions, allow_indexing,
		cli_passthrough, user_modified, plan, cli_flags,
		COALESCE(env_vars, '[]'),
		created_at, updated_at, deleted_at,
		COALESCE(workspace_id, ''), COALESCE(role, ''), COALESCE(icon, ''),
		COALESCE(reports_to, ''), COALESCE(skill_ids, '[]'),
		COALESCE(desired_skills, '[]'), COALESCE(custom_prompt, ''),
		COALESCE(status, 'idle'), COALESCE(pause_reason, ''),
		last_run_finished_at,
		COALESCE(max_concurrent_sessions, 1), COALESCE(cooldown_sec, 0),
		COALESCE(skip_idle_runs, 0), COALESCE(consecutive_failures, 0),
		COALESCE(failure_threshold, 3), COALESCE(executor_preference, ''),
		COALESCE(budget_monthly_cents, 0),
		COALESCE(settings, '{}'), COALESCE(permissions, '{}')
	FROM agent_profiles`
```

Then `GetAgentProfile` calls `Rebind(agentProfileSelectColumns + " WHERE id = ? AND deleted_at IS NULL")` and this method calls `Rebind(agentProfileSelectColumns + " WHERE id = ?")`. Touching `ListAgentProfiles` too is optional, but folding it in keeps all three in lockstep going forward.

cubic flagged this one too.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/backend/internal/orchestrator/watcher_dispatch_wiring.go
```diff
@@ -38,6 +40,104 @@ func (s *Service) initWatcherCoordinator() {
...
+	s.logger.Warn("github "+kind+": agent profile soft-deleted, self-healing watcher",
+		zap.String("watch_id", watchID),
+		zap.String("profile_id", profileID),
+		zap.String("profile_name", name))
+	if disable != nil {
+		if err := disable(ctx, watchID, cause); err != nil {
+			s.logger.Error("github "+kind+": self-heal disable failed",
+				zap.String("watch_id", watchID), zap.Error(err))
+		}
+	}
```
carlosflorencio, line 148: Tiny one: the three log calls below concatenate `kind` into the message string (`"github "+kind+": ..."`). That gives the log aggregator two separate message strings to group on ("github issue: ..." and "github review: ...") instead of one filterable family. Most of the orchestrator passes that kind of axis as a structured zap field instead.

```suggestion
func (s *Service) preflightDeletedProfileForGitHub(
	ctx context.Context, kind, profileID, watchID string,
	disable func(ctx context.Context, watchID, cause string) error,
) bool {
	lookup := s.getProfileLookup()
	if lookup == nil || profileID == "" {
		return false
	}
	deleted, name, err := lookup.LookupProfile(ctx, profileID)
	if err != nil {
		s.logger.Warn("github watcher: profile lookup failed, falling through",
			zap.String("kind", kind),
			zap.String("profile_id", profileID),
			zap.Error(err))
		return false
	}
	if !deleted {
		return false
	}
	cause := formatDeletedProfileCause(profileID, name)
	s.logger.Warn("github watcher: agent profile soft-deleted, self-healing",
		zap.String("kind", kind),
		zap.String("watch_id", watchID),
		zap.String("profile_id", profileID),
		zap.String("profile_name", name))
	if disable != nil {
		if err := disable(ctx, watchID, cause); err != nil {
			s.logger.Error("github watcher: self-heal disable failed",
				zap.String("kind", kind),
				zap.String("watch_id", watchID),
				zap.Error(err))
		}
	}
	return true
}
```

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/backend/internal/orchestrator/watcher_dispatch_wiring.go
```diff
@@ -38,6 +40,104 @@ func (s *Service) initWatcherCoordinator() {
 		}
 	}
 	s.watcherCoordinator.SetTaskCreator(s.issueTaskCreator)
+	if lookup := s.getProfileLookup(); lookup != nil {
+		s.watcherCoordinator.SetProfileLookup(lookup)
+	}
+}
```
carlosflorencio: A bit of an asymmetric locking situation here that's worth tightening. Commit 0f4824d correctly put `s.profileLookup` behind `s.mu` (read in `getProfileLookup`, write in `SetProfileLookup`). But `s.watcherCoordinator` and `s.issueTaskCreator` are still written without the lock, both in this `initWatcherCoordinator` body and in `SetIssueTaskCreator` over at `event_handlers_github.go:133-136`. Meanwhile `SetProfileLookup` reads `s.watcherCoordinator` under the lock and `dispatchWatcherEvent` (line 161 below) reads it with no lock at all.

In production both setters run sequentially at boot before any bus handlers fire, so today it's fine. But the comment on lines 26-31 explicitly says tests swap creators between scenarios, and that's the path `-race` will eventually catch.

Suggested fix: take `s.mu.Lock()` around the whole body here, and have `SetIssueTaskCreator` set `s.issueTaskCreator` under the same lock before calling this. Inside the locked section, read `s.profileLookup` directly instead of calling `getProfileLookup()` (which re-takes the read lock and would deadlock).

---
_Generated by [Claude Code](https://claude.ai/code)_

carlosflorencio (commented):
Reviewed end to end. The self-heal logic is solid: preflight runs before `Reserve` (so a deleted-profile event never leaves a stale dedup row), the typed `*DeletedProfileError` is consumed via `errors.Is`/`errors.As` everywhere it matters, all three disable-watcher store tests already use a ±1-second window so the timestamp-precision flake CodeRabbit raised earlier is mitigated, and the 409 `watchers` field is additive on the wire. Test coverage matches the body checklist case-for-case.

Four small things worth doing before merge, none of them blockers, all left as inline comments:

1. `github/store.go` swallows real migration errors with `_, _ =`. The new `last_error` / `last_error_at` columns are the only ones whose readers panic on scan if they end up missing, so the fail-loud guard the Jira/Linear stores use is worth borrowing.
2. `GetAgentProfileIncludingDeleted` duplicates the 35-column SELECT projection from `GetAgentProfile`. Worth extracting as a `const` so the next column add can't drift between them.
3. `s.watcherCoordinator` and `s.issueTaskCreator` are written without the mutex that `SetProfileLookup` carefully added in 0f4824d. Production is safe because both setters run at boot, but the asymmetry will catch out a future test under `-race`.
4. The new GitHub preflight logger calls concatenate `kind` into the message string instead of passing it as a zap field, which fragments log aggregation.

(Skipped CodeRabbit's `reconciler_test.go` `fakeStore` finding because it's wrong: the fake delegates to `GetAgentProfile` and returns `sql.ErrNoRows` on miss, not `(nil, nil)`.)

Nice work overall, looking forward to seeing the follow-up that covers the adjacent surfaces (`automations`, `office_routines`, etc.) you called out at the bottom.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/backend/internal/agent/settings/handlers/handlers.go
```diff
@@ -466,8 +466,9 @@ func (h *Handlers) httpDeleteProfile(c *gin.Context) {
 		var inUseErr *controller.ErrProfileInUseDetail
 		if errors.As(err, &inUseErr) {
 			c.JSON(http.StatusConflict, gin.H{
-				"error":           "agent profile is used by active session(s)",
+				"error":           "agent profile is used by active session(s) or watcher(s)",
 				"active_sessions": inUseErr.ActiveSessions,
+				"watchers":        inUseErr.Watchers,
```
carlosflorencio, line 471: The backend now returns `watchers` in the 409 response, but the frontend delete flow still only models active sessions. `deleteAgentProfileAction` only returns `{ status: "conflict", activeSessions }`, and `AgentProfileDeleteConflictDialog` only renders sessions.

That means the new backend behavior does not deliver the intended confirmation UX for the main new case: “this profile is referenced by watchers, continue?” A watcher-only conflict can become a generic failed-delete path or a dialog without any watcher details, depending on the response shape.

Can we carry `watchers` through `DeleteProfileResult`, add a `WatcherReference` type, and render them in `AgentProfileDeleteConflictDialog` alongside active sessions? A small frontend test for a watcher-only 409 would cover the regression.

apps/backend/internal/agent/settings/controller/profile_crud.go
```diff
@@ -213,24 +213,37 @@ func (c *Controller) DeleteProfile(ctx context.Context, id string, force bool) (
 	return &result, nil
 }
 
-// prepareProfileDeletion checks for active sessions and cleans up ephemeral tasks before deletion.
+// prepareProfileDeletion checks for active sessions and referencing watchers,
+// then cleans up ephemeral tasks. Returns *ErrProfileInUseDetail when force
+// is false and any active session OR referencing watcher exists — the UI uses
+// the breakdown to render a confirmation dialog. Force=true skips both checks
+// and lets the dispatch coordinator's pre-flight handle the watcher self-heal
```
carlosflorencio: This force-delete path does not actually disable the referencing watchers. It only skips the dependency check and relies on “the dispatch coordinator’s pre-flight handle the watcher self-heal on its next poll,” but that pre-flight only runs after a watcher emits a new unseen issue/PR event.

If the watcher has no new matching external item after the profile is deleted, it stays enabled indefinitely while pointing at a soft-deleted profile. That leaves the UI and stored watcher state inconsistent with the confirmation copy.

I think the forced delete path should disable the currently referencing watchers immediately, using the same `Disable*WithError` methods added in this PR, and stamp the deleted-profile cause there. The self-heal preflight can remain as a fallback for profiles deleted by reconciler/other paths.

carlosflorencio:
Thanks for working on this! Added some comments to further improve the fix 💪

nlenepveu:
Thanks for the thorough review @carlosflorencio — your verdict checklist is exactly the maintainer bar I was hoping to hit. Folding all six findings into this PR:

- **(1) github migration precheck** — adding `tableColumns` + `addWatchSelfHealColumns` to mirror the jira/linear idiom. The reader-side scan panic is exactly why these new columns deserve the fail-loud treatment even though their cleanup_policy neighbour gets the silent-swallow.
- **(2) `agentProfileSelectColumns` const** — extracting once; covers `GetAgentProfile`, `GetAgentProfileIncludingDeleted`, and `ListAgentProfiles` in lockstep.
- **(3) zap field for `kind`** — applying your suggestion verbatim. Agreed on the log-aggregation argument.
- **(4) lock symmetry** — wrapping `initWatcherCoordinator` body under `s.mu`, moving the `issueTaskCreator` assignment under the same lock in `SetIssueTaskCreator`, reading `profileLookup` directly inside the locked section (avoids the read-lock re-entry deadlock you flagged).
- **(5) frontend dialog** — pulling into this PR rather than deferring. Threading `watchers` through `deleteAgentProfileAction` + `AgentProfileDeleteConflictDialog` + adding the watcher-only 409 frontend test.
- **(6) force-delete eager disable** — calling `Disable*WithError` directly in `prepareProfileDeletion` for any watchers we just enumerated, with the deleted-profile cause. The lazy preflight stays as the safety net for reconciler-driven deletes.

Pushing the round shortly.

apps/backend/internal/github/store.go
nlenepveu: Addressed in 07fdf0d7 — added `tableColumns` + `addWatchSelfHealColumns` to github/store.go and replaced the four `_, _ = ALTER` calls. Idempotent precheck on `github_review_watches` + `github_issue_watches`; mirrors the jira/linear shape, with the boot now failing loud rather than silently dropping a driver error that the unconditional readers would later turn into a scan panic.

apps/backend/internal/agent/settings/store/sqlite.go
nlenepveu: Addressed in 07fdf0d7 — extracted `agentProfileSelectColumns` const at the top of `internal/agent/settings/store/sqlite.go`. All three call sites (GetAgentProfile, GetAgentProfileIncludingDeleted, ListAgentProfiles) now use `Rebind(agentProfileSelectColumns + " WHERE ...")`. Next column add lands in one spot.

apps/backend/internal/orchestrator/watcher_dispatch_wiring.go
nlenepveu: Addressed in 07fdf0d7 — `kind` is now passed as `zap.String("kind", kind)` in all three log calls. Message strings collapse to "github watcher: profile lookup failed, falling through" / "github watcher: agent profile soft-deleted, self-healing" / "github watcher: self-heal disable failed" so the aggregator groups them as one filterable family.

apps/backend/internal/orchestrator/watcher_dispatch_wiring.go
nlenepveu: Addressed in 07fdf0d7 — renamed to `initWatcherCoordinatorLocked` (callers MUST hold `s.mu` write). `SetIssueTaskCreator` now takes the lock, writes `s.issueTaskCreator`, then calls the locked variant. Inside the locked section we read `s.profileLookup` directly (no `getProfileLookup()` re-entry deadlock). `dispatchWatcherEvent` reads both fields via new `getIssueTaskCreator()` / `getWatcherCoordinator()` RLock helpers. Race-detector tests still pass.

apps/backend/internal/agent/settings/handlers/handlers.go
nlenepveu: Addressed in 07fdf0d7 (+ formatting in e31bfaaa). Threaded `watchers` through: new `WatcherReference` type in `agent-profile-errors.ts`, `DeleteProfileResult` carries both lists, `AgentProfileDeleteConflict` carrier prop on `AgentProfileDeleteConflictDialog` renders watchers grouped by integration kind (Linear / Jira / GitHub Issues / GitHub PR Reviews). 409 trigger now fires on either `active_sessions` OR `watchers` non-empty so a watcher-only conflict pops the dialog. New `agent-profile-delete-dialog.test.tsx` covers all four shapes (watcher-only / sessions-only / combined / null).

apps/backend/internal/agent/settings/controller/profile_crud.go
nlenepveu: Addressed in 07fdf0d7 — `WatcherDependencyChecker` grew `DisableWatchersByAgentProfile(ctx, profileID, cause)`. When `prepareProfileDeletion` runs with `force=true`, the new `disableReferencingWatchers` helper invokes it and logs the disable count. The cmd/kandev adapter dispatches to the per-integration `Disable*WithError` store methods (linear / jira / github_issue / github_review). The lazy preflight stays as the safety net for reconciler-driven deletes that don't pass through this controller. Test covers the force-path eager disable.

carlosflorencio:
There are some conflicts now, let me know if you want me to address them. Nice work 💪

carlosflorencio:
I went back through this carefully and checked each of the six change requests against the current head. All six are genuinely addressed:

- the ALTER errors are now handled via addWatchSelfHealColumns with a column precheck
- the duplicated SELECT is extracted into agentProfileSelectColumns
- the log now uses zap.String for kind instead of string concatenation
- the locking is fixed with initWatcherCoordinatorLocked plus the RLock helpers
- the 409 now threads watchers through WatcherReference into the delete dialog
- force delete disables referencing watchers via DisableWatchersByAgentProfile

Test coverage is good. The only thing left is the rebase you already mentioned, since the branch is still conflicting with main and the failing E2E shards are tied to the stale base. Once it is rebased and CI is green this is good to go.

Two minor optional notes: the eager-disable last_error string omits the profile name while the lazy path includes it, and force delete disables watchers just before the row delete, so a failed delete would leave them disabled. Both are low risk.

nlenepveu:
Rebased onto latest `main` and resolved all conflicts — the branch is now `MERGEABLE` again. The conflicts were all in the watcher/integration files where upstream's `max_inflight_tasks` throttle work (#1021) had landed in the same structs/migrations/SQL this PR touches; both features are integrated side-by-side:

- **`jira`/`linear` stores + models**: `max_inflight_tasks` and `last_error`/`last_error_at` now coexist in the struct, the insert/select column split, and the idempotent migrations (`addMaxInflightTasksColumn` + `addIssueWatchLastErrorColumns` both run).
- **`github/service.go`**: upstream split it into `service_issues.go` / `service_reviews.go`; moved the four new wrapper methods (`ListEnabled*Watches`, `Disable*WithError`) into the right split files.
- **`WatcherSource` interface**: now carries upstream's `WatchID`/`MaxInflightTasks` *and* this PR's `SelfHeal`; the sources and test fakes implement all three.
- **`dispatchWatcherEvent`**: keeps upstream's per-watcher throttle (`acquireWatcherSlot` + `defer release()`) layered on top of this PR's thread-safe `getWatcherCoordinator()` accessor.

Also folded in your two optional notes:

1. **Eager-disable cause now includes the profile name** — `disableReferencingWatchers` uses a `formatDeletedProfileCause` helper (80-rune cap, matching the preflight) so the banner shows `agent profile "Kilo Profile" (id) was deleted` instead of a bare UUID. Pinned with a test assertion.
2. **Force-delete disables watchers *after* the row delete succeeds** — moved out of `prepareProfileDeletion` into `DeleteProfile` post-delete, so a failed delete no longer strands watchers disabled against a still-live profile; the preflight remains the fallback if the post-delete disable itself fails.

Local gates all green: `make fmt/build/vet/lint` (0 issues) + full `go test`, and frontend `typecheck`/`lint`/`test` (2774 passing). CI re-running now.

nlenepveu:
CI is fully green now ✅ — all 10 E2E shards pass, including Shard 5/10 which was the only one still red after the rebase.

That shard turned out to be a **pre-existing bug in this PR's own e2e helper**, not the stale base: `getLinearIssueWatch` called `.find()` directly on the `GET /api/v1/linear/watches/issue` response, but that endpoint returns a `{ watches: [...] }` envelope (it always has — identical on the pre-rebase branch), so it threw `TypeError: watches.find is not a function`. Fixed by destructuring the envelope (commit `c0ba71f8`); verified locally with `pnpm e2e:run --host` (1 passed) and now confirmed green in CI.

Final state: rebased onto latest `main`, `MERGEABLE`, all checks passing, both optional notes folded in. Good to go from my side.

nlenepveu:
Heads up: one CI check came back red after the last push — **Run Backend Tests** (run 26719846305) — but it's a **goroutine-leak flake, not a regression**:

```
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 982 ... agentctl.(*Client).StreamWorkspace ...
 Goroutine 959 ... net/http.(*Server).Serve ...
 Goroutine 983 ... agentctl.(*WorkspaceStream).writeLoop ...
```

- It's in `internal/agent/runtime/lifecycle` / `agentctl` (streaming teardown) — a package this PR **doesn't touch** (all changes are in watcher/integration/settings).
- goleak reports *"on successful test run"* — every test passed; it just caught async stream/server goroutines that hadn't drained at `TestMain`.
- The same suite is green locally (`make test`, 0 failures) and passed in the earlier CI run for this commit.

I don't have admin rights to re-run the job — could you kick off a re-run of the failed Backend Tests job? It should go green. Everything else (lint, Build, Frontend, Backend windows, all 10 E2E shards incl. the previously-failing 5/10) is passing.

carlosflorencio:
Rebased with main, the backend tests seem to be failing.

nlenepveu:
Backend tests are green on `4b4d2f77` — CI's Backend Tests job passed and the full suite runs clean locally (`go test -tags fts5 ./...`, 0 failures). PR is CLEAN/MERGEABLE and ready to merge.

PR #1122, opened 2026-05-28
fix(web): keep markdown table headers readable on narrow content

carlosflorencio:
Long inline-code identifiers in a table cell (e.g. a fully-qualified Go test name) were starving the label column and forcing header words to wrap one character per line ("Failin g test"). Headers now stay on one line, body cells break long tokens, and inline code inside cells no longer behaves as one unbreakable atom — so dense agent-output tables stay readable.

## Important Changes

- \`apps/web/app/globals.css\` — \`th { white-space: nowrap }\`, \`td { overflow-wrap: anywhere; word-break: break-word; vertical-align: top }\`, plus an override for \`td/th code\` so inline-code no longer pins as a single atom. Existing \`overflow-x-auto\` wrapper retained as a fallback for unusually wide tables.
- \`apps/backend/cmd/mock-agent/scenarios.go\` — new \`markdown-table\` scenario that reproduces the original ugly table (narrow labels + long inline-code values), triggered with \`/e2e:markdown-table\` in dev/mock mode.

## Validation

- \`go build ./cmd/mock-agent/\` — passes.
- \`go test ./cmd/mock-agent/ -count=1\` — 134 tests pass.
- \`pnpm --filter @kandev/web lint\` — clean.
- Manual: \`/e2e:markdown-table\` against mock agent → headers read as words, long test names wrap inside their cell, columns size sensibly.

## Possible Improvements

Low risk — CSS-only behavior change in agent-rendered markdown. Worst case is that very wide tables with many columns may still trigger horizontal scroll via the existing \`overflow-x-auto\` wrapper, which is the intended fallback.

## Checklist


carlosflorencio:
Both Greptile and Claude flagged the same two CSS issues — addressed in f0725004:

- **`th code` defeating `th { white-space: nowrap }`**: scoped the override to `.markdown-body td code` only, so inline code inside a header cell now inherits the parent's `nowrap`.
- **Dead `overflow-wrap` / `word-break` on `th`**: moved those properties out of the shared `th, td` block into a `td`-only block, so the shared rule no longer carries inert properties.

cubic reported no issues. CodeRabbit was rate-limited — happy to retrigger if needed.

PR #1197, opened 2026-05-31
fix(web): stop changes panel flicker and restore planning chat streaming

carlosflorencio:
The changes panel was reloading on duplicate git status snapshots that only
differed by timestamp, and planning-mode chat could stall when session-scoped
message frames were missed while metadata events kept arriving. Git status
updates now compare meaningful fields only, and the session message hook
backfills while an active turn is running after a user prompt.

## Important Changes

- Compare git status snapshots by files, stats, and branch summary instead of timestamp alone.
- Poll for new messages during RUNNING turns and refetch when the active turn completes.

## Validation

- `make fmt`
- Backend `test` and `lint`
- `pnpm --filter @kandev/web lint typecheck test`
- Focused unit tests for git status dedup and session message backfill guards

## Possible Improvements

Low risk — backfill polling is gated to connected RUNNING sessions with an active-turn user prompt; watch for extra fetch load on long-running turns.

## Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/web/hooks/domains/session/use-session-messages.test.ts
+++ b/apps/web/hooks/domains/session/use-session-messages.test.ts
@@ -29,7 +29,6 @@ afterEach(() => {
 import {
   hasUserPromptInActiveTurn,
   hasUserOrAgentMessage,
-  isActiveTurnCompletion,
   isTurnSettleTransition,
   shouldRunMessageBackfill,
   runBackfillRound,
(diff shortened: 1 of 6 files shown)


carlosflorencio:
Addressed all 6 review threads in the latest push:

- Removed `useResyncOnActiveTurnComplete` — turn-end refetch is handled solely by `useResyncOnTurnSettle`, eliminating duplicate `message.list` calls at turn boundaries.
- Added an in-flight guard to `useRunningMessageBackfill` so slow fetches cannot pile up on the 3s interval.
- Introduced shared `gitStatusWouldMutate` used by both the WS handler and `setGitStatus`; multi-repo duplicate sibling snapshots no longer overwrite `byEnvironmentId` or invalidate the diff cache.
- Documented `hasFileStatsChanged` as a fast early-exit before the per-file deep comparison.

PR #1216, opened 2026-06-01
fix(backend): cap gh/git fork concurrency via shared subproc throttles

carlosflorencio:
macOS hosts with CrowdStrike Falcon froze when kandev burst dozens of concurrent `gh` and `git` subprocesses because `syspolicyd` serialises every fork/exec; routing every invocation through process-wide semaphores (gh cap 8, git cap 12) keeps the EDR from queueing and the backend stays responsive.

## Important Changes

- New `internal/common/subproc` package owns the shared `gh` and `git` throttles plus `RunGH*` / `RunGit*` helpers (caps overridable via `KANDEV_GH_MAX_CONCURRENT` and `KANDEV_GIT_MAX_CONCURRENT`).
- All `gh` and `git` execs across worktree, github, repoclone, improvekandev, orchestrator executor, agent runtime lifecycle, and agentctl process now flow through the throttle; the streaming `git diff` path in agentctl uses explicit `Acquire`/`Release` so long-lived `Start`/`Wait` pairs also count.
- Batched the PR-watch GraphQL fetches and coalesced overlapping workspace refreshes with `singleflight` to cut `gh` fork volume at the source.
- Threaded `context.Context` through `WorkspaceTracker.ApplyFileDiff` so cancellation propagates into throttled execs.

## Validation

- `make fmt`
- `make lint` (0 issues)
- `make test` (full backend suite, including new `subproc`, `gh_throttle`, `git_throttle`, and singleflight regression tests, all green with `-race`)

## Possible Improvements

Low risk: caps are conservative defaults and can be tuned via env vars; the main behavioural change is that bursts of git/gh work now serialise once 8/12 are in flight, which should be invisible outside extreme parallelism.

## Checklist

### DIFF
--- a/apps/backend/internal/agent/handlers/git_handlers.go
+++ b/apps/backend/internal/agent/handlers/git_handlers.go
@@ -5,14 +5,17 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
+
+	"go.uber.org/zap"
+	"golang.org/x/sync/singleflight"
 
 	"github.com/kandev/kandev/internal/agent/runtime/agentctl"
 	"github.com/kandev/kandev/internal/agent/runtime/lifecycle"
 	"github.com/kandev/kandev/internal/common/logger"
 	ws "github.com/kandev/kandev/pkg/websocket"
-	"go.uber.org/zap"
 )
 
 // PRCreatedCallback is called after a PR is successfully created. `repo` is
@@ -50,12 +53,20 @@ type SessionReader interface {
 
 // GitHandlers provides WebSocket handlers for git worktree operations.
 // Operations are executed via agentctl which runs in the worktree context.
+//
+// commitsGroup / diffGroup deduplicate concurrent identical requests
+// (same session, same params) so a frontend re-render storm doesn't fan
+// out into N agentctl HTTP calls that each spawn a git subprocess under
+// the agentctl-side throttle. Reproduced in PR #1216 trace: ~10
+// session.git.commits arrived for the same sessionID inside 30ms.
 type GitHandlers struct {
 	lifecycleMgr         ExecutionLookup
 	sessionReader        SessionReader
 	logger               *logger.Logger
 	onPRCreated          PRCreatedCallback
 	onGitOperationFailed GitOperationFailedCallback
+	commitsGroup         singleflight.Group
+	diffGroup            singleflight.Group
 }
 
 // NewGitHandlers creates a new GitHandlers instance.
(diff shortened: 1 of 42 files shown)


carlosflorencio:
Addressed the bot review feedback in `a4df519f`.

**Fixed:**

- **P1 (Greptile/cubic): poller publishes feedback on failed `SyncTaskPR`** — added `SyncFailed bool` to `PRWatchSyncResult` and gated the publish loop on it so a stale `task_pr` row can no longer trigger a "merged" / "checks changed" event.
- **P1/P2 (cubic): `refreshStaleTasksPerTask` returns before goroutines finish** — added `sync.WaitGroup` and a `select` on `ctx.Done()` for the semaphore send so `inflightWorkspaceRefreshes` can't be cleared while writes are still happening.
- **Major (CodeRabbit): `ApplyFileDiff` context-cancel could overwrite user edits** — `ctx.Canceled` / `DeadlineExceeded` now short-circuits before the `writeDesiredContent` fallback.
- **Major/P2 (all bots): throttle queue time eating operation timeouts** — split `acquireCtx` from `execCtx` in `DetectGHToken` (lifecycle + orchestrator), `readCurrentBranchForLocal`, `resolveGitIndexPath`, and the worktree manager `branchExists` / `currentBranch` / `pullBaseBranch` / `pullCurrentBranchOrFallback` / `fetchBranchToLocal`. Throttle Acquire uses the parent (or a 30s budget); the exec keeps its short timeout.
- **Major (CodeRabbit): `parseGitStatusOutput` bypass** — the hottest polling path now goes through `subproc.RunGitOutput`.
- **Low (Claude/cubic): explicit `release()` in `capDiffOutput`** — switched to `defer release()`.

**Skipped (with reasoning):**

- **Claude: unowned `refreshStaleWorkspaceWatches` goroutine** — agreed it should own a `WaitGroup` per `apps/backend/AGENTS.md`, but the goroutine lifecycle is tied to `Service` itself and the cleanup needs to thread through `Service.Stop`. Worth a dedicated PR.
- **Claude/CodeRabbit: `testing/synctest` in throttle tests** — these are pure semaphore tests with millisecond budgets; would be a behavior-neutral test refactor in this PR's footprint. Filing as a follow-up.
- **CodeRabbit: redundant `UpdatePRWatchTimestamps` in `applyBatchedSearchingWatch`** — harmless duplicate, low value, skipping.
- **Claude: pre-existing `io.ReadAll` error discard in `capDiffOutput`** — explicitly pre-existing, out of scope for this PR.

Verified with `make fmt`, `make lint` (0 issues), `go test -race` on all affected packages, and `make test` (full backend suite, all green).

carlosflorencio:
Round 2 of fixes pushed in `f70944d1`.

**Newly addressed:**

- **Cubic P1 (re-raised on `f70944d1`'s parent): exec ctx built before throttle acquire** — applied Option C (Acquire first, build exec ctx after slot is held) in the four 5s-budget probes: `DetectGHToken` (lifecycle + orchestrator), `readCurrentBranchForLocal`, `resolveGitIndexPath`. Previous round derived `execCtx` from `acquireCtx` at creation time, so the 5s clock kept ticking during queue wait. Now the 5s deadline starts at `cmd.Run()` instead.

**Skipped (with reasoning):**

- **Cubic P2 (`manager_git.go`): passing parent ctx removes per-step bounds on throttle wait.** The worktree manager's parent ctx is the orchestrator's task ctx, which already carries a deadline; combined with the 15s/25s inspect/fetch/pull budgets and cap=12, realistic throttle queue time (sub-second) is well inside the noise. Adding the full Acquire-then-exec-ctx dance here would be code churn for a theoretical edge case. The behavior matches the documented `branchExists` contract ("could not tell" on transient failure).
- **Cubic P2 (`throttle.go:70` Acquire race):** Standard Go `select` semantics — if both `sem <- struct{}{}` and `<-ctx.Done()` are ready, the runtime picks one uniformly. The "spurious" success is harmless: the caller acquires a slot, then is free to check `ctx.Err()` before doing work. Re-checking inside `Acquire` would only paper over caller-side logic that's already responsible for ctx propagation.
- **Cubic/Greptile P2 (`service_pr_watch.go:411` fallback ctx canceled):** Stale — that path now waits on `wg.Wait()` before returning, so `syncCtx` (the 60s `WithTimeout` from `refreshStaleWorkspaceWatches`) isn't canceled until all spawned goroutines have finished writing.
- **Claude (still open): `refreshStaleWorkspaceWatches` unowned goroutine.** Acknowledged from round 1. Wiring this into a proper `Service.Stop` shape touches the broader service lifecycle and belongs in its own PR.
- **CodeRabbit/Claude: `synctest` in throttle tests.** Behavior-neutral test refactor, filing as follow-up rather than expanding this PR's footprint.
- **Claude: discarded release closures in `setGitThrottleCapForTest` tests.** Safe because the test seam replaces the entire channel; happy to revisit if `SetCapForTest` ever changes to drain-and-resize.
- **Claude: redundant `UpdatePRWatchTimestamps` in `applyBatchedSearchingWatch`.** Harmless duplicate, low value.

Verified with `make fmt`, `make lint` (0 issues), `go test -race` on all affected packages.

PR #1274, opened 2026-06-04
fix(repo): exclude archived tasks from repository delete guard

nlenepveu:
## Problem

Deleting a repository fails with *"still linked to active tasks"* even when no task is active. The guard (`HasActiveTaskSessionsByRepository` / `CountActiveTaskSessionsByRepository` in `task/repository/sqlite/session.go`) blocks deletion whenever a linked `task_session` is in an active state (`CREATED`/`STARTING`/`RUNNING`/`WAITING_FOR_INPUT`).

The trap: when an agent finishes a turn it parks the session in `WAITING_FOR_INPUT` indefinitely. **Archiving the task does not transition its sessions out of that state** — archival only sets `tasks.archived_at`/`state`. Since the guard never filtered on `archived_at`, a long-finished, archived task kept its repository permanently undeletable. The UI label says "active tasks", but the real predicate is "active task *sessions*", and archived tasks are invisible on the board — so users see "no active tasks" yet cannot delete.

Observed on a real instance: 124 sessions stuck in `WAITING_FOR_INPUT` (110 on archived tasks), 0 actually running, blocking 12 repositories.

## Fix

Join `tasks` and add `AND t.archived_at IS NULL` to both guard queries, mirroring the existing idiom in `session_branch.go`. Archived tasks no longer block deletion; live (non-archived) tasks with resumable sessions still block it exactly as before.

## Tests

- `TestCountActiveTaskSessionsByRepository_ExcludesArchivedTasks` — archived task's active session does not count; a live task on the same repo still does (guards against over-broadening).
- `TestHasActiveTaskSessionsByRepository_ExcludesArchivedTasks` — boolean guard mirrors the count.

Verified red before the fix, green after.

```
make -C apps/backend fmt   # clean
go vet ./internal/task/repository/sqlite/   # clean
go build ./...   # clean
go test -race -count=1 ./internal/task/repository/sqlite/   # ok
golangci-lint run ./internal/task/repository/sqlite/...   # 0 issues
```

## Note / possible follow-up

This fixes the guard. A deeper option (out of scope here to keep the change minimal) is to have task archival reap its sessions into a terminal state, so archived tasks never retain `WAITING_FOR_INPUT` rows at all.

### DIFF
--- a/apps/backend/internal/task/repository/sqlite/session.go
+++ b/apps/backend/internal/task/repository/sqlite/session.go
@@ -779,12 +779,16 @@ func (r *Repository) HasActiveTaskSessionsByEnvironment(ctx context.Context, env
 
 func (r *Repository) HasActiveTaskSessionsByRepository(ctx context.Context, repositoryID string) (bool, error) {
 	var exists int
+	// Only sessions of live (non-archived) tasks count; archived tasks never
+	// block repository deletion.
 	err := r.ro.QueryRowContext(ctx, r.ro.Rebind(`
 		SELECT 1
 		FROM task_sessions s
 		INNER JOIN task_repositories tr ON tr.task_id = s.task_id
+		INNER JOIN tasks t ON t.id = s.task_id
 		WHERE s.state IN ('CREATED', 'STARTING', 'RUNNING', 'WAITING_FOR_INPUT')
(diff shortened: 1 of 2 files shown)


carlosflorencio:
Nice!

PR #1282, opened 2026-06-06
fix(web): point sidebar Home to office dashboard in office mode

zeval:
The AppSidebar **Home** item always navigated to the Kanban view (`/`), even when the user was in office mode — so clicking Home from an `/office/*` route dropped the user out of Office. Home is now office-aware: it targets the office dashboard (`/office`) while on any office route and the board (`/`) otherwise.

## Important Changes

- `app-sidebar-primary-nav.tsx`: Home `href` is now `inOffice ? "/office" : "/"` (the `useInOffice()` hook was already computed but unused). `exactMatch` is retained, so in office mode Home highlights only on the dashboard, not on `/office/inbox`/`/office/tasks`.

## Validation

- `pnpm --filter @kandev/web typecheck lint test` — all pass
- `make build-web && make build-backend`
- `pnpm e2e tests/office/sidebar-navigation.spec.ts` — 6 passed, including the two new cases (Home → `/office` dashboard from an office route; Home → `/` board from a regular route)

## Checklist

- [ ] Tests added/updated
- [ ] Docs updated if needed
- [ ] Self-reviewed the diff

### DIFF
--- a/apps/web/components/app-sidebar/app-sidebar-primary-nav.tsx
+++ b/apps/web/components/app-sidebar/app-sidebar-primary-nav.tsx
@@ -19,7 +19,13 @@ export function AppSidebarPrimaryNav({ collapsed }: AppSidebarPrimaryNavProps) {
 
   return (
     <div className="flex flex-col gap-0.5">
-      <AppSidebarNavItem icon={IconHome} label="Home" href="/" collapsed={collapsed} exactMatch />
+      <AppSidebarNavItem
+        icon={IconHome}
+        label="Home"
+        href={inOffice ? "/office" : "/"}
+        collapsed={collapsed}
+        exactMatch
+      />
       {inOffice && (
         <AppSidebarNavItem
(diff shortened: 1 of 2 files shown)


zeval:
Thanks @claude — addressed the review:

1. **No-op navigation in the Kanban test** — fixed in 4c770748d. The test now starts from `/settings/system/status` (a non-home regular route) before clicking Home, so the assertion confirms a real navigation back to `/` and the board renders. Re-ran locally: both Home-destination tests pass.

2. **Unused `officeSeed` in the non-office test** — keeping as-is intentionally. The `testPage` fixture in `office-fixture.ts` already depends on `officeSeed` (it resets/activates the seeded office workspace), so the onboarding seed runs regardless of whether the test names the fixture. Dropping `officeSeed: _` would remove no setup cost, and every sibling test in this file uses the same parameter for consistency.

zeval:
@coderabbitai review

PR #1286, opened 2026-06-07
fix(web): preserve manual task selection during archive/delete

jcfs:
## Summary
- Fix race where archiving or deleting a task could clobber the user's manual navigation to another task during the in-flight API call.
- `resolveActiveTaskId` always preferred the caller-captured pre-API `wasActiveTaskId` over the live store value, so if the user clicked another task while archive/delete was still resolving, the auto-switch logic still treated the removed task as "active" and jumped to `allRemainingTasks[0]` — overwriting the user's pick.
- Replaced with `shouldSwitchAfterRemoval`, a two-condition guard that only auto-switches when (a) the store still points at the removed task, or (b) the store has been WS-cleared to `null` **and** the captured value matches the removed task.

## Implementation notes
The captured `wasActiveTaskId` was originally added to survive the WS `task.deleted` / `task.updated(archived_at)` handler clearing `activeTaskId` to `null` before `removeTaskFromBoard` ran. That protection is still needed, but it shouldn't override a non-null store value — a non-null store value that differs from the removed task means the user deliberately moved on.

`shouldSwitchAfterRemoval` reads the live `activeTaskId` and only honors `opts.wasActiveTaskId` when the live value has been cleared to `null`. `resolveOldEnvId` is unchanged — once we decide to switch, we still need the pre-WS-clear `activeSessionId` to compute `oldEnvId`. No call-site changes required: the six callers (archive sidebar, archive PR-merged banner, archive action-message, sheet archive, sidebar delete, sheet delete) all benefit transparently.

## Test plan
- [x] New `apps/web/hooks/use-task-removal.test.ts` covering the five guard cases:
  1. `activeTaskId === taskId` → switch to next task.
  2. `activeTaskId === otherId` (manual move) → no switch, URL untouched.
  3. `activeTaskId == null`, `wasActiveTaskId === taskId` (WS race) → switch.
  4. `activeTaskId == null`, `wasActiveTaskId !== taskId` → no switch.
  5. No remaining tasks + still on removed task → redirect to `/`.
- [x] `pnpm --filter @kandev/web test -- use-task-removal` — passes.
- [x] `pnpm --filter @kandev/web typecheck` / `lint` — clean.
- [ ] CI to re-run existing `archive-task-redirect.spec.ts`, `delete-task-redirect.spec.ts`, `sidebar-delete-confirm.spec.ts` — same paths exercised, expected green.

## Risks & rollback
- **Risk**: a caller relies on the legacy "switch even after the user moved on" behavior. Audited all six call sites — every one describes "switch *if the removed task was active*", not "always switch".
- **Risk**: a `setActiveTask(null)` path outside the WS handler trips the `wsCleared` branch incorrectly. None found; `activeTaskId` only changes via user navigation or the two WS handlers.
- **Rollback**: revert the single hook change; the new test file can stay deleted. No data, API, or schema changes.

### DIFF
--- a/apps/web/hooks/use-task-removal.test.ts
+++ b/apps/web/hooks/use-task-removal.test.ts
@@ -1,0 +1,212 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { renderHook } from "@testing-library/react";
+import type { StoreApi } from "zustand";
+
+const replaceTaskUrlMock = vi.fn();
+const performLayoutSwitchMock = vi.fn();
+const listTaskSessionsMock = vi.fn();
+
+vi.mock("@/lib/links", () => ({
+  replaceTaskUrl: (...args: unknown[]) => replaceTaskUrlMock(...args),
+}));
+
+vi.mock("@/lib/state/dockview-store", () => ({
+  performLayoutSwitch: (...args: unknown[]) => performLayoutSwitchMock(...args),
+}));
+
+vi.mock("@/lib/api", () => ({
+  listTaskSessions: (...args: unknown[]) => listTaskSessionsMock(...args),
+}));
+
+import { useTaskRemoval } from "./use-task-removal";
+
+type TaskRow = { id: string; primarySessionId: string | null };
+
+type FakeState = {
+  tasks: { activeTaskId: string | null; activeSessionId: string | null };
+  kanban: { tasks: TaskRow[] };
+  kanbanMulti: { snapshots: Record<string, { tasks: TaskRow[] }> };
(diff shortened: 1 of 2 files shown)


jcfs:
Addressed both non-blocking suggestions from @claude in d5df3a14:

1. `shouldSwitchAfterRemoval` now uses strict `=== null` for consistency with the surrounding strict-equality comparison.
2. Added a unit test documenting the no-opts caller path — when the WS handler has already cleared `activeTaskId` to `null` and no `opts.wasActiveTaskId` is supplied, the auto-switch is intentionally skipped.

PR #1290, opened 2026-06-07
fix(web): tighten dockview topbar button sizing

zeval:
After the sidebar merge shrank the dockview topbar to `h-10`, the right-side action buttons looked cramped — too tall for the bar with uneven top/bottom padding, and the editors button group sat taller than its neighbors. This restores balanced, tasteful spacing.

## Validation
- `pnpm exec prettier` + web lint (pre-commit hooks, passed)
- Visual check of the dockview topbar

## Checklist
- [ ] I have performed a self-review of my code
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have made corresponding changes to the documentation

### DIFF
--- a/apps/web/components/task/editors-menu.tsx
+++ b/apps/web/components/task/editors-menu.tsx
@@ -45,7 +45,7 @@ export function EditorsMenu({ activeSessionId }: EditorsMenuProps) {
   }, [defaultEditorId, enabledEditors]);
 
   return (
-    <div className="inline-flex rounded-md border border-border overflow-hidden">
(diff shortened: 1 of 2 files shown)


zeval:
Fixed in b3ce535b6 — added `focus-visible:ring-inset` to both segmented editor buttons so their 2px ring renders inside the bounds and is no longer clipped by the container's `overflow-hidden` (kept for corner rounding). Focus indicator now fully visible top/bottom. Thanks — identified by cubic.

PR #1294, opened 2026-06-07
fix(web): gate changes push button on ahead count

carlosflorencio:
The Changes panel could offer a Push action for commits that were already present on the tracked remote because the button fell back to the local commit-list count. This makes the action depend only on git's ahead count so pushed PR commits remain reviewable without showing a misleading push affordance.

## Validation

- `make fmt`
- `make typecheck test lint` (initial run hit a transient CLI `src/ports.test.ts` port-availability failure after backend and web tests passed)
- `pnpm -C apps/cli test`
- `make lint`

## Possible Improvements

Low risk: this only changes the Push affordance gating/count; commit rows still show their existing pushed/unpushed state independently.

## Checklist

- [ ] I have tested these changes locally
- [ ] I have added or updated tests where appropriate
- [ ] I have updated docs where needed

### DIFF
--- a/apps/web/components/task/changes-panel-repo-groups.tsx
+++ b/apps/web/components/task/changes-panel-repo-groups.tsx
@@ -197,7 +197,7 @@ export function CommitsGroupActions({
 
 /**
  * Per-repo group inside the Commits section. Like {@link RepoGroupItem} but
- * for commit rows; surfaces a Push button when the repo has unpushed commits.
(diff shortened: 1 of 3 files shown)


carlosflorencio:
Addressed Greptile's stale JSDoc note in f58b7898 by updating the commits group comment to describe the ahead-of-remote Push affordance.

PR #1295, opened 2026-06-07
fix(web): use recent task after removal

carlosflorencio:
Deleting or archiving the active task could jump to an old sidebar task because removal fallback used snapshot order. This now preserves the user's recent-task context and only falls back to snapshot order when no recent remaining task exists.

## Important Changes

- Select the next task after delete/archive from stored recent-task history before using sidebar snapshot order.
- Add a regression test covering a newer recent task appearing after an older task in snapshot order.

## Validation

- `pnpm --filter @kandev/web test -- --run hooks/use-task-removal.test.ts`
- `make fmt typecheck test lint`

## Possible Improvements

The fallback still depends on snapshot order when local recent-task history is empty, which preserves existing behavior for first-time sessions.

## Checklist

- [ ] Tests updated
- [ ] Documentation updated, if needed
- [ ] Screenshots or recordings added, if UI changed

### DIFF
--- a/apps/web/hooks/use-task-removal.test.ts
+++ b/apps/web/hooks/use-task-removal.test.ts
@@ -19,6 +19,7 @@ vi.mock("@/lib/api", () => ({
 }));
 
 import { useTaskRemoval } from "./use-task-removal";
+import { setRecentTasks } from "@/lib/recent-tasks";
 
 type TaskRow = { id: string; primarySessionId: string | null };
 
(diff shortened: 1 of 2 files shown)


carlosflorencio:
Addressed Claude's coverage suggestion in e382e117 by adding a regression where the removed active task is first in recent history and the selector still switches to the next recent remaining task.

PR #1308, opened 2026-06-08
fix(web): keep dockview group alive on task switch to stop layout collapse

zeval:
## Summary

Switching between tasks could corrupt the dockview layout — the agent/chat panel got merged into the terminal group and the grid collapsed into a vertical stack (captured from a real broken DOM: root `dv-vertical`, top group Files/Changes, bottom group Terminal + Opus). The corrupted layout was then persisted per-env, so it survived reloads and got worse with each switch. Root cause is in the fast-path env switch: the incoming session chat was added only *after* the outgoing chat was removed, so when that chat was the last panel in its group the group was emptied and destroyed mid-swap.

## Important Changes

- `tryFastEnvSwitch` now adds the incoming session panel **before** removing the outgoing one, anchored at the outgoing panel's current index, so its group never goes empty (and so is never destroyed) during the swap. Once the group dies, the captured group id is stale and — post unified-sidebar (#1165) — the old `referenceGroup: "sidebar"` fallback in `addIncomingSessionPanel` is dead, leaving an undefined position that dockview resolves by dumping the panel into the active group and collapsing the grid.
- Removed `computeSurvivingIndex`: removing panels ahead of the insert point shifts the new panel into the same final slot, so the survivor-index math is no longer needed.

## Validation

- `pnpm --filter @kandev/web test` — 396 passed (incl. new regression test in `dockview-env-switch.test.ts`)

## Checklist

- [ ] Tests added/updated
- [ ] Docs updated if needed
- [ ] Self-reviewed the diff

### DIFF
--- a/apps/web/lib/state/dockview-env-switch.test.ts
+++ b/apps/web/lib/state/dockview-env-switch.test.ts
@@ -206,6 +206,52 @@ describe("performEnvSwitch", () => {
   });
 });
 
+describe("performEnvSwitch fast-path group survival", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("adds the incoming session panel before closing the outgoing chat (keeps its group alive)", () => {
+    // Regression: when the outgoing session chat is the only surviving panel in
+    // its group (e.g. the default layout's chat column), removing it FIRST
+    // empties — and so destroys — the group. The captured group id then no
+    // longer exists, and post-#1165 the old `referenceGroup: "sidebar"` fallback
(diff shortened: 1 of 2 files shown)


zeval:
Addressed the index-math note (claude suggestion) in the latest commit — documented that the raw `findIndex` ↔ survivor-index equivalence holds only while every panel `removeEphemeralPanels` strips ahead of the insert slot is ephemeral/stale, and flagged the off-by-one risk if `shouldRemoveDuringSwitch` ever retains a non-stale panel before the outgoing slot.

PR #1326, opened 2026-06-10
fix(orchestrator): apply max inflight cap to Sentry watcher tasks

nlenepveu:
## Summary

The per-watch **MaxInflightTasks** cap was never enforced for the **Sentry** watcher integration, even though it was configured, stored, and validated. Sentry watches could fan out unbounded tasks regardless of the limit. This PR fixes that and removes the design gap that caused it.

> [!IMPORTANT]
> **Behavior change: Sentry watcher caps are now enforced.** Previously a Sentry issue watch ignored its `MaxInflightTasks` limit and could create unbounded in-flight tasks. After this PR a Sentry watch is throttled to its configured cap (default 5) exactly like Linear and Jira — once the cap is reached, further detections are deferred until open tasks drain. Operators relying on the old (buggy) unbounded behavior will now see task creation throttled.
>
> No schema/migration change, no API change. Existing task rows count identically (the metadata keys are unchanged).

## Root cause

The orchestrator throttle gate (`acquireWatcherSlot`) counts open watcher-created tasks via `CountOpenWatcherCreatedTasks`, which resolved the per-integration task-metadata key through a hard-coded `watcherMetadataKeyByIntegration` switch in the SQLite repository (`internal/task/repository/sqlite/task.go`). That switch handled only `"linear"` and `"jira"`; `"sentry"` fell through to the empty default, which short-circuits the count to `(0, nil)`. So the gate always saw 0 open Sentry tasks and `count + pending >= cap` was effectively never true (only same-tick bursts were throttled).

Everything else in the Sentry path was already correct: the event carries `MaxInflightTasks`, the source exposes `WatchID`/`MaxInflightTasks`, and `BuildTaskRequest` writes `sentry_issue_watch_id` into task metadata. The only gap was the missing mapping entry — a generic persistence layer that had to know the integration catalogue.

## Fix + refactor

1. **Fix** — Sentry's cap is now enforced.
2. **Refactor** — ownership of the metadata key moves out of the SQLite repository onto each integration. Each `WatcherSource` exposes `WatchMetadataKey()` returning the same constant it writes in `BuildTaskRequest`; the throttle passes that key to `CountOpenWatcherCreatedTasks(metadataKey, watchID)`. The repository now keys purely on the metadata key (with a bare-identifier guard against SQL injection + a loud error on misuse) and no longer knows which integrations exist. `watcherMetadataKeyByIntegration` is deleted — a new capped watcher needs **no repository change**.
3. **Dedup** — `dispatchWatcherEvent` no longer takes a redundant integration literal; it reads `src.Name()`.

## Changes
- `internal/orchestrator/source_{sentry,linear,jira}.go` — per-source metadata-key constant + `WatchMetadataKey()`; the constant is the single source of truth shared with `BuildTaskRequest`.
- `internal/orchestrator/watcher_dispatch.go` — add `WatchMetadataKey()` to the `WatcherSource` interface.
- `internal/orchestrator/watcher_throttle.go` / `watcher_dispatch_wiring.go` — thread the metadata key through the gate; bypass on empty key; derive the integration label from `src.Name()`.
- `internal/task/repository/sqlite/task.go` — delete `watcherMetadataKeyByIntegration`; `CountOpenWatcherCreatedTasks` now takes `metadataKey` and validates it as a `[A-Za-z0-9_]` identifier.
- `internal/task/repository/interface.go` — interface signature/doc.
- Tests — throttle gate (incl. empty-key bypass + cross-integration isolation), repo regression (Sentry counting, decoupling proof for an unregistered future key, malformed-key/SQL-injection rejection), and source/counter fakes.
- `internal/integrations/AGENTS.md` — document that a capped watcher implements `WatchMetadataKey()` (no repository edit needed).

## Test plan
- [x] `make -C apps/backend lint` — 0 issues
- [x] `make -C apps/backend build` — OK
- [x] `make -C apps/backend test` — no failures
- [x] Regression test fails before the fix (`got 0`), passes after
- [x] Local code review — clean
- [x] CI green (backend + Windows + all E2E shards/containers), CodeRabbit + Cubic — no actionable findings


carlosflorencio:
FYI backend tests failing

PR #1379, opened 2026-06-15
Fix workspace deletion from settings

zajca:
## Summary

- Route Settings workspace deletion through the non-office workspace endpoint so it works when `features.office=false`.
- Cascade workspace deletion through workspace-owned tasks and workflows before deleting the workspace row.
- Add regression coverage for deleting a workspace with task/workflow data.

## Root Cause

The Settings delete action called `/api/v1/office/workspaces/:id`, but production disables Office routes. The legacy `/api/v1/workspaces/:id` path only deleted the workspace row and did not clean up workspace-owned task/workflow data first.

## Verification

- `go test ./internal/task/service -count=1`
- `pnpm --filter @kandev/web test -- app/actions/workspaces.test.ts`

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/1379?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/task/service/service_resources.go
+++ b/apps/backend/internal/task/service/service_resources.go
@@ -16,6 +16,7 @@ import (
 )
 
 const spritesTokenEnvKey = "SPRITES_API_TOKEN"
+const workspaceDeletePageSize = 500
 
 // Workspace operations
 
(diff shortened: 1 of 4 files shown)


carlosflorencio:
Thanks! approved the ci in both PRs to run

jcfs:
Thanks for this, and sorry it sat so long.

The backend half of this PR landed in #1547 (`657b92eb6`) — `listAllTasksForWorkspaceDelete`, `workspaceDeletePageSize`, and `TestService_DeleteWorkspaceDeletesWorkspaceOwnedTasksAndWorkflows` all came from here, and `DeleteWorkspace` has since grown a transactional cascade with secret and worktree cleanup on top. So that part is in and shipping.

The frontend half was deliberately left out of #1547 ("Settings deletion continues through the Office route"), and you were right that it's a real bug: the office router group is only mounted when `services.OfficeSvcs != nil` (`internal/backendapp/helpers.go`), which is nil when `features.office` is off — the prod default. The e2e coverage runs under the e2e profile where office is `true`, which is why CI never caught it.

The one thing the unconditional swap misses is that office tables have no FK cascade to `workspaces`; they're purged by a separate statement list in `internal/office/repository/sqlite/workspace_deletion.go`. Pointing every install at `/api/v1/workspaces/:id` would orphan that data wherever office *is* enabled. So the fix needs to pick the endpoint based on the flag.

Closing this as superseded, with a follow-up PR for the remaining frontend half — thanks for finding it.

PR #1399, opened 2026-06-16
fix(web): suppress kanban spinner for TODO tasks

jcfs:
## Summary
- Kanban cards in the `TODO` column no longer render the blue running spinner when a stale or transient `primarySessionState` (e.g. `RUNNING`, `STARTING`, `CREATED`) is still attached to the task.
- `TODO` is the canonical "queued, work not yet started" column; any active session activity reported there is either stale (task was moved back from `IN_PROGRESS` with a session still alive) or transient. The spinner was misleading in both cases.

## Implementation notes
Single-line guard at the top of `shouldShowTaskRunningSpinner` in `apps/web/lib/ui/state-icons.tsx`: when `taskState === "TODO"`, return `false` regardless of `primarySessionState`. The helper has exactly one consumer (`apps/web/components/kanban-card-content.tsx`), so the change is contained — `TopbarWorkingIndicator` uses live session lookup and is unaffected.

Rejected a backend fix (don't publish `primarySessionState` for TODO tasks) because the blast radius is wider (HTTP DTO + `task.updated` WS payload) and the frontend should still defend against stale state. Also rejected broadening `ACTIVE_SESSION_STATES` exclusions per task state — would couple session-state policy to workflow-column policy.

The clarification-request affordance (`shouldUseQuestionTaskIcon`) runs independently in `kanban-card-content.tsx` and still wins, so `TODO` + `WAITING_FOR_INPUT` + pending clarification keeps the question icon.

## Test plan
- [x] New `it("suppresses the spinner for TODO regardless of primary session state", ...)` block in `apps/web/lib/ui/state-icons.test.tsx` asserting `false` for `TODO` paired with `RUNNING`, `STARTING`, `CREATED`, `COMPLETED`, `WAITING_FOR_INPUT`, and `IDLE`.
- [x] Existing `shouldShowTaskRunningSpinner` cases (REVIEW/COMPLETED/FAILED/CANCELLED + RUNNING → true; IN_PROGRESS + non-active → false) remain green — TODO guard does not affect those branches.
- [x] `pnpm --filter @kandev/web test -- state-icons`, typecheck, lint.
- [ ] CI to re-run the full web test suite on push.

## Risks & rollback
Low risk. Single file behavior change; no DB, no API contract, no serialized state touched. If a custom workflow legitimately wants TODO tasks with active sessions to show the spinner, we can swap the rule to "TODO + session in CREATED → false" but keep `RUNNING` true. Rollback = revert the two-file commit.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/1399?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/web/lib/ui/state-icons.test.tsx
+++ b/apps/web/lib/ui/state-icons.test.tsx
@@ -71,4 +71,16 @@ describe("shouldShowTaskRunningSpinner", () => {
     expect(shouldShowTaskRunningSpinner("IN_PROGRESS", "WAITING_FOR_INPUT")).toBe(false);
     expect(shouldShowTaskRunningSpinner("IN_PROGRESS", "IDLE")).toBe(false);
   });
+
+  it("suppresses the spinner for TODO regardless of primary session state", () => {
+    // TODO is the queued/not-started column. A stale primary session state
+    // (e.g. task moved back from IN_PROGRESS with the session still alive)
+    // must not paint the running spinner on the kanban card.
+    expect(shouldShowTaskRunningSpinner("TODO", "RUNNING")).toBe(false);
+    expect(shouldShowTaskRunningSpinner("TODO", "STARTING")).toBe(false);
+    expect(shouldShowTaskRunningSpinner("TODO", "CREATED")).toBe(false);
+    expect(shouldShowTaskRunningSpinner("TODO", "COMPLETED")).toBe(false);
+    expect(shouldShowTaskRunningSpinner("TODO", "WAITING_FOR_INPUT")).toBe(false);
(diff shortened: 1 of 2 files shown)


jcfs:
Addressed Greptile nit in 0ab63a959: reworded the test description on line 34 from "any task state" to "non-TODO task states" since TODO now short-circuits the helper to `false`. Test body unchanged.

PR #1418, opened 2026-06-17
feat(github): publish GitHubPRMerged event on PR merge

StefanFVogel:
## Summary

Introduces a generic `GitHubPRMerged` bus event so workflow subscribers can fire on-merge actions without re-deriving the `MergedAt: nil -> time` transition from a full `TaskPR` diff. T07 of the Mantis-Lifecycle-Sync walking skeleton, but the event itself is integration-agnostic (no Mantis dependency).

- New constant `events.GitHubPRMerged = "github.pr_merged"` and payload struct `events.GitHubPRMergedEvent{TaskID, PRNumber, MergedAt}` in `apps/backend/internal/events/types.go`.
- `Service.SyncTaskPR` (in `apps/backend/internal/github/service_pr_watch.go`) now snapshots `MergedAt` before the field reassignments and a new helper `publishPRMergedIfTransitioned` emits the event after the DB write only when the field went from `nil` to a concrete time. Subsequent syncs with the same `MergedAt` are no-ops, so the event fires exactly once per merge.
- Tests in the new `apps/backend/internal/github/service_pr_watch_test.go` cover the three acceptance criteria: first sync (open, `merged_at=nil`) publishes nothing; second sync (`merged_at=<time>`) publishes exactly once; third identical sync publishes nothing (idempotency). A second test guards the already-merged-on-create case so the event does not double-fire when a task is imported with a PR that was merged upstream.

## Note on `GitHubPRStateChanged`

The existing `events.GitHubPRStateChanged` constant at `events/types.go:231` is **declared but unused** in the codebase. This PR intentionally leaves it as-is per the task scope, but if downstream work in T08 (workflow subscriber + `on_pr_merged` trigger) confirms that `GitHubPRMerged` fully replaces the role `GitHubPRStateChanged` was reserved for, that constant becomes safe to delete in a follow-up.

## Test plan

- [x] `make -C apps/backend fmt` — no changes needed
- [x] `make -C apps/backend vet` — clean
- [x] `make -C apps/backend test` — full backend suite passes, including the existing `service_pr_sync_test.go` path
- [x] `golangci-lint run ./... --new-from-rev=<merge-base>` — 0 issues, function-length / complexity limits respected via the small extracted helper
- [ ] Manual: confirm an actual GitHub PR merge fires the event end-to-end once the T08 subscriber lands

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/1418?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/events/types.go
+++ b/apps/backend/internal/events/types.go
@@ -1,6 +1,8 @@
 // Package events provides event types and utilities for the Kandev event system.
 package events
 
+import "time"
+
 // Event types for tasks
 const (
 	TaskCreated      = "task.created"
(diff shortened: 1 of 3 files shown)


carlosflorencio:
Nice, approved the CI.

Issue #1444, opened 2026-06-18
Workflow: auto-start steps with on_turn_complete auto-promote after the first agent turn (Linear-dispatched tasks land in Review)


nlenepveu:
## Summary

Tasks created by the Linear IssueWatch integration that are meant to land in **In Progress** auto-promote to **Review** after the agent's first turn. Originally looked like a one-off misconfigured workflow — but the same anti-pattern ships in **three built-in workflows**, so this is a shipped footgun rooted in a half-finished ADR-0015 migration. Reporting with a proposed direction and happy to send a PR once the maintainers confirm the shape.

## Root cause

A step that has **both** `auto_start_agent` (on_enter) **and** an `on_turn_complete: move_to_step` transition, with `auto_advance_requires_signal = 0`, auto-advances on the agent's first bare turn-end.

Flow (Linear-dispatched workflow, In Progress → Review):
1. `WatcherDispatchCoordinator.Dispatch` creates the task on In Progress and auto-starts the agent (`on_enter: auto_start_agent`).
2. The agent's **first turn ends** (~4 s — it has barely read the dispatch prompt).
3. `handleAgentReady → processOnTurnCompleteViaEngine` evaluates `on_turn_complete`. Because `auto_advance_requires_signal == 0`, the ADR-0015 signal gate is skipped and `move_to_step → Review` fires immediately.

There is **no first-turn / auto-start exemption** in the turn-complete path — every turn end is treated identically. `on_turn_complete: move` overloads one primitive for two incompatible meanings: *"the agent's turn ended"* (observable) vs. *"this step's goal is achieved"* (intended). This is exactly the "halts are overloaded" failure that [ADR 0015 — Explicit completion signal for auto-advance](docs/decisions/0015-explicit-completion-signal-for-auto-advance.md) was written to prevent.

**Why it still happens:** ADR 0015 shipped the disambiguator (`auto_advance_requires_signal` → advance only on `step_complete_kandev`) **opt-in, with the dangerous behavior as the silent default**, and the seed workflows were never migrated to opt in. The codebase is half-migrated.

## This is not a single misconfigured workflow

The same `auto_start_agent` + `on_turn_complete: move_to_step` combination ships in three built-ins, **none** of which set `auto_advance_requires_signal`:

| Built-in | Offending step | Transition |
|----------|----------------|-----------|
| `config/workflows/kanban.yml` | In Progress | → Review on turn-complete |
[...]

PR #1447, opened 2026-06-18
fix(workflow): gate default Kanban In Progress on completion signal

nlenepveu:
## What

Gate the default **Kanban** workflow's *In Progress* step on an explicit completion signal so it advances to *Review* when the agent signals done — not on the agent's first bare turn-end.

Fixes the premature-promotion bug from #1444 for the shipped default workflow.

## Why

The `simple` (Kanban) template — seeded as the default workflow for new workspaces — auto-starts an agent on *In Progress* and moves to *Review* on `on_turn_complete`. With `auto_advance_requires_signal` unset, that move fires on the **first** turn-end. For an auto-started agent (e.g. a Linear-dispatched task) the first turn can be ~4 s of just reading the prompt, so the task lands in *Review* before any real work.

The template's own description already promises the right behavior — *"the agent runs automatically, then move to Review **when done**"* — so this gates the transition on the `step_complete_kandev` signal (ADR-0015) to match that contract. The agent advances when it signals completion, or via the manual "Mark complete & advance" fallback.

## Changes

- `config/workflows/kanban.yml` — set `auto_advance_requires_signal: true` on *In Progress* (with an explanatory comment).
- `config/workflows/loader.go` — the strict YAML DTO (`stepDefYAML`) didn't carry the field, so the config couldn't even be parsed (`KnownFields(true)` rejects unknown keys). Added the field + mapping.
- `internal/workflow/repository/sqlite.go` — the hand-rolled default-seed `INSERT` omitted the `auto_advance_requires_signal` column, so the flag wouldn't persist for seeded default workflows. Added it. (The template-picker path already writes the column via `CreateStep`.)
- Tests: template-level (`kanban_signal_test.go`) and end-to-end seed (`seed_signal_test.go`).

## Scope

Backward compatible: default-workflow seeding only applies to workflows that have **no** steps, so existing workspaces are untouched. Autonomous `office-default`/`routine` workflows are intentionally left unchanged — the broader "should the unsafe default be inverted?" question is tracked in #1444.

## Test

```
go test ./config/workflows/... ./internal/workflow/... ./internal/orchestrator/...
```
All green; `go vet`, `gofmt`, and `golangci-lint` (changed-file, base `origin/main`) clean.

Refs #1444

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/1447?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/config/workflows/kanban.yml
+++ b/apps/backend/config/workflows/kanban.yml
@@ -30,6 +30,12 @@ steps:
     is_start_step: true
     show_in_command_panel: true
     allow_manual_move: true
+    # The agent auto-starts here and the description promises a move to Review
+    # "when done". Gate the on_turn_complete move on an explicit completion
+    # signal (step_complete_kandev) so the task is not promoted on the agent's
+    # first bare turn-end — e.g. an auto-started agent that has barely read its
+    # prompt — which would land work-in-progress tasks in Review prematurely.
+    auto_advance_requires_signal: true
     events:
       on_enter:
         - type: auto_start_agent
--- a/apps/backend/config/workflows/kanban_signal_test.go
+++ b/apps/backend/config/workflows/kanban_signal_test.go
@@ -1,0 +1,24 @@
+package workflows
+
+import "testing"
+
+// TestKanbanInProgress_GatesOnCompletionSignal pins that the default ("simple")
+// Kanban workflow advances In Progress -> Review only on an explicit completion
+// signal, not on the agent's first bare turn-end.
+//
+// The template's own description states the agent "runs automatically, then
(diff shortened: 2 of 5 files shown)


nlenepveu:
Closing per discussion in #1444 — default Kanban intentionally advances on turn-end; the right fix is the per-workflow completion-signal toggle, which we've adopted. Not changing the shipped default.

PR #1450, opened 2026-06-19
fix(orchestrator): inherit workflow default agent profile when auto-starting a profile-less session

nlenepveu:
## Problem

A task in a workflow step with the `auto_start_agent` on-enter action does not start its agent when the prepared session has no `agent_profile_id` — even though the task's workflow defines a default agent profile. The combined step prompt is recorded in the conversation, but no agent process ever runs.

This happens whenever the session reaching auto-start was prepared without a profile (for example, a task whose metadata `agent_profile_id` is empty). The only workaround is to move the task out of the step and back in, which re-triggers auto-start and works.

## Root cause

The on-enter auto-start path (`autoStartStepPrompt`) records the step prompt (`recordAutoStartMessage`) before launching, then calls `StartCreatedSession` with the session's stored profile, which is empty here.

In `StartCreatedSession` the resolution order was:

1. fall back to `session.AgentProfileID`
2. `if effectiveProfileID == "" { return "agent_profile_id is required" }` ← aborts here
3. `resolveEffectiveAgentProfile(...)` (workflow step override / workflow default) — never reached

The required-profile guard fired before the workflow-default fallback could resolve a profile. The prompt was already persisted, so the conversation showed the message with no running agent. Moving the task back into the step worked only because the session switch (`maybySwitchSessionForProfile`) then carried the resolved profile.

## Fix

Run `resolveEffectiveAgentProfile` before the required-profile guard. With an empty caller profile it resolves the step override then the workflow default; a non-empty caller is still overridden only when the step pins its own agent. The guard now fires only when no profile resolves at any level.

Strictly more permissive: the previously-resolvable case that used to error now launches; the genuinely-no-profile case still returns the same error.

## Test

`TestStartCreatedSession_EmptyProfileFallsBackToWorkflowDefault` seeds a CREATED session with an empty profile under a workflow that defines a default agent, calls `StartCreatedSession` with an empty profile, and asserts the launch resolves and persists the workflow default instead of returning `agent_profile_id is required`. Fails before the fix (with the exact error), passes after.

## Verification

- `gofmt` clean, `go build`, `go vet` clean
- `golangci-lint run ./internal/orchestrator/...` → 0 issues
- `go test -race ./internal/orchestrator/...` → pass

### DIFF
--- a/apps/backend/internal/orchestrator/task_operations.go
+++ b/apps/backend/internal/orchestrator/task_operations.go
@@ -266,17 +266,22 @@ func (s *Service) StartCreatedSession(ctx context.Context, taskID, sessionID, ag
 		return nil, fmt.Errorf("session is not in CREATED or WAITING_FOR_INPUT state (current: %s)", session.State)
 	}
 
-	// Use agent profile from request, fall back to session's stored value
+	// Use agent profile from request, fall back to session's stored value.
 	effectiveProfileID := agentProfileID
 	if effectiveProfileID == "" {
 		effectiveProfileID = session.AgentProfileID
 	}
+
+	// Resolve the workflow step override / workflow default before the
+	// required-profile guard, so a session without its own agent_profile_id
+	// inherits the workflow's default agent. resolveEffectiveAgentProfile keeps
+	// the caller profile only when neither a step override nor a workflow
+	// default applies; either of those overrides a non-empty caller.
+	effectiveProfileID = s.resolveEffectiveAgentProfile(ctx, taskID, "", effectiveProfileID)
+
 	if effectiveProfileID == "" {
 		return nil, fmt.Errorf("agent_profile_id is required")
 	}
-
-	// Override with workflow step's agent profile if one is configured.
-	effectiveProfileID = s.resolveEffectiveAgentProfile(ctx, taskID, "", effectiveProfileID)
 
 	// If the workflow step overrode the profile, update the session record in DB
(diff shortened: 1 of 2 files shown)


carlosflorencio (commented):
Looks good overall. The fix is the right ordering: resolve the workflow step/default profile before enforcing the required-profile guard, and the regression test covers the empty prepared-session case.

One small cleanup before merge: the new comment in `StartCreatedSession` says “a non-empty caller is overridden only when the step pins its own agent.” That is not quite accurate because `resolveStepAgentProfile` resolves `step override -> workflow default`, so a workflow default can also override a non-empty caller profile. The behavior looks pre-existing and intended, but please adjust the comment to say “step override or workflow default” so it matches the actual resolution order.

The failing `opencode-review-fork` check appears to be CI/review infra (`no merge base`), not a code issue.

nlenepveu:
Thanks @carlosflorencio. You're right — `resolveEffectiveAgentProfile` overrides a non-empty caller with whatever `resolveStepAgentProfile` returns, which is the step override *or* the workflow default, so the comment was inaccurate. Reworded it to:

> resolveEffectiveAgentProfile keeps the caller profile only when neither a step override nor a workflow default applies; either of those overrides a non-empty caller.

Pushed as a comment-only amend to the existing commit (17a91156). The `opencode-review-fork` failure is the fork-PR `no merge base` infra limitation, not a code issue, as you noted.

PR #1491, opened 2026-06-25
feat: bind repository to Linear/Jira/Sentry issue watchers

nlenepveu:
## Summary

Adds an optional **repository binding** (`repository_id` + `base_branch`) to the **Linear, Jira, and Sentry** issue-watchers, bringing them to parity with the GitHub watcher.

### Motivation

Today, Linear/Jira/Sentry watchers create tasks with no repository associated, so the agent launches into a blank scratch git repo (`initGitRepo` → empty `.gitkeep` + an "Initial commit") instead of the target codebase — i.e. "started with no repository checkout". The `workspace_path` override isn't a fit (it's only set by the UI starting-folder picker and runs the agent in-place with no per-task isolation, which doesn't suit a fan-out watcher).

With a repository bound, watcher-created tasks now carry `Repositories`, so they take the existing `len(req.Repositories) > 0` path and run against that codebase — exactly like the GitHub watcher.

## Backward compatibility (invariant)

An empty `repository_id` produces `Repositories == nil`, preserving today's exact repo-less behaviour. Only a non-empty `repository_id` changes anything. This is covered by per-integration tests (source builders assert `nil` when unbound and populated when bound; store tests assert unbound defaults; a migration test asserts existing rows backfill to unbound).

## Changes

**Backend (Linear / Jira / Sentry, mirroring each other):**
- Schema: expand-only `repository_id` / `base_branch` columns (`TEXT NOT NULL DEFAULT ''`, empty = unbound) added via idempotent `ADD COLUMN` migrations, following the existing `addIssueWatch*Columns` pattern. Fresh DBs get them inline in `CREATE TABLE`.
- Model + `Create`/`Update` DTOs + `applyIssueWatchPatch`; store INSERT/SELECT/UPDATE columns.
- `New{Linear,Jira,Sentry}IssueEvent` carry the binding, populated in the publishers.
- `BuildTaskRequest` in `source_{linear,jira,sentry}.go` populates `IssueTaskRequest.Repositories` only when bound (the existing `IssueTaskRequest`/`IssueTaskRepository` types are reused unchanged).
- Validation via a new `RepositoryLookup` dependency (wired post-construction like `watchreset.TaskDeleter`, through an adapter over the task service): checks the repo belongs to the watch's workspace and fills an empty base branch with the repo's default branch; rejects missing/cross-workspace/soft-deleted repos.

**Frontend (Vite/React SPA):**
- A shared `WatcherRepositoryFields` (repository + base-branch picker) added to all three watcher dialogs, reusing the existing `use-repository-branches` hook and following PR #1124's "(use step default)" sentinel pattern for the empty state.
- Types + form state + payload builders carry `repositoryId` / `baseBranch`; an empty selection maps back to `""` (unbound).

**Docs:** an implementation plan under `docs/plans/watcher-repository-binding/plan.md` (following the repo's `docs/plans` convention). Happy to drop it if you'd prefer a leaner PR.

## Test plan

- `make -C apps/backend fmt build` — clean
- `go test -race ./internal/{linear,jira,sentry,orchestrator,backendapp}/...` — pass (incl. store round-trip, migration, create/edit validation, and source-builder bound/unbound tests)
- `golangci-lint run --new-from-rev=<base>` — 0 issues on changed code
- `pnpm --filter @kandev/web typecheck && lint (--max-warnings 0)` — clean
- Affected web unit tests pass
- Manually verified end-to-end against a dev instance: binding a repo to a Linear watcher creates tasks in a worktree of that repo (not a scratch repo); editing the binding persists; unbinding restores repo-less behaviour.

## Notes for maintainers

- Deleted-repository dispatch self-heal (symmetric to the deleted-agent-profile preflight in #1094) is intentionally deferred — fail-open at dispatch, validate at config time. Documented in the plan's "Self-heal interplay" section.
- This is a fork PR, so the `*-same-repo` review/deploy workflows won't run here without the maintainer `safe-to-review` label.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/1491?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/backendapp/helpers.go
+++ b/apps/backend/internal/backendapp/helpers.go
@@ -516,14 +516,22 @@ func registerRoutes(p routeParams) {
 	if p.services.GitHub != nil {
 		p.services.GitHub.SetCascadeTaskDeleter(handoffSvc)
 	}
+	// repoLookup validates a watcher's optional repository binding (workspace
+	// ownership + default-branch fill) on create/update. Shared across the three
+	// repo-less watchers; one concrete adapter satisfies each package's
+	// structurally-identical RepositoryLookup interface.
+	repoLookup := &repositoryLookupAdapter{svc: p.taskSvc}
 	if p.services.Jira != nil {
 		p.services.Jira.SetTaskDeleter(handoffSvc)
+		p.services.Jira.SetRepositoryLookup(repoLookup)
 	}
 	if p.services.Linear != nil {
 		p.services.Linear.SetTaskDeleter(handoffSvc)
+		p.services.Linear.SetRepositoryLookup(repoLookup)
 	}
 	if p.services.Sentry != nil {
 		p.services.Sentry.SetTaskDeleter(handoffSvc)
+		p.services.Sentry.SetRepositoryLookup(repoLookup)
 	}
 	p.orchestratorSvc.SetWorkspaceMaterializer(handoffSvc)
 	// Phase 8 prompt enrichment — wire the office scheduler's
--- a/apps/backend/internal/backendapp/main.go
+++ b/apps/backend/internal/backendapp/main.go
@@ -476,6 +476,9 @@ func startAgentInfrastructure(
 	// reconciler when its agent type left the registry) self-heal on the
 	// next poll instead of looping on "profile not found" forever.
 	orchestratorSvc.SetProfileLookup(&profileLookupAdapter{store: repos.AgentSettings})
+	// Watcher dispatch self-heals a binding whose repository was soft-deleted
+	// after the watch was configured, instead of creating an orphan task row.
+	orchestratorSvc.SetRepositoryChecker(&repositoryLookupAdapter{svc: services.Task})
 
 	// Wire the watcher-dependency enumerator into the agent settings
 	// controller so the profile-delete UI can surface "this will also
--- a/apps/backend/internal/backendapp/turn_adapters.go
+++ b/apps/backend/internal/backendapp/turn_adapters.go
@@ -107,3 +107,20 @@ func (a *taskDeleterAdapter) DeleteTask(ctx context.Context, taskID string) erro
 	}
 	return err
 }
+
+// repositoryLookupAdapter satisfies the linear/jira/sentry RepositoryLookup
+// interface over the task service. It is the validation seam for a watcher's
+// optional repository binding. The task service's GetRepository filters
+// soft-deleted rows and errors on a miss, so a missing or deleted repository
+// maps to ok=false and watcher create/update rejects the binding.
+type repositoryLookupAdapter struct {
+	svc *taskservice.Service
+}
+
+func (a *repositoryLookupAdapter) GetRepository(ctx context.Context, id string) (string, string, bool) {
+	repo, err := a.svc.GetRepository(ctx, id)
+	if err != nil || repo == nil {
+		return "", "", false
+	}
+	return repo.WorkspaceID, repo.DefaultBranch, true
+}
--- a/apps/backend/internal/backendapp/turn_adapters.go
+++ b/apps/backend/internal/backendapp/turn_adapters.go
@@ -124,3 +124,21 @@ func (a *repositoryLookupAdapter) GetRepository(ctx context.Context, id string) 
 	}
 	return repo.WorkspaceID, repo.DefaultBranch, true
 }
+
+// RepositoryExists satisfies orchestrator.RepositoryChecker. It uses the
+// workspace listing (which excludes soft-deleted repos) so a definitive
+// "absent" is distinguishable from a transient error: a non-nil err lets the
+// dispatch pre-flight fail open, while (false, nil) means the bound repository
+// was removed and the watcher should self-heal.
+func (a *repositoryLookupAdapter) RepositoryExists(ctx context.Context, workspaceID, repositoryID string) (bool, error) {
+	repos, err := a.svc.ListRepositories(ctx, workspaceID)
+	if err != nil {
+		return false, err
+	}
+	for _, repo := range repos {
+		if repo.ID == repositoryID {
+			return true, nil
+		}
+	}
+	return false, nil
+}
--- a/apps/backend/internal/common/securityutil/git.go
+++ b/apps/backend/internal/common/securityutil/git.go
@@ -32,6 +32,18 @@ func IsValidBranchName(branch string) bool {
 	return validBranchNameRegex.MatchString(branch)
 }
 
+// IsValidBaseBranchRef validates a base-branch ref using the IsValidBranchName
+// allowlist, transparently stripping an "origin/" prefix (the regex disallows
+// "/" as the first character). Shared so service-tier base-branch rejection
+// uses the same allowlist as task base-branch overrides and the agentctl-side
+// sanitiser.
+func IsValidBaseBranchRef(ref string) bool {
+	if rest, ok := strings.CutPrefix(ref, "origin/"); ok {
+		return IsValidBranchName(rest)
+	}
+	return IsValidBranchName(ref)
+}
+
 // IsKnownSafeGitFlag returns true if the argument is a known safe git flag used by this codebase.
 // This prevents argument injection where user input could introduce malicious flags.
 // Only flags actually used by the Kandev codebase are whitelisted.
--- a/apps/backend/internal/common/securityutil/git.go
+++ b/apps/backend/internal/common/securityutil/git.go
@@ -38,10 +38,17 @@ func IsValidBranchName(branch string) bool {
 // uses the same allowlist as task base-branch overrides and the agentctl-side
 // sanitiser.
 func IsValidBaseBranchRef(ref string) bool {
-	if rest, ok := strings.CutPrefix(ref, "origin/"); ok {
-		return IsValidBranchName(rest)
+	rest := ref
+	if stripped, ok := strings.CutPrefix(ref, "origin/"); ok {
+		rest = stripped
 	}
-	return IsValidBranchName(ref)
+	// git check-ref-format also rejects a trailing "/" and consecutive "//",
+	// which the IsValidBranchName regex (its char class includes "/") permits.
+	// Reject them here so "main/" / "origin/main/" / "a//b" don't pass.
+	if strings.HasSuffix(rest, "/") || strings.Contains(rest, "//") {
+		return false
+	}
+	return IsValidBranchName(rest)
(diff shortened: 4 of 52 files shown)


nlenepveu:
Thanks for the thorough reviews (CodeRabbit, cubic, Greptile). Addressed in `f60b1237`:

**Fixed**
- **Unrelated edits blocked by a deleted bound repo** (Greptile P1): `UpdateIssueWatch` now only re-validates the binding when the PATCH actually touches `repository_id`/`base_branch`, so editing prompt/filter/JQL on a watch whose repo was soft-deleted no longer fails. (linear/jira/sentry + regression test)
- **Stale base branch on repo rebind** (CodeRabbit, cubic): switching repository without an explicit branch now resets the branch so the new repo's default is used. (all three services + test)
- **Workspace switch kept stale repo/branch** (CodeRabbit, cubic): the dialogs now clear `repositoryId`/`baseBranch` on workspace change (extracted to a `formWithWorkspace` helper).
- **Sentinel collision** (CodeRabbit, cubic): `__no_repository__`/`__default_branch__` → `kandev:no-repository`/`kandev:default-branch` (`:` is illegal in git refs and never in a repo UUID).
- **Data fetch in component** (CodeRabbit) + **refresh guard set before resolve** (cubic): the picker's force-refresh moved into `useRepositories` (opt-in `forceRefresh`); the per-workspace guard is set only on success so a failed fetch retries.
- **`mainmain` duplicate branch render** (from manual testing): base-branch options deduped by name (local+remote `main` collapsed).
- **Migration guard consistency** (Greptile P2): added the `len(cols)==0` guard to the linear/jira repo-column migrations to match sentry.
- **Plan doc wording** (CodeRabbit): corrected the "nullable" / "use step default" descriptions.

**Deferred (with rationale)**
- **Lookup error collapsed to `ErrInvalidConfig`** (CodeRabbit): cleanly distinguishing a transient store error from not-found needs a typed not-found sentinel in the task repository layer; out of scope for this feature. Current behaviour fails safe — it refuses to bind to a repository it can't verify. Happy to follow up if you'd like the task-layer sentinel.
- **`pollInterval` `Number()` coercion per keystroke** (cubic): pre-existing behaviour (this PR only relocated the field) and already guarded by submit-time validation (60–3600). Left as-is to keep the diff focused.
- **Base-branch ref validation** (cubic): the value comes from the repo's real branch list or its default branch, not free-form input; a malformed ref would fail worktree creation cleanly. Can add explicit ref validation if preferred.

nlenepveu:
Follow-up rounds (`f60b1237`, `2fe55050`, `0032620f`) addressed the actionable review items:

- Binding re-validation now gates on the binding **value changing** (not field presence), so the dialog re-sending an unchanged binding — or editing unrelated fields on a watch whose repo was soft-deleted — no longer re-validates.
- Repo rebind resets the base branch to the new repo's default.
- Workspace switch clears the repo/branch (deduped into a shared `clearWorkspaceScopedForm`, which now also no-ops when the workspace is unchanged).
- `useRepositories` no longer uses a shared in-flight marker (stable effect deps + the `cancelled` flag make it unnecessary), removing the cross-workspace blocking / stale-clear concerns.
- Sentinels made collision-proof; base-branch options deduped; `BaseBranch` preservation now asserted in the rebind tests.

**Still deferred (rationale):**
- **Dispatch-time revalidation of a soft-deleted bound repo** — this is the deleted-repository self-heal the spec documents as a follow-up (symmetric to the deleted-agent-profile preflight in #1094); it fails open today and is intentionally out of this PR's scope.
- **Lookup error → `ErrInvalidConfig`** — distinguishing a transient store error from not-found needs a typed not-found sentinel in the task repository layer; current behaviour fails safe (refuses to bind to an unverifiable repo). Happy to do the task-layer follow-up if wanted.
- **Base-branch ref validation** — the value comes from the repo's real branch list or default branch, not free-form input, and a malformed direct-API value fails worktree creation cleanly; there's no existing git-ref validator to reuse, so I left this out to keep the PR scoped.

Happy to pick up any of the deferred items if you'd prefer them in this PR.

carlosflorencio:
Thanks for the work here. Two things I think we should tighten before merging:

1. The new watcher `baseBranch` value is accepted from the API and persisted after trimming, but I do not see validation that it is a safe git ref. Since this can come from non-UI callers and later flows into task/worktree launch, can we reject invalid non-empty branches at create/update time using the same allowlist pattern we use elsewhere for base branches? A small regression test for Linear/Jira/Sentry create and update would cover it.

2. Could we add focused coverage for the new frontend repository-binding helpers and picker behavior? The sentinel mapping, workspace-scoped clearing, and `useRepositories(..., forceRefresh)` behavior are important for avoiding stale cross-workspace repo IDs and making sure the backend receives either `""` or a real repository ID. A small unit test for `watcher-repository-default.ts`, plus a focused hook test or mobile watcher E2E around selecting a repository/base branch, would make this easier to trust.

nlenepveu:
Latest review wave (Claude CI / OpenCode):

- **Blocker — missing unit tests for `watcher-repository-default.ts`**: fixed in `9732268d` — added `watcher-repository-default.test.ts` covering `resolveRepositoryId`, `resolveBaseBranch`, `clearWorkspaceScopedForm` (including the same-workspace no-op), and `branchPlaceholder` (9 cases).

Two suggestions assessed and intentionally left as-is:

- **`useRepositories` may double-fetch without an in-flight ref (strict mode/HMR)**: the in-flight ref was removed in the previous round specifically because, keyed by workspace, it could be mis-cleared by a stale `finally` and could block a different workspace's fetch (flagged P2 by cubic). Production React doesn't double-invoke effects; the dev-only duplicate is a `no-store` GET whose result is idempotent (`setRepositories` overwrites with identical data) and `forcedRef`/`isLoaded` prevent post-success refetches. Re-adding a shared guard reintroduces the cross-workspace bug, so this is a deliberate trade-off. Open to a per-request-token guard if you'd prefer belt-and-suspenders.
- **Picker shows "(no repository)" when the bound repo was soft-deleted**: edge case in the deleted-repository path that the spec already lists as a follow-up (dispatch-time self-heal). The binding itself is preserved on save (unchanged bindings skip re-validation), so there's no data loss — only a display gap. Will fold into the deleted-repo follow-up.

nlenepveu:
Thanks @carlosflorencio — both addressed in `77031603`:

**1. Base-branch ref validation.** `resolveRepositoryBinding` now rejects a non-empty `baseBranch` that isn't a safe git ref, at both create and update, before it's persisted or copied into watcher tasks. It reuses the existing allowlist via a new `securityutil.IsValidBaseBranchRef` (wraps `IsValidBranchName`, stripping an `origin/` prefix — the same pattern `service_branch_update.go`'s `isSafeBaseBranchRef` uses for task base branches). Regression tests added for Linear/Jira/Sentry covering both create and update with a malformed ref (`bad..ref` → `ErrInvalidConfig`). Empty `baseBranch` still defers to the repo's default branch.

**2. Frontend coverage.**
- `apps/web/lib/watcher-repository-default.test.ts` — sentinel round-trips (`resolveRepositoryId`/`resolveBaseBranch`), the `clearWorkspaceScopedForm` no-op-when-unchanged + clear-on-switch cases, and the three `branchPlaceholder` states.
- `apps/web/hooks/domains/workspace/use-repositories.test.ts` — `forceRefresh` fetches a fresh list even when the workspace is already loaded and writes it to the store; the lazy path doesn't fetch when loaded and does when not; no-ops when disabled / null workspace.

This pins the "backend receives either `\"\"` or a real repository ID" contract (sentinel mapping) and the cross-workspace clearing. I can add a mobile watcher E2E around the repo/branch picker on top if you'd prefer end-to-end coverage too — say the word.

carlosflorencio:
Thanks for the follow-up fixes. I think there are still two things worth addressing before merge:

1. The stale/deleted repository binding path still looks risky. `CreateTask` persists the task row before `createTaskRepositories` runs, so if a watcher is bound to a repo that was later soft-deleted, repository association can fail after the task row already exists. `watcher_dispatch` then releases the reservation, so a later poll can repeat the same hidden task creation. Can we either revalidate/self-heal the watcher binding before creating the task, or make task + repository association atomic so this cannot leave orphan rows?

2. The new base-branch validation is a good improvement, but it still appears to allow trailing-slash refs such as `main/` or `origin/main/` because `IsValidBranchName` permits them and `IsValidBaseBranchRef` delegates to it after stripping `origin/`. Git ref-format rejects those, so we should reject them before persisting the watcher config rather than failing later during worktree launch.

The previously reported missing frontend utility tests look addressed now.

nlenepveu:
Both addressed in `c808d5fb`:

**1. Stale/deleted bound repository.** `CreateTask` isn't atomic (task row, then `createTaskRepositories`), and wrapping that central method felt too broad for this PR — so I went the self-heal route you offered. `WatcherDispatchCoordinator` now pre-flights the bound repository after `BuildTaskRequest`, before `CreateIssueTask`: if it's gone, it self-heals the watch (`SelfHeal` → disable + `last_error`, same as the deleted-profile preflight) and releases the reservation, so no orphan task row is created and a later poll doesn't repeat. Existence is checked via workspace `ListRepositories` membership rather than `GetRepository`, so a definitive "absent" (→ self-heal) is distinguishable from a transient error (→ fail open, pipeline runs). Wired through `Service.SetRepositoryChecker` → `repositoryLookupAdapter` over the task service. Tests cover deleted-repo self-heal, fail-open on error, and live-repo passthrough.

**2. Trailing-slash refs.** `IsValidBaseBranchRef` now rejects a trailing `/` and consecutive `//` (which the `IsValidBranchName` regex permits) before validating, so `main/` and `origin/main/` are rejected at create/update. Added `securityutil/git_test.go` covering these plus the existing cases. (Left `IsValidBranchName` itself unchanged to avoid changing behavior for its other callers; the watcher path now goes through the stricter `IsValidBaseBranchRef`.)

The orphan-row concern is fully closed for the Linear/Jira/Sentry watchers via this self-heal. If you'd prefer the stronger guarantee of atomic task+repository creation in `task/service.CreateTask` (which would also cover the GitHub watcher and any future caller), I'm happy to do that as a focused follow-up — it's just a wider change than this PR's scope.

carlosflorencio:
Nice, merging 💪

PR #1497, opened 2026-06-25
feat: add Devin ACP agent integration

jijiamoer:
Devin ACP now works end to end in Kandev because the host utility strips `ACP_BACKEND` before spawning Devin, so the agent keeps its fallback credential path even when the host inherits ACP settings. I verified the path against a live backend, a direct `acpdbg` prompt round trip, and the backend API.

## Important Changes
- Added the `devin-acp` agent runtime and bootstrap path so Kandev can probe and serve Devin natively.
- Made `RuntimeConfig.StripEnv` the shared source for persistent, probe, and inference subprocess env stripping while preserving one-shot temp env handling.
- Stripped `ACP_BACKEND` from Devin child processes so the agent keeps its fallback credential path under inherited ACP host settings.

## Validation
- `env -u GOROOT TMPDIR=/private/tmp go test -tags fts5 ./internal/agent/agents ./internal/agent/hostutility ./internal/agent/runtime/lifecycle ./internal/agentctl/server/process`
- `env -u GOROOT TMPDIR=/private/tmp make lint`
- `env -u GOROOT TMPDIR=/private/tmp make typecheck`
- `env -u GOROOT TMPDIR=/private/tmp make typecheck test lint` (failed in `apps/cli/src/ports.test.ts` on the unrelated black-holed-host timeout case)
- `env -u GOROOT -u ACP_BACKEND TMPDIR=/private/tmp make acpdbg ARGS='probe devin-acp --timeout 90s --stderr --out work/devin-acpdbg'`
- `env -u GOROOT -u ACP_BACKEND TMPDIR=/private/tmp apps/backend/bin/acpdbg prompt --timeout 180s --stderr --prompt 'Reply with exactly kandev-devin-ok and nothing else.' devin-acp`
- `env -u GOROOT TMPDIR=/private/tmp KANDEV_HOME_DIR=/private/tmp/kandev-devin-home ACP_BACKEND=windsurf KANDEV_LOG_LEVEL=info apps/backend/bin/kandev __backend --port 38430`
- `curl -fsS http://127.0.0.1:38430/health`
- `curl -fsS http://127.0.0.1:38430/api/v1/agent-capabilities/devin-acp`
- `curl -fsS -X POST http://127.0.0.1:38430/api/v1/utility/inference-agents/devin-acp/refresh`

## Checklist


apps/backend/internal/agent/hostutility/stripenv.go
```diff
@@ -0,0 +1,17 @@
+package hostutility
+
+import "github.com/kandev/kandev/internal/agent/agents"
+
+// stripEnvFor 返回 agent 在持久会话路径声明的 StripEnv，作为一次性
+// probe/inference 子进程的唯一真相源（RuntimeConfig.StripEnv）。
```
carlosflorencio: Can you keep the comments in english only?

carlosflorencio (commented):
Thanks for the contribution, can you also update the readme to mention support for devin?

jijiamoer:
Updated README to mention Devin support in 17b4eca0. I also pushed 687dfd3c to address the automated StripEnv follow-ups and added the targeted validation to the PR description.

apps/backend/internal/agent/hostutility/stripenv.go
jijiamoer: Done in bff87dea. I translated the remaining StripEnv-related comments to English and verified there are no Han characters left in the PR-related Go/Markdown files.

carlosflorencio:
The Devin `InstallScript()` currently returns prose instead of executable shell:

```go
return "Install Devin CLI from https://devin.ai"
```

That value is inserted into the Sprites prepare script via `{{kandev.agents.install}}` and executed under `set -euo pipefail`. Selecting Devin for a Sprites-backed run will try to execute `Install Devin CLI...` as a command and fail before agentctl starts.

Please either return a real shell install script for Devin, or return an empty string and explicitly treat remote install as unsupported. Devin's CLI docs list `curl -fsSL https://cli.devin.ai/install.sh | bash` and document `devin acp`; if we use that path, we should also ensure `$HOME/.local/bin` is on `PATH` for the same prepare/run path.

carlosflorencio:
One more architecture-level concern: `DevinACP.RemoteAuth()` currently returns `nil`, while the agent docs/comments say `devin acp` falls back to stored local credentials when `ACP_BACKEND` is absent.

That means the existing remote credential pipeline has no Devin method to show/select, and Sprites/SSH cannot copy or bootstrap Devin auth for an ephemeral or remote environment. The install-script fix would get the binary onto the machine, but the agent would still generally start unauthenticated unless that remote machine had already been logged in manually.

Please either add a `RemoteAuth` method for Devin or explicitly scope this integration to already-authenticated local hosts. Based on Devin's CLI docs, the likely options are:

- an env method for `WINDSURF_API_KEY`, since `devin acp` reads it when set; and/or
- a files method for the stored Devin credentials path used by `devin auth login` if that path is stable enough to support.

This keeps Devin aligned with the existing Kandev architecture where remote-capable agents declare their auth contract via `RemoteAuth()` instead of relying on per-machine manual setup.

jijiamoer:
@carlosflorencio thanks for the careful review — both architectural concerns are now addressed and pushed in `0193deaf` (Devin auth/install) and `93fa9cd7` (credential perms hardening).

### 1. `InstallScript()` returning prose

It now returns executable shell built around the official installer you pointed to:

```sh
curl -fsSL https://cli.devin.ai/install.sh -o "$tmp"
if ! bash "$tmp" && [ ! -x "$HOME/.local/bin/devin" ]; then
  exit 1
fi
```

It also persists `$HOME/.local/bin` on `PATH` across the common profile files so the same prepare/run path can find `devin`, and it treats the binary's presence as the success signal rather than the installer's exit code — the installer's trailing `devin setup` returns 1 on "Login canceled" in headless/non-interactive envs even when the binary installed fine.

### 2. `RemoteAuth()` returning `nil`

`RemoteAuth()` now declares both options you suggested, so Devin flows through the existing remote-credential pipeline like the other remote-capable agents instead of relying on a pre-logged-in host:

- **files** — copies `~/.local/share/devin/credentials.toml` to the remote
- **env** — `WINDSURF_API_KEY`, with a setup script that writes the same `credentials.toml` on the remote (mirrors how `devin acp` reads the key)

While wiring this up I also tightened the shared credential uploader to write files `0600` instead of `0644` (`93fa9cd7`), since these are secrets.

This keeps Devin aligned with the existing `RemoteAuth()` contract rather than per-machine manual setup. Added unit tests for both paths (`devin_acp_test.go`, `credential_uploader_test.go`).

carlosflorencio:
Approved the CI to run 💪

carlosflorencio:
I do not think this is ready to merge yet. The latest commits fixed the earlier install-script and RemoteAuth concerns, and CI is green, but I still see three merge-blocking issues:

1. **Docker credentials are still not mounted.** `DevinACP.Runtime()` sets `SessionDirTemplate` but not `SessionDirTarget`. Docker only mounts the seeded session directory when both fields are set, so the copied `~/.local/share/devin/credentials.toml` still will not be visible inside the container. Please add `SessionDirTarget: "/root/.local/share/devin"` and cover it with a Docker/session-dir test.

2. **`credentials.toml` is created before private permissions apply.** The env auth `SetupScript` writes with `cat > ~/.local/share/devin/credentials.toml` and only runs `chmod 600` afterwards. That can briefly create the API key file with the remote shell's default umask. Please set `umask 077` before the redirection, or write through a securely-created temp file and rename it into place.

3. **Devin passthrough still inherits `ACP_BACKEND`.** The normal ACP subprocess path strips `ACP_BACKEND`, but CLI passthrough launches go through `InteractiveStartRequest.Env` and `mergeEnv`, which starts from `os.Environ()`. If Kandev itself inherited `ACP_BACKEND`, passthrough `devin` still sees it and skips the local credential fallback. Please plumb `StripEnv` through the passthrough/interactive runner env path, or disable Devin passthrough until that path can strip it.

Once those are addressed, I think the shape of the integration is otherwise sound.

carlosflorencio:
I can also address the mentioned points on my comment, let me know 💪

PR #1541, opened 2026-06-29
fix: skip cleanup block when stop fails for terminal sessions

dominikkucharski:
Archiving a task with a previously-active agent left the worktree and cleanup script behind. `ArchiveTaskTree` calls `cancelActiveRuns` first (sessions become CANCELLED in the DB), then `CleanupTaskResources` tries to stop them — but `StopSession` returns `ErrExecutionNotFound` because the agent is already gone. The stop failure was recorded in `failedStops`, which caused `cleanupDestructiveTaskResources` to skip environment teardown entirely.

## Important Changes

- `taskStopTarget` gains a `terminal bool` field — set when a session is already in a state where the agent has no live execution (CANCELLED, COMPLETED, FAILED, IDLE).
- New `isCleanableSessionState` helper (distinct from `orchestrator.isTerminalSessionState`, which excludes IDLE) identifies sessions whose stop failures are expected and must not block cleanup.
- `buildStopTargets` marks executor-running rows whose session is already cleanable as `terminal`; sessions without an executor-running row that are already cleanable are skipped entirely.
- `stopTaskRuntimeTargets` logs and skips — rather than recording in `failedStops` — stop errors for terminal targets, so cleanup always proceeds when the agent is known to be gone.

## Validation

Reproduced with two stuck tasks: worktrees remained after archive, `executors_running` rows persisted, DDEV environments were not destroyed. Traced root cause through `cancelActiveRuns` → `StopSession` → `ErrExecutionNotFound` → `failedStops` → skipped teardown.

Manual cleanup of the two stuck tasks confirmed the fix addresses the right code path. Unit tests cover `isCleanableSessionState`, `buildStopTargets` (terminal and non-terminal rows, and terminal sessions with no executor row), and `stopTaskRuntimeTargets` (terminal failure skipped vs. non-terminal failure added to `failedStops`).

## Diagram

```mermaid
sequenceDiagram
    participant A as ArchiveTaskTree
    participant C as cancelActiveRuns
    participant B as buildStopTargets
    participant S as stopTaskRuntimeTargets
    participant P as performTaskCleanup

    A->>C: mark sessions CANCELLED
    A->>B: buildStopTargets(activeSessions)
    note over B: executor_running rows for CANCELLED sessions → terminal=true
    B-->>A: []taskStopTarget
    A->>S: stopTaskRuntimeTargets(targets)
    note over S: stop fails (ErrExecutionNotFound) + terminal=true → Debug log, skip failedStops
    S-->>A: failedStops = empty
    A->>P: performTaskCleanup(failedStops=empty)
    note over P: cleanupDestructiveTaskResources runs → worktree + cleanup script removed
```

## Possible Improvements

Low risk. The only scenario where this loosens the guard is a genuine live session that somehow ends up in a terminal DB state while its agent is still running — that would require a separate bug upstream. IDLE is included intentionally (an idle session has released its execution slot and returns `ErrExecutionNotFound` on stop); the comment in `isCleanableSessionState` explains the divergence from the orchestrator helper of the same pattern.

## Checklist

### DIFF
--- a/apps/backend/internal/task/service/service_tasks.go
+++ b/apps/backend/internal/task/service/service_tasks.go
@@ -1020,10 +1020,13 @@ func (s *Service) runAsyncTaskCleanup(
 	}()
 }
 
-// isTerminalSessionState reports whether a session is already in a terminal state.
-// Terminal sessions have no running agent process; stop failures are expected and
-// must not block cleanup.
-func isTerminalSessionState(state models.TaskSessionState) bool {
+// isCleanableSessionState reports whether a session has no running agent process
+// and stop failures are therefore expected. Unlike the orchestrator's
+// isTerminalSessionState (which excludes IDLE), this helper is used only during
+// task cleanup to decide whether a stop failure should block environment teardown.
+// IDLE is included because an idle session has already released its execution slot
+// and will return ErrExecutionNotFound just like CANCELLED/COMPLETED/FAILED.
+func isCleanableSessionState(state models.TaskSessionState) bool {
 	switch state {
 	case models.TaskSessionStateCancelled,
 		models.TaskSessionStateCompleted,
@@ -1056,7 +1059,7 @@ func (s *Service) buildStopTargets(ctx context.Context, taskID string, activeSes
 			target := taskStopTarget{
 				sessionID:   running.SessionID,
 				executionID: strings.TrimSpace(running.AgentExecutionID),
-				terminal:    isTerminalSessionState(sessionStates[running.SessionID]),
+				terminal:    isCleanableSessionState(sessionStates[running.SessionID]),
 			}
 			targets = append(targets, target)
 			seen[target.sessionID] = struct{}{}
@@ -1071,7 +1074,7 @@ func (s *Service) buildStopTargets(ctx context.Context, taskID string, activeSes
 		}
 		// Sessions without an executor_running row that are already in a terminal
 		// state have no running process; skip creating a stop target.
-		if isTerminalSessionState(sess.State) {
+		if isCleanableSessionState(sess.State) {
 			continue
 		}
 		target := taskStopTarget{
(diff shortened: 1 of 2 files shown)


carlosflorencio:
I have approved CI to run on this PR. Thanks for the contribution.

One request before merge: please add a focused regression test for the cascade cleanup case this fixes. The test should seed a terminal session (for example CANCELLED or COMPLETED) that still has an executors_running row, make StopExecution return an error/not-found, call CleanupTaskResources, wait for async cleanup, and assert the executor row and cleanup-owned resources are removed. A companion case for a non-terminal RUNNING session where StopExecution fails should continue to assert cleanup remains blocked.

dominikkucharski:
Quick status update on the three review findings:

**Greptile P2 / CodeRabbit (regression tests)** — both are stale; they reviewed commit `f1b1897`. The rename to `isCleanableSessionState` (with an explicit comment explaining the IDLE divergence from the orchestrator helper) landed in `17819fc`, and the `CleanupTaskResources` end-to-end regression tests landed in `bcc5a92`.

**Codex P1 (preserve row unless stopper distinguishes already-missing)** — investigated, no code change needed. The suggestion is to check `errors.Is(err, executor.ErrExecutionNotFound)` specifically rather than skipping all stop errors for terminal sessions. Looking at the actual implementation:

- `executor.StopExecution` (executor_interaction.go:71–86): calls `agentManager.StopAgentWithReason`; on *any* failure — including transient errors — it returns `ErrExecutionNotFound`. There is no other error type.
- `executor.StopSession` → `stopWithSession`: returns either `nil` (execution found, CANCELLED marked, goroutine launched) or `ErrExecutionNotFound` (execution not in lifecycle manager).

Both stop methods only ever return `nil` or `ErrExecutionNotFound`. Checking `errors.Is(err, ErrExecutionNotFound)` would produce identical behaviour to the current fix — every stop failure is already typed as not-found. There is no signal to distinguish "transient failure on a live agent" from "agent already gone".

carlosflorencio:
Approved the ci again 💪

carlosflorencio:
Thanks, I will create a new release soon to include this.

dominikkucharski:
Thanks @carlosflorencio  !

PR #1562, opened 2026-07-01
fix: keep user-selected terminal session active instead of auto-handoff

jingrandev:
Clicking a terminal (COMPLETED / FAILED / CANCELLED) session tab snaps the UI back to another running session for the task. The user explicitly opened the terminal session to review it, so the auto-handoff fights their selection.

`maybeAdoptSessionOnTransition` hands the active session off to a replacement when it reaches a terminal state — intended for workflow-driven runs. But it didn't distinguish a session the user pinned (via `setActiveSession`) from one the system auto-selected (via `setActiveSessionAuto`). When the backend re-emitted the terminal `session.state_changed` for a session the user had just clicked open, the handler overrode their choice and flipped `activeSessionId` back to a running session, and `useAutoSessionTab` followed the store back to the primary tab.

## Important Changes
- `maybeAdoptSessionOnTransition`: skip the terminal-replacement handoff when the terminal session is the one the user pinned (`pinnedSessionId === sessionId`). Workflow-driven handoffs go through `setActiveSessionAuto` (which does not set the pin) and keep the existing replacement behavior unchanged.

## Validation
- `pnpm --filter @kandev/web test src/lib/ws/handlers/agent-session.test.ts`
- Manual: in a multi-session task, click the tab of a COMPLETED/FAILED session and confirm the view stays on it. Previously it bounced back to the primary session within ~1s once the backend re-emitted the terminal `state_changed`.

## Possible Improvements
Low risk. The pin flag is the existing signal that separates user-initiated (`setActiveSession`) from system-initiated (`setActiveSessionAuto`) selection, so the guard reuses an established semantic rather than introducing a new one.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/1562?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/web/lib/ws/handlers/agent-session.test.ts
+++ b/apps/web/lib/ws/handlers/agent-session.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines -- comprehensive session-state handler tests */
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import { registerTaskSessionHandlers } from "./agent-session";
 import type { StoreApi } from "zustand";
@@ -458,11 +459,10 @@ describe("session.state_changed → respects user-pinned session", () => {
     vi.clearAllMocks();
   });
 
-  it("does not hand off when the user pinned the active session even as it reaches a terminal state", () => {
-    // Regression for the "click a non-primary tab, it snaps back to primary"
-    // bug: setActiveSession pins the session, and that pin must be honored
-    // when the backend re-emits the terminal state_changed — the user opened
-    // this terminal session on purpose to review it.
+  it("hands off when the pinned active session reaches a terminal state", () => {
+    // Genuine RUNNING→COMPLETED transition: previousState is non-terminal,
+    // so the workflow handoff should still fire even though the session
+    // is pinned.
     const store = makeStore({
       tasks: {
         activeTaskId: "t-1",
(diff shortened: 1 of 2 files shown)


carlosflorencio:
Thanks for the contribution. I think the fix is in the right area, but the new guard looks a bit too broad.

`maybeAdoptSessionOnTransition` reads state after the session has already been upserted, so a real `RUNNING -> COMPLETED` transition and a replay of an already-terminal session both look like “pinned active session is terminal” at this point. That means this can also block the existing workflow handoff case where the old pinned session genuinely finishes and the UI should follow the replacement running session.

Could you narrow the guard by checking the previous stored state? For example, pass `existingSession?.state` into `maybeAdoptSessionOnTransition` and skip the handoff only when the pinned session was already terminal before this event. I’d keep the old “pinned active session reaches terminal and hands off” test, then add a separate replay test where the previous state is already `COMPLETED` and `setActiveSessionAuto` should not fire.

Issue #1585, opened 2026-07-03
[Bug]: Executor state desync (v0.71.0) — live agents stuck as 'starting' (pid/port 0) & turns never completed → request timeouts + sessions stuck 'running'


point-source:
## Summary

On **v0.71.0** (headless), agent executors get permanently desynced from the persisted `executors_running` state. A running, healthy agent is recorded in the DB as `status='starting'` with `pid=0` / `agentctl_port=0` forever, and turns opened against it are never marked complete (`completed_at` stays `NULL`). This one lifecycle-event-persistence defect surfaces as **two** user-visible symptoms:

1. **Sending a new message to an idle/`WAITING_FOR_INPUT` session times out** with:
   ```
   Request timed out. The agent may be processing a complex task. Please try again.
   ```
   (the same symptom as #1578, which was closed 2026-07-02 — it still reproduces on 0.71.0).
2. **Sessions/turns appear "running" in the UI forever** after the agent has actually finished and gone idle, and never resolve.

A full server restart temporarily fixes both (it rebuilds executor↔session routing in memory), but the persisted rows are never corrected, so the problem recurs and the stale rows accumulate indefinitely.

## Affected area

Agent lifecycle / executor state persistence (`executors_running`, `task_session_turns`, `task_sessions` state machine).

## Kandev version

v0.71.0

## Install mode

Homebrew (linux-x64), running headless (`kandev --headless`) as a long-lived systemd service. Process tree: `kandev` → `kandev __backend` → `agentctl` → `claude-agent-acp` (Node ACP bridge) → `claude` (Claude Agent SDK, per-session HTTP MCP to the kandev session server).

## Environment

- OS: Linux, x86_64, headless (no browser session; accessed over HTTP)
- Backend + `agentctl` + per-session ACP bridges all healthy and listening
- Long uptime between restarts (hours to days)

## Kandev configuration

- Agent profile: Claude Agent SDK executor over ACP, per-session HTTP MCP (`http://localhost:<port>/mcp`), resumable executors (`resume_token` set).

## Task/session context

- Workspace ID: redacted
- Task ID: redacted
- Session ID: redacted

(Concrete IDs omitted intentionally; the evidence below is drawn from the live DB but scrubbed of any project-identifying data.)

## Steps to reproduce

1. Start a resumable agent-mode session; let the agent run and complete a turn (posts a final message, session goes to `WAITING_FOR_INPUT` / idle).
[...]

Issue #1578, opened 2026-07-02
Session revival after idle hits "context deadline exceeded" instead of resuming


iamcobolt:
## Summary

A KanDev task session that had already finished its turn (posted a final summary message and went idle) failed to come back to life for a new user message. Instead of resuming normally, the request errored out.

## Repro / observed sequence

1. Agent completes a turn and posts its final summary message (turn considered done).
2. Session sits idle for ~23 minutes with no further activity.
3. User sends a new message on the same task ("are you stuck?").
4. Response comes back as an error instead of a normal agent reply:
   ```
   Request timed out. The agent may be processing a complex task. Please try again.
   ```
   with `metadata.error`:
   ```
   failed to ensure session is running: agent not ready for prompt: context deadline exceeded
   ```

## Expected behavior

Sending a new message to an idle-but-completed task session should either:
- transparently restart/reattach the underlying agent session within a reasonable deadline, or
- surface a clear, actionable error ("session needs to be restarted" / a retry button) rather than a generic timeout that reads as "the agent is still stuck processing."

## Actual behavior

The revival attempt itself timed out (`context deadline exceeded`), and the error message presented to the user ("may be processing a complex task") is misleading — the agent wasn't processing anything, the underlying session process/container had simply gone away or become unready, and re-establishing it exceeded whatever deadline is used internally.

## Notes

- This happened after the session had been idle for ~20+ minutes post-completion, which suggests some idle-timeout/teardown of the underlying execution environment (e.g. sandbox/container) that isn't gracefully handled on the next inbound message.
- Task/session IDs available on request if useful for server-side log correlation (this was reported from the affected session's own transcript).

carlosflorencio:
Let me know if https://github.com/kdlbs/kandev/pull/1581 fixes the issue. I'll include this in tomorrow's release.

PR #1594, opened 2026-07-03
fix(lifecycle): promote executors_running to 'ready' on agent ready (part of #1585)

iamcobolt:
## Summary

Adds the missing `executors_running` → `ready` state transition so a session wedged in `STARTING` can self-heal.

**Part of #1585** (executor state desync). This is the smallest, self-contained slice — the STARTING self-heal — not the whole issue (see "Scope" below).

## Root cause (producer/consumer mismatch)

`shouldHealStuckStartingSession` (`orchestrator/task_operations.go`) unsticks a `STARTING` session → `WAITING_FOR_INPUT` when its `executors_running` row reports `status == "ready"`. But **nothing ever wrote `"ready"`** — the row is created `"starting"` (`lifecycle/persistence.go: buildRunningFromExecution`) and never promoted, and there wasn't even a `"ready"` constant (only `"starting"` / `"prepared"`). So the heal condition was never true: a healthy agent's session stayed wedged in `STARTING` until a server restart.

## Fix (minimal)

- Add `models.ExecutorRunningStatusReady = "ready"` and use the constant at the consumer (was a bare literal with no producer — the exact drift that caused this).
- Promote the row from `markReadyEventWithContext` — which fires on **boot-ready** and on **every turn-complete** — via the **existing** narrow setter `UpdateExecutorRunningStatus` (no new SQL/method). The helper is nil-safe and swallows `ErrExecutorRunningNotFound`, so it never fails the ready transition.

## Scope (deliberately tight — ~110 lines)

`status` is the **only** `executors_running` column any consumer reads (`shouldHealStuckStartingSession`); `pid` / `agentctl_port` / `agentctl_url` / `last_seen_at` are read by nothing today, so this doesn't write them — they belong with the follow-ups that would consume them.

Not in this PR (separate follow-ups, each needs its own consumer):
- Stamp `task_session_turns.completed_at` on turn completion.
- Startup GC/reconcile of orphaned `starting` rows (which would introduce a `last_seen_at` reader).
- Re-initialize live recovered instances so idle-message resume works post-restart (dormant today — every `RecoverInstances` returns `nil`).

Complementary to #1582 (resume-readiness context fix), which addresses the "context deadline exceeded" timeout half of the same #1578/#1585 cluster. The two touch `task_operations.go` in different functions (~240 lines apart) and merge cleanly.

## Validation

`go build ./...`, the `lifecycle` / `sqlite` / `orchestrator` test packages, and changed-file `golangci-lint --new-from-rev=origin/main` all pass. New manager-helper test asserts the promote-to-`ready`, the no-writer/empty-session no-ops, and not-found swallowing.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/1594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/agent/runtime/lifecycle/manager_interaction.go
+++ b/apps/backend/internal/agent/runtime/lifecycle/manager_interaction.go
@@ -1113,6 +1113,13 @@ func (m *Manager) markReadyEventWithContext(ctx context.Context, executionID, ev
 
 	m.executionStore.UpdateStatus(executionID, v1.AgentStatusReady)
 
+	// Promote the durable executors_running row to 'ready' + stamp last_seen_at
+	// so it reflects a live agent. Without this the row stays 'starting' forever
+	// (pid/port 0) and consumers that gate on status=='ready' treat the live
+	// agent as stale — the #1585 desync. Fires on both boot-ready and every
+	// turn-complete, giving a natural liveness heartbeat.
+	m.promoteExecutorRunningReady(ctx, execution)
+
 	m.logger.Info("execution ready",
 		zap.String("execution_id", executionID),
 		zap.String("event_type", eventType))
--- a/apps/backend/internal/agent/runtime/lifecycle/persistence.go
+++ b/apps/backend/internal/agent/runtime/lifecycle/persistence.go
@@ -30,6 +30,12 @@ type ExecutorRunningWriter interface {
 	// DeleteExecutorRunningBySessionID removes the row when an execution is
 	// torn down. Idempotent: no-op if no row exists.
 	DeleteExecutorRunningBySessionID(ctx context.Context, sessionID string) error
+
+	// UpdateExecutorRunningStatus narrowly updates the status column (and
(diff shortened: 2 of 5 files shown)


iamcobolt:
Closing as superseded — #1587 (merged) already persists live executor runtime state and adds the `ExecutorRunningStatusReady`/`Running` transitions, which covers the STARTING self-heal this PR targeted (and more). Thanks @carlosflorencio — arrived at the same root cause independently and yours is the more complete fix.

One tiny leftover, not worth its own PR: the consumer at `orchestrator/task_operations.go:1655` still compares against a bare `"ready"` literal rather than `models.ExecutorRunningStatusReady` — easy to fold into a future cleanup if desired.

PR #1662, opened 2026-07-10
fix: show pending-question icon in task lists without opening the task

### LINKED ISSUE #1657: [Bug]: Sidebar task list and kanban cards don't show the agent's pending-question icon until the task is opened
### Summary

When an agent asks a clarification question (or requests tool permission), the task's row in the sidebar task list and its kanban card only show the yellow "waiting for input" question icon **after the task has been opened once**. On a fresh page load, a task that is blocked on a question is indistinguishable from a task that finished its turn — both show the green review check — so users have to click into every task to find out which ones are actually blocked.

### Affected area

Web UI

### Kandev version

Observed on v0.75.0; code paths confirmed unchanged on `main` (78d13394, post-v0.77.0).

### Install mode

npm install (self-hosted, Linux), also reproducible in local dev.

### Steps to reproduce

1. Start a task whose agent asks a clarification question (e.g. Claude Code's `AskUserQuestion`) and leave the task **unopened**.
2. Reload the browser page (or open it fresh in another tab/device) without clicking the task.
3. Look at the task's row in the sidebar task list and its kanban card.

### Expected behavior

The blocked task shows the yellow `IconMessageQuestion` ("waiting for input") indicator in the sidebar and on the kanban card, without having to open it.

### Actual behavior

The task shows the green review check (or a plain state icon). Only after clicking the task once — which loads its messages — does the question icon appear, and it then also survives reloads only while the messages stay in the store.

### Root cause

The "needs input" indicators are derived **client-side from loaded messages**, but messages are only fetched when a task is opened:

- `TaskStateIcon` (`apps/web/components/task/task-item.tsx`) decides via `shouldUseQuestionTaskIcon(state, hasPendingClarification)` / `shouldUsePermissionTaskIcon(hasPendingPermission)`.
- Those flags come from `buildPendingFlags(state.messages.bySession, …)` (`apps/web/components/task/task-session-sidebar.tsx` → `task-session-sidebar-aggregate.ts`), which runs `hasPendingClarification()` / `hasPendingPermissionRequest()` over `messages.bySession[sessionId]`.
- `messages.bySession` is only populated when a session's chat is opened (`hooks/domains/session/use-session-messages.ts`) or via WS for already-subscribed sessions. For every unopened task it is `undefined`, so all pending flags read `false`.
- The kanban card has the same gap through `useTaskPendingClarification(task.primarySessionId)` (`apps/web/components/kanban-card-content.tsx`).

The data the list views *do* have at boot (`TaskDTO.primary_session_state` from the boot payload, `GET /api/v1/workflows/:id/snapshot`, and task list endpoints) cannot disambiguate: a session sits at `WAITING_FOR_INPUT` **both** when it finished a turn normally (`setSessionWaitingForInput` also moves the task to REVIEW) and when it is genuinely blocked on a pending `clarification_request` / `permission_request` message. So the snapshot alone can't drive the icon, and the message-derived flags aren't there until the task is opened.

### Proposed fix

Denormalize the pending state onto the task snapshot so list surfaces can render it without loading messages:

- Backend: add a nullable `primary_session_pending_action` (`"clarification" | "permission"`) to `TaskDTO`, computed by one batch query over `task_session_messages` (pending `clarification_request` wins, else latest `permission_request` still pending) at the existing DTO assembly sites (`buildTaskDTOsWithSessionInfo` in `apps/backend/internal/task/handlers/task_http_handlers.go`, which also serves the workflow/workspace snapshots, and `bootStateBuilder.taskDTOsWithSessionInfo` in `apps/backend/internal/backendapp/boot_state.go`), plus the `task.updated` WS payload (`addTaskSessionEventFields`). Only sessions currently in `WAITING_FOR_INPUT` are queried, so the extra cost is one indexed query per list call.
- Frontend: in the sidebar/kanban flag resolution, prefer the live message-derived flags once a session's messages are loaded, and fall back to the snapshot field otherwise — additionally gated on `primarySessionState === "WAITING_FOR_INPUT"` so a stale snapshot flag self-clears when the session moves on.

I have this working locally and will open a PR shortly.

jean-losi:
A task blocked on an agent clarification/permission request looked identical to a finished one in the sidebar task list and on kanban cards until it was opened once, because the "needs input" indicator was derived exclusively from client-loaded messages; the backend now denormalizes the blocking state onto the task snapshot so list surfaces show the icon on a fresh page load.

## Important changes

- New `TaskDTO.primary_session_pending_action` (`"clarification" | "permission"`, omitted when none), computed by one batch query over `task_session_messages` (pending clarification wins, else latest-permission-wins — mirroring the web client's message-derived semantics). Only primary sessions currently in `WAITING_FOR_INPUT` are queried.
- Populated at the existing DTO assembly sites (task list/get + workflow/workspace snapshots via `buildTaskDTOsWithSessionInfo`, boot payload via `bootStateBuilder`) and in task lifecycle WS event payloads.
- Sidebar rows and kanban cards prefer the live message-derived pending flags once a session's messages are loaded, and fall back to the snapshot field otherwise — gated on `primarySessionState === "WAITING_FOR_INPUT"` so a stale snapshot flag self-clears when the session moves on. Sessions that finished a turn normally (also `WAITING_FOR_INPUT`, no pending request) keep the green check.

## Validation

- `go test ./internal/task/... ./internal/backendapp/...` — green, incl. a new repo test for `GetPendingSessionActions` (pending/missing status counts as pending, stale pending permission superseded by a decided one does not, clarification wins over permission).
- `cd apps/web && pnpm run typecheck` clean; `pnpm exec vitest run components/task lib/ssr lib/kanban lib/ws hooks src/boot-payload.test.ts` — 2287 tests green, incl. new cases for `resolvePendingFlags`, `useTaskPendingClarification`, and `toKanbanTask` field parity.
- New e2e spec `tests/task/sidebar-pending-question.spec.ts`: seeds a mock-agent `/e2e:clarification` task that is **never opened**, plus a quietly-finished control task, reloads on an unrelated task, and asserts the blocked row shows `task-state-waiting-for-input` while the control row does not. Verified it **fails on `main` without this fix** (icon absent) and passes with it.
- Neighboring specs re-run green: `chat/clarification.spec.ts`, `chat/permission-approval.spec.ts`, `kanban/kanban-board.spec.ts` (29 passed).

## Possible improvements

Low risk. The snapshot field is only refreshed with task lifecycle events, so a question that arrives while the page is already open still won't flag a never-opened task until a task event or reload; wiring `pending_action` into `session.state_changed` publishers would close that residual gap.

Closes #1657

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/1662?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/backendapp/boot_state.go
+++ b/apps/backend/internal/backendapp/boot_state.go
@@ -623,6 +623,11 @@ func (b bootStateBuilder) taskDTOsWithSessionInfo(ctx context.Context, tasks []*
 		b.logBootError("get task detail primary session info", err)
 		return taskDTOs(tasks)
 	}
+	pendingBySession, err := b.p.taskSvc.GetPendingActionsForWaitingSessions(ctx, primaryInfoByTask)
+	if err != nil {
+		b.logBootError("get task detail pending session actions", err)
+		pendingBySession = map[string]taskmodels.SessionPendingAction{}
+	}
 	result := make([]taskdto.TaskDTO, 0, len(tasks))
 	for _, task := range tasks {
 		if task == nil {
@@ -643,7 +648,7 @@ func (b bootStateBuilder) taskDTOsWithSessionInfo(ctx context.Context, tasks []*
 			sessionCount = &count
 		}
 		info := bootSessionInfo(primaryInfoByTask[task.ID])
-		result = append(result, taskdto.FromTaskWithSessionInfo(
+		taskDTO := taskdto.FromTaskWithSessionInfo(
 			task,
 			primarySessionID,
 			sessionCount,
@@ -654,7 +659,14 @@ func (b bootStateBuilder) taskDTOsWithSessionInfo(ctx context.Context, tasks []*
 			info.agentName,
 			info.workingDirectory,
 			info.sessionState,
-		))
+		)
+		if primary := primaryInfoByTask[task.ID]; primary != nil {
+			if action, ok := pendingBySession[primary.ID]; ok {
+				val := string(action)
+				taskDTO.PrimarySessionPendingAction = &val
+			}
+		}
+		result = append(result, taskDTO)
 	}
 	return result
 }
--- a/apps/backend/internal/backendapp/boot_state_routes.go
+++ b/apps/backend/internal/backendapp/boot_state_routes.go
@@ -540,21 +540,22 @@ func mapKanbanTaskState(task taskdto.TaskDTO) map[string]any {
 		})
 	}
 	return map[string]any{
-		"id":                  task.ID,
-		"workflowStepId":      task.WorkflowStepID,
-		"title":               task.Title,
-		"description":         task.Description,
-		"position":            task.Position,
-		"state":               task.State,
-		"repositoryId":        primaryRepositoryID,
-		"repositories":        repositories,
-		"primarySessionId":    task.PrimarySessionID,
-		"primarySessionState": task.PrimarySessionState,
-		"sessionCount":        task.SessionCount,
-		"reviewStatus":        nullString(string(task.ReviewStatus)),
-		"parentTaskId":        nullString(task.ParentID),
-		"updatedAt":           task.UpdatedAt,
-		"createdAt":           task.CreatedAt,
+		"id":                          task.ID,
+		"workflowStepId":              task.WorkflowStepID,
+		"title":                       task.Title,
+		"description":                 task.Description,
+		"position":                    task.Position,
+		"state":                       task.State,
+		"repositoryId":                primaryRepositoryID,
+		"repositories":                repositories,
+		"primarySessionId":            task.PrimarySessionID,
+		"primarySessionState":         task.PrimarySessionState,
+		"primarySessionPendingAction": task.PrimarySessionPendingAction,
+		"sessionCount":                task.SessionCount,
+		"reviewStatus":                nullString(string(task.ReviewStatus)),
+		"parentTaskId":                nullString(task.ParentID),
+		"updatedAt":                   task.UpdatedAt,
+		"createdAt":                   task.CreatedAt,
 	}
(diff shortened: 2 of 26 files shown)


carlosflorencio:
I compared this PR against #1663, which is the parallel fix for the same issue.

My recommendation is to consolidate onto #1663 and port the useful E2E coverage from this PR, rather than merging this branch as-is.

The main differences I see:

- #1663 scopes pending clarification/permission detection to the latest session turn. This avoids showing a pending-question icon from stale pending rows left behind by an older turn. This PR currently scans pending action rows across broader session history, which is the main correctness risk.
- #1663 sends `primary_session_pending_action: null` explicitly in DTOs. That gives the frontend a reliable distinction between known clear and field omitted by an older payload.
- #1663 covers more frontend update paths: task cards, the mobile task switcher, agent-session kanban sync, and `kanban.update` preservation.
- #1663 treats the task-list/boot pending-action lookup as best-effort, logging and falling back instead of letting the supplemental lookup break task loading.
- This PR currently has unresolved review threads and a failed frontend formatting check, while #1663 is in a cleaner state.

The thing I would keep from this PR is the E2E regression in `apps/web/e2e/tests/task/sidebar-pending-question.spec.ts`. It exercises the actual reload/sidebar behavior and is valuable coverage. I think the best path is to port that test into #1663, then use #1663 as the consolidation branch.

carlosflorencio:
If you don't mind i will close this PR because its already merged. Will create a new release soon.

PR #1816, opened 2026-07-20
fix: clear stale archived state when unarchiving from task detail top bar

dk-blackfuel:
## Problem

Unarchiving a task from the **task-detail top bar** returned a success toast, but the task **never came back** — the archived UI (Unarchive button + archived banner) stayed, and the task appeared stuck as archived.

## Root cause

The detail view seeds `archived_at` from a **one-shot `fetchTask`** (`useTaskDetails` in `task-page-content.tsx`) that only re-fetches when the **task id changes**. Unarchive publishes `task.updated` with `archived_at=null`, which the WS handler correctly uses to re-add the task to `state.kanban.tasks` (the board only ever holds non-archived tasks).

But `resolveEffectiveTask` → `mergeBaseWithKanban` spread the stale fetch and overrode only title/description/step/position/state — **never `archived_at`**. So `buildArchivedValue` kept returning `isArchived: true`, and `TaskTopBar` kept rendering the archived UI. No refetch ever cleared it.

## Fix

In `mergeBaseWithKanban`, clear `archived_at` when a matching kanban entry exists — a task present on the board is by definition not archived (archive removes it; unarchive re-adds it). This trusts the live store over the stale one-shot fetch, matching the Unarchive button's documented design ("restores the task in the kanban/store — no manual refetch needed").

The three pure resolver helpers (`resolveEffectiveTask` / `mergeBaseWithKanban` / `buildTaskFromKanban`) were relocated from the heavy `task-page-content.tsx` into the already-unit-tested `task-page-content-helpers.ts` so the fix is directly covered (net −65 lines in the `.tsx`).

## Testing

- **Unit** (`task-page-content-helpers.test.ts`): 3 new regression tests — stale `archived_at` cleared when in kanban, preserved when absent (still archived), live kanban fields still win. All 15 in file pass; the 3 new ones fail without the fix.
- **E2E** (`e2e/tests/task/unarchive-detail-topbar.spec.ts`): archives a task, opens its detail page, unarchives, asserts the Unarchive button detaches in place and the route is unchanged. Runs in CI — the backend binary can't be built in this environment (no Go toolchain), so it wasn't executed locally.
- `tsc --noEmit`, `eslint`, and `prettier --check` clean on all changed files.
- Adjacent suites (`task-top-bar`, `tasks-unarchive`, `tasks` WS handlers): 26 pass.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/1816?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/web/components/task/task-page-content-helpers.test.ts
+++ b/apps/web/components/task/task-page-content-helpers.test.ts
@@ -1,12 +1,47 @@
 import { describe, expect, it, vi } from "vitest";
 import type { Task } from "@/lib/types/http";
+import type { KanbanState } from "@/lib/state/slices";
 import {
+  buildArchivedValue,
   buildDebugEntries,
   hasResolvedTaskDetails,
+  resolveEffectiveTask,
   resolveTaskContentState,
   resolveTaskProps,
   syncActiveTaskSession,
 } from "./task-page-content-helpers";
+
+type KanbanTask = KanbanState["tasks"][number];
+
+function makeArchivedTaskDetails(overrides: Partial<Task> = {}): Task {
+  return {
+    id: "task-1",
+    title: "Archived task",
+    description: "",
+    workflow_step_id: "step-1",
+    position: 0,
+    state: "TODO",
+    workspace_id: "ws-1",
+    workflow_id: "wf-1",
+    priority: 0,
+    repositories: [],
+    created_at: "",
+    updated_at: "2026-07-19T00:00:00Z",
+    archived_at: "2026-07-19T00:00:00Z",
+    ...overrides,
+  } as Task;
+}
+
+function makeKanbanTask(overrides: Partial<KanbanTask> = {}): KanbanTask {
+  return {
+    id: "task-1",
+    title: "Restored task",
+    workflowStepId: "step-1",
+    position: 0,
+    state: "TODO",
+    ...overrides,
+  } as KanbanTask;
+}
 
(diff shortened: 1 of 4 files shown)


carlosflorencio:
Thanks for the contribution!

There’s one case I think needs addressing before merge: `mergeBaseWithKanban` clears `archived_at` whenever a kanban card exists, but those cards can be stale. Hydration doesn’t remove cards missing from a newer snapshot, and the workflow snapshot may skip refetching. In that situation, `fetchTask` can confirm that a task is archived while the stale card makes the detail page render it as active.

Could we avoid using card presence alone as the archive signal? One option is to clear the detail state explicitly after a successful unarchive. Another is to carry an explicit, freshness-aware archive state through the WS/store model. It would also be useful to add the inverse regression test: fetched archived details plus a stale kanban card should remain archived.

One smaller follow-up: the kanban-only fallback can copy workspace/workflow IDs from task details that belong to a different task. That behavior predates this PR, so it doesn’t need to block this fix, but it’s worth avoiding while these helpers are being made testable.

PR #1817, opened 2026-07-20
fix(plugins): serialize empty sync-result arrays instead of null

jcfs:
## Summary
- `POST /api/plugins/sync` returned `null` instead of `[]` for `added`/`installed`/`missing`/`errors` whenever a category was empty, because Go's `encoding/json` marshals a nil slice as `null` and `Sync()` only appends to these slices, leaving them nil when nothing changed.
- The frontend's `SyncResult` type declares these fields as non-nullable arrays, and `summarizeSyncResult()` reads `result.added.length` unconditionally — so a routine "everything up to date" sync crashed the Plugins settings page with `can't access property "length", e.added is null`.
- `Sync()`/`bootScan()` now normalize any nil slice field to an empty array before returning.

## Test plan
- [x] `go test ./internal/plugins/...` — added `TestServiceSync_EmptyResultSerializesEmptyArraysNotNull`, which JSON-marshals a no-op sync result and asserts it contains no `null`
- [x] `go build ./...` and `gofmt -l` clean on changed files

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/1817?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/plugins/service_sync.go
+++ b/apps/backend/internal/plugins/service_sync.go
@@ -58,47 +58,35 @@ func (s *Service) sync(ctx context.Context, includeTarballs bool) *SyncResult {
 	s.syncMu.Lock()
 	defer s.syncMu.Unlock()
 
-	result := &SyncResult{}
+	// Slices start non-nil: encoding/json marshals a nil slice as JSON null,
+	// but SyncResult's fields are typed as non-nullable arrays on the
+	// frontend (apps/web/lib/types/plugins.ts) — a null there crashes
+	// summarizeSyncResult's unconditional `.length` reads. append()-ing a
+	// possibly-nil result onto these keeps them non-nil.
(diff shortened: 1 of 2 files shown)


jcfs:
Addressed the suggestion from Claude and Greptile: `SyncResult`'s slices are now initialized non-nil at the struct literal in `sync()` instead of normalized at each return path, and the regression test now checks exact JSON field values instead of a `strings.Contains("null")` substring search. See 987abcbae.

PR #1829, opened 2026-07-20
chore(make): add cmd.exe-safe start-windows target

JnManso:
On native Windows `make start` fails — the root target relies on `printf`, `find`, `cp`, and `exec` (absent under cmd.exe), while the backend build needs cmd.exe for its `set VAR=VAL&` env syntax, so no single shell satisfies both. A cmd.exe-safe `start-windows` target, following the existing `test-windows` pattern, lets Windows users build and run from source.

## Important Changes

- `start-windows`: installs deps, builds web, syncs embedded assets, builds the backend, and starts the server using only cmd.exe-safe commands.
- `sync-embedded-web-windows`: mirrors the Vite `dist` with `robocopy /MIR` (`/XF` keeps the committed `.gitignore` and `keep.txt`) instead of `find`/`cp`; robocopy's sub-8 "success" exit codes are normalized to 0 so make does not read them as failures.

## Validation

- Windows 11, GNU Make via cmd.exe: `make start-windows` builds and boots the backend end-to-end (confirmed against `/api/v1/workspaces`).
- `make sync-embedded-web-windows`: preserves `.gitignore`/`keep.txt`, purges stale assets, and is idempotent across re-runs; the missing-`dist` guard fails the build cleanly.
- `make help` lists the target; `python3 .github/scripts/lint-harness-files.py --all` passes.

## Possible Improvements

Low risk — purely additive and Windows-only; no existing target is modified. Adjacent and out of scope: the backend Makefile's `adhoc_sign_darwin` (Windows branch) uses `echo (…)`, valid under cmd.exe but a syntax error under sh, so building under Git bash breaks — which is why this target must run where Make invokes cmd.exe. It could be hardened to work under both shells independently.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/1829?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,7 @@ help:
 	@echo "  start            Install deps, build, and start backend + web in production mode"
 	@echo "  start-verbose    Start in production mode with info logs from backend + web"
 	@echo "  start VERBOSE=1  Same as start-verbose"
+	@echo "  start-windows    cmd.exe-safe start for native Windows (no printf/find/cp/exec)"
 	@echo ""
 	@echo "Service Commands:"
 	@echo "  service-install          Install deps, build current checkout, install user service"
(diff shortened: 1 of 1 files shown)


JnManso:
Thanks for the reviews — applied all three:

- **Executable path (line 274)** and **robocopy paths (sync-embedded-web-windows)**: converted to backslashes via `Invalid parameter - /,\,...`. FWIW forward slashes did work in my end-to-end runs, but backslashes are unambiguous for cmd.exe and robocopy, so no reason not to.
- **Guard error message**: good catch — it pointed at `make build-web`, which itself isn't cmd.exe-safe. Now points at `make start-windows`.

Re-verified after the change: `make start-windows` builds and boots the backend end-to-end, and the sync target preserves `.gitignore`/`keep.txt` while purging stale assets. Also fixed the PR title type (`build` → `chore`) for the conventional-commits check.

PR #1967, opened 2026-07-26
fix(agents): relax OpenCode install detection

carlosflorencio:
## Summary
- Change OpenCode ACP discovery to treat any installed `opencode` binary on `PATH` as installed.
- Stop running `opencode --version` as a required installation gate in `IsInstalled()`.
- Keep pinned install instruction for `opencode-ai@1.18.4` as guidance, but do not fail discovery for newer/older binaries.
- Update discovery compatibility notes so version mismatches are handled by runtime ACP probe/auth path, and `--version` remains metadata-only.

## Why
Kandev was incorrectly reporting OpenCode as unavailable when the installed CLI version did not exactly match `1.18.4` (for example `1.18.5`), even though models were usable in runtime ACP flow.

## Validation
- `go test ./internal/agent/agents -run TestOpenCodeACPDiscovery -count=1`
- `go test ./internal/agent/agents`

## Notes
- Commit created with conventional message and targeted tests.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/1967?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/agent/agents/ACP_BRIDGE_VERSIONS.md
+++ b/apps/backend/internal/agent/agents/ACP_BRIDGE_VERSIONS.md
@@ -16,11 +16,10 @@ part of `AgentCommand`, which existing lifecycle diagnostics log as
 OpenCode intentionally uses the direct `opencode acp` command for discovery,
 normal sessions, container commands, and one-shot inference. This keeps startup
 offline-compatible and ensures discovery validates the executable actually
-launched. Its installer remains pinned to `opencode-ai@1.18.4`, and discovery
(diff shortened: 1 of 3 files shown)


carlosflorencio:
@codex review

PR #1988, opened 2026-07-27
fix(github): stop SELECT * scans from breaking on TaskPR schema drift

ClemDNL:
## Problem

Linking a GitHub PR to a task (`Link GitHub pull request` dialog) failed with:

```
associate PR with task: missing destination name workspace_id in *github.TaskPR
```

That's a Go/sqlx error, not a data problem — it means the `github_task_prs` table
had a column the running binary's `TaskPR` struct had no field for. sqlx's
`StructScan` rejects any `SELECT *` result column with no matching destination
field.

That mismatch is not hypothetical: `github_task_prs.workspace_id` and
`TaskPR.WorkspaceID` were added together, but a self-update that applies a
newer release's migration (adding a column) followed by a rollback to an
older binary leaves the column permanently in place — SQLite migrations are
additive and never revert. Any future column addition reproduces this same
failure class, not just `workspace_id`.

Confirmed the affected task's `workspace_id` itself was valid and pointed at
a real, non-archived workspace — the error text just uses "missing
destination name" to mean "no matching Go struct field", which reads as
"something is missing" but isn't a data-integrity issue at all.

## Fix

`internal/github/store.go`: replaced every `SELECT * FROM github_task_prs` /
`SELECT gtp.*` TaskPR read with an explicit column projection
(`taskPRColumns` / `taskPRColumnsQualified`), so an unexpected extra column
no longer breaks:

- `GetTaskPR`
- `GetTaskPRByRepository`
- `GetTaskPRByRepoAndNumber`
- `ListTaskPRsByTask`
- `ListTaskPRsByTaskIDs`
- `ListTaskPRsByWorkspaceID`

Writes (`ReplaceTaskPR`, `CreateTaskPR`, `UpdateTaskPR`) already used
explicit columns and needed no change.

## Testing

Added `TestTaskPRReads_ToleratesExtraColumn`
(`store_taskpr_schema_drift_test.go`), which:

1. Simulates schema drift via `ALTER TABLE github_task_prs ADD COLUMN
   future_only_column ...`.
2. RED: before the fix, `GetTaskPR` failed with `missing destination name
   future_only_column in *github.TaskPR` — the exact reported error class.
3. GREEN: after the fix, all six read paths succeed and return the correct
   PR data despite the extra column.

- `make fmt` — clean
- `make build` — clean
- `make test` — `internal/github` package passes; the only backend-wide
  failures are pre-existing `internal/task/service` sandbox
  filesystem-permission failures (`parent directory cannot be accessed`),
  confirmed unrelated by reproducing them against the unmodified base
  commit.
- `make lint` (golangci-lint) — 0 issues

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/1988?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Demo evidence (browser-driven, real UI)

Reproduced the reported condition end-to-end against a running instance: built this branch,
started the backend fresh, stopped it, ran `ALTER TABLE github_task_prs ADD COLUMN
future_only_column ...` directly on the SQLite file to simulate the schema-drift condition,
restarted, then drove the actual "Link GitHub pull request" dialog through a real browser.

1. Dialog opened from the task card's Link → GitHub Pull Request menu:

   Dialog open

2. PR URL entered:

   PR URL entered

3. Saved successfully — no sqlx error, despite the extra `future_only_column` still present
   in `github_task_prs`:

   Linked successfully

Verified server-side too: the `github_task_prs` row persisted with correct
`task_id`/`workspace_id`/`owner`/`repo`/`pr_number`/`pr_title`, with
`future_only_column` still sitting there unused.

(Screenshots hosted on a separate `screenshots/pr-1988` branch on the fork — not part of
this PR's diff.)

### DIFF
--- a/apps/backend/internal/github/store.go
+++ b/apps/backend/internal/github/store.go
@@ -1395,11 +1395,35 @@ func (s *Store) CreateTaskPR(ctx context.Context, tp *TaskPR) error {
 	return err
 }
 
+// taskPRColumns is the explicit column list for every github_task_prs read.
+// Using `SELECT *` here is unsafe: sqlx's StructScan errors out
+// ("missing destination name X in *github.TaskPR") the moment the table has
+// a column the current TaskPR struct doesn't declare a field for. That drift
+// is not hypothetical — a self-update that applies a newer release's
+// migration (adding a column) followed by a rollback to an older binary
+// leaves exactly this mismatch permanently on disk, since SQLite migrations
+// are additive and never reverted. Projecting the known columns keeps every
+// read working regardless of what the table has picked up beyond them.
+const taskPRColumns = `id, workspace_id, task_id, repository_id, owner, repo, pr_number, pr_url,
+	pr_title, head_branch, base_branch, author_login, state, review_state, checks_state,
+	mergeable_state, review_count, pending_review_count, required_reviews, comment_count,
+	unresolved_review_threads, checks_total, checks_passing, additions, deletions,
+	created_at, merged_at, closed_at, last_synced_at, updated_at`
+
+// taskPRColumnsQualified is taskPRColumns with each column qualified by the
+// `gtp` alias, for queries that join github_task_prs against another table.
+const taskPRColumnsQualified = `gtp.id, gtp.workspace_id, gtp.task_id, gtp.repository_id, gtp.owner, gtp.repo,
+	gtp.pr_number, gtp.pr_url, gtp.pr_title, gtp.head_branch, gtp.base_branch, gtp.author_login,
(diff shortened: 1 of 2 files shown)


ClemDNL:
Not sure about this one

PR #2007, opened 2026-07-28
fix(worktree): stop planting self-referential links in local repositories

### LINKED ISSUE #2005: Windows: Local tasks plant a self-referential junction inside the repo, and owned links are rejected on every relaunch
Two distinct bugs in the workspace-source link code, found together while investigating the first. They can be split into separate issues if preferred.

Measured on Windows 11, Go 1.26.0, kandev @ `2850086db`.

---

## Bug 1 — a Local-executor task plants a self-referential junction inside the user's own repository

**Reproduce:** start a task with executor **Local** on a local repository. A directory named after the repository appears *inside* that repository, and it is a directory junction pointing back at the repository itself.

```
D:\repos\my-service\
  └── my-service  -->  D:\repos\my-service
```

`git status` goes from clean to `?? my-service/` the moment it is created (measured). ripgrep/fd, IDE indexers and file watchers descend into it.

### Cause

1. `manager_launch.go:369-372` — for a non-worktree executor, `launchResolveWorkspacePath` sets `workspacePath = req.RepositoryPath`. **The user's repository becomes the workspace root.**
2. `manager_launch.go:1047-1051` — `reconcileWorkspaceRepositories(workspacePath, …)` is then called with `root == req.RepositoryPath`.
3. `types.go:634-658` — `RepoSpecs()` synthesizes a length-1 spec from the single-repo fields, so the **primary** repository is in the list.
4. `workspace_sources_reconcile.go:36-56` — `reconcileWorkspaceRepositories` has no guard for "this spec *is* the root", so it calls `EnsureOwnedDirectoryLink(root, RepoName, RepositoryPath)` → `<repo>/<RepoName>` → `<repo>`.

The same happens on resume via `manager_execution.go:471-478`, with specs from `service_turns.go:718-753`, which also returns index 0.

The `worktree` executor is unaffected — `manager_launch.go:363-365` returns early, and the call is gated to `local`/`local_pc`.

### Notes

`buildRemoteWorkspaceRepositories` (`workspace_source_materializer.go:479-486`) already skips the primary for exactly this reason.

A **positional index-0 skip would be wrong** here, though: for a host-materialized multi-repo local task, `materializeHostWorkspaceSources` re-roots the workspace at `worktreeMgr.TaskRoot(...)` (`workspace_source_materializer.go:137-142`) and `localWorkspaceEntries` links every repo including index 0 — there `repositories[0]` is a genuine sibling that must keep its link. The comparison has to be filesystem identity.

This also contradicts the feature's own stated invariant, `docs/plans/attach-workspace-sources/plan.md:85`: *"Never copy, move, delete, or write marker files into user-owned local sources."*

Unix takes the same code path through `os.Symlink` (`directory_link_unix.go:7`) — **not verified on Unix**, but a symlink is ignored by git and most walkers, so the impact there should be much smaller.

---

## Bug 2 — on Windows, `EnsureOwnedDirectoryLink` rejects every unchanged link, so relaunch and resume fail

Measured with a plain sibling link (`root` and `target` distinct directories — the ordinary multi-repo case):

```
1st EnsureOwnedDirectoryLink(root, "api", target): created=true  err=<nil>
2nd EnsureOwnedDirectoryLink(root, "api", target): created=false err=owned link target mismatch: api
```

### Cause

`directory_link.go:91-101` compares `filepath.EvalSymlinks(link)` against `filepath.EvalSymlinks(target)`. **`filepath.EvalSymlinks` does not traverse a Windows junction** — it returns the link's own path, merely normalized. So `actual` is `<root>\api` while `expected` is the target directory, and the comparison always mismatches.

Measured behaviour of a junction created by `directory_link_windows.go` (Go 1.26.0):

| call | result on a junction |
|---|---|
| `os.Lstat(link).Mode()` | `ModeIrregular` — the `os.ModeSymlink` bit is **false** |
| `filepath.EvalSymlinks(link)` | the link's own path, normalized (does **not** follow) |
| `os.Readlink(link)` | the target |
| `os.Stat(link)` | follows the junction |
| `os.SameFile(Stat(link), Stat(target))` | `true`; `false` for a different directory |

`reconcileWorkspaceRepositories` propagates that error, so **any relaunch or resume of a task that has workspace source links fails on Windows**. The same comparison is reached from `reconcileWorkspaceSources` (folders) and from `materializeDirectoryLinks` (`workspace_source_materializer.go:203`, `:569`).

This is also why the Bug 1 junction gets stuck: once planted, the next launch errors out before anything can clean it up.

---

## Proposed fix

1. **Guard** in `reconcileWorkspaceRepositories`: skip a spec whose repository is the workspace root, compared by filesystem identity (`os.Stat` + `os.SameFile`), not by index and not by path text.
2. **Self-heal**: a narrowly-scoped `worktree.RemoveSelfReferentialDirectoryLink(root, name)` that removes the entry only when it is a platform directory link whose target *is* `root`. Uses `os.Remove`, never `os.RemoveAll` — on Windows `RemoveDirectory` unlinks the junction and leaves the target untouched; if the entry has become a real directory it fails with `ENOTEMPTY` instead. Needed because `EnsureOwnedDirectoryLink` would otherwise never correct an existing self-link.
3. **`EnsureOwnedDirectoryLink`**: match the existing link by `os.SameFile` instead of comparing `EvalSymlinks` output. This also removes the 8.3-short-path and path-case fragility for free.

Verified against a real repository on Windows: the junction is removed and `HEAD`, the top-level entry count and `git status` are all identical to the pre-test baseline.

## Secondary findings, not addressed

- **Multi-repo local without host materialization.** With the guard, `repositories[0]` is skipped but `repositories[1..n]` still get junctions created *inside* `repositories[0]` when `req.WorkspacePath == ""`. The real question is whether any multi-repo local launch reaches `launchResolveWorkspacePath` without going through materialization; if so, the fix is to route it to the task root, not to add more guards.
- **Attached folders.** `reconcileWorkspaceSources` (`workspace_sources_reconcile.go:14-34`) has the same missing root-equality guard — attaching the repository's own folder to a local task would self-link identically. Not confirmed reachable.
- **Test fragility.** `TestReconcileWorkspaceRepositories_RecreatesMissingOwnedLink` asserted with `os.Readlink`, which returns the long form and never string-equals a `t.TempDir()` on a host whose `TMP` carries an 8.3 component (e.g. `C:\Users\JOHNDO~1`). It failed on Windows before this change.

JnManso:
Starting a task with the Local executor left a directory junction pointing at the repository from inside the user's own checkout, and git, ripgrep and IDE indexers all descended into it. On Windows a second defect made that state permanent — every relaunch was rejected before anything could clear it — so both are fixed here. A link planted by an earlier release is reported rather than deleted, and the user removes it.

## Important Changes

- `reconcileWorkspaceRepositories` skips a spec whose repository *is* the workspace root. The comparison is filesystem identity, not index: a host-materialized multi-repo local task roots the workspace at `~/.kandev/tasks/<taskDir>`, where `repositories[0]` is a real sibling that must keep its link.
- `EnsureOwnedDirectoryLink` now matches an existing link with `os.SameFile` instead of comparing `filepath.EvalSymlinks` output. `EvalSymlinks` does not traverse a Windows junction — it returns the link's own path — so every unchanged link was rejected as a target mismatch. This affects all four call sites; the new comparison is strictly stronger (identity rather than path text) and also drops the 8.3 short-path and path-case fragility.
- New read-only `worktree.IsSelfReferentialDirectoryLink(root, name)` reports an entry that is a platform directory link whose target is `root` itself; `reconcileWorkspaceRepositories` logs it with shell-neutral cleanup guidance and leaves it in place. Nothing is deleted: Kandev writes no ownership marker into user-owned sources, so such an entry cannot be shown to be ours — a user, or the repository itself, may keep a link of that name and target on purpose — and a stat-then-remove sequence could not close the window between the check and the unlink. Preventing new self-links is what fixes the reported bug.

## Validation

- `CGO_ENABLED=0 go test ./internal/worktree/... ./internal/agent/runtime/lifecycle/` — the 11 new and 1 repaired tests pass. Remaining failures in those packages are pre-existing, verified by running the same packages in a clean `git worktree` at `HEAD` and diffing the failure lists. `internal/worktree` goes 28 → 29, and the one delta, `TestBranchExists_RespectsContextDeadline`, also fails on the clean baseline (1 of 3 runs). `internal/agent/runtime/lifecycle` goes 13 → 12, the difference being `TestReconcileWorkspaceRepositories_RecreatesMissingOwnedLink`, repaired here.
- `go vet` and `golangci-lint run` on both packages — 0 issues.
- Red-first: with the guard reverted, 3 of the 4 new reconcile tests fail and the 4th (the task-root non-regression case) still passes. The `EnsureOwnedDirectoryLink` defect reproduces against unmodified code as `created=true` on the first call and `owned link target mismatch` on the second.
- Manual, Windows 11: planted a junction in a real repository with `mklink /J` and confirmed `git status` went from clean to an untracked recursive entry, which is the reported symptom. That run also exercised the removal path this PR no longer performs; `HEAD`, the top-level entry count and `git status` were all identical to the pre-test baseline afterwards, so the unlink touched only the link. The shipped behaviour reports the entry instead.

Two caveats worth flagging. Tests ran with `CGO_ENABLED=0` because cgo cannot link on my machine (an unrelated `ld.exe` problem with a non-ASCII character in the profile path), so the `fts5` path is only covered by CI. And I have no Unix machine here — on Unix `filepath.EvalSymlinks` does follow symlinks, so the old comparison worked and the new one should be equivalent, but that is reasoning rather than measurement.

## Possible Improvements

Low risk overall; the widest blast radius is the `EnsureOwnedDirectoryLink` comparison, shared by four call sites. Two related findings are documented in the issue but deliberately left out of scope: `reconcileWorkspaceSources` has the same missing root-equality guard for attached folders, and a multi-repo local launch that bypasses host materialization would still link siblings inside `repositories[0]`.

## Checklist

Closes #2005


JnManso:
@carlosflorencio flagging this directly, since it grew past what I expected when I opened the issue.

The reported symptom was the first bug — a Local-executor task planting a directory junction inside the user's own repository. While fixing it I found a second one that I think is the more serious of the two: on Windows `filepath.EvalSymlinks` does not traverse a junction, it returns the link's own path. So `EnsureOwnedDirectoryLink` rejected *every* unchanged link as a target mismatch, and each relaunch or resume of a task with workspace source links failed. That is also why the first bug's junction never got cleaned up — the next launch errored out before anything could clear it.

Three things are your call:

1. **Scope.** Both fixes are in one PR because they live in the same pair of functions and one explains the other. The second is independent though, and I am happy to split it out if you would rather review them separately.
2. **`EnsureOwnedDirectoryLink`.** The fix changes how it matches an existing link — `os.SameFile` rather than comparing `EvalSymlinks` output. Four call sites share it. The new comparison is strictly stronger (filesystem identity instead of path text), but it is still a semantic change to a fail-closed check, so it deserves your eyes more than the rest of the diff.
3. **Two findings left out of scope**, both written up in #2005: `reconcileWorkspaceSources` has the same missing root-equality guard for attached folders, and a multi-repo local launch that bypasses host materialization would still link siblings inside `repositories[0]`. Either can come into this PR or go separately — I did not want to widen a bug fix on my own judgement.

One caveat on validation: cgo cannot link on my machine, an unrelated `ld.exe` problem with the non-ASCII character in my profile path, so I ran everything with `CGO_ENABLED=0` and the `fts5` path is only covered by CI. I also have no Unix machine here — `EvalSymlinks` does follow symlinks on Unix, so the old comparison worked there and the new one should be equivalent, but that is reasoning rather than something I measured.

JnManso:
Applied all four points from the automated reviews in `b71372d47`.

**Codex P2 — non-directory repository path.** Real regression I introduced, thanks. Comparing by filesystem identity alone matched a repository path that had been replaced by a regular file, so the skip branch bypassed the `IsDir()` validation and a launch could proceed with a workspace path that is not a directory. `sameDirectory` now requires both sides to be directories, so that case falls through to the existing target check. Covered by `TestReconcileWorkspaceRepositories_RejectsFileAsRepositoryPath`.

**Copilot — `.` and `..` in `RepoName`.** Correct, they survive the `filepath.Base` round-trip and were only caught deeper in the owned-link helpers, surfacing as a confusing "owned link entry already exists". Extracted `isWorkspaceEntryName` and used it for both repositories and folders, so the two validations stay consistent. Covered by `TestReconcileWorkspaceRepositories_RejectsTraversalRepoName`.

**Copilot — skips hiding regressions.** Agreed, and my reasoning for the skip was wrong: a junction needs no elevation on Windows and a symlink is always available on the Unix runners, so a creation failure there is a regression, not an unsupported platform. All four are now hard failures. The one remaining skip in the file is pre-existing and genuinely conditional — `os.Symlink` for the control-ancestor test does need Developer Mode.

**Copilot — stale error message.** Fixed: `canonicalize link target` → `inspect link target`, since `os.Stat` no longer canonicalizes.

`go vet` and `golangci-lint` clean; the reconcile and directory-link tests pass, now 9 and 9.

JnManso:
@coderabbitai good catch on the entry-name validation, applied in `0ebee3728` — with one correction to the finding.

Measured on Windows before changing anything: of the cases raised, only the bare separator was actually accepted. `filepath.Base("/")` returns `\` on Windows, so `/` was already rejected, and `C:` was too. On Unix the accepted case would be `/`. The predicate now rejects a bare root on either platform, plus any name carrying a volume.

On the severity, though — I do not think this was a Security & Privacy issue. Every path failed closed: with a root as the entry name, `filepath.Join` resolves back to the workspace root, `os.Lstat` finds a real directory rather than a link, and `EnsureOwnedDirectoryLink` returns `owned link entry already exists`. Nothing was created, traversed or overwritten. The real defect was that the rejection happened three frames below the validation that exists for exactly this, so the error was misleading. Worth fixing, but as clarity rather than as a vulnerability.

Also added `TestIsWorkspaceEntryName` as a table test over the predicate. That gap is what let this through: `TestReconcileWorkspaceRepositories_RejectsTraversalRepoName` only asserted that reconciliation errored, which passed even while the name was accepted here and rejected further down — so it could not distinguish the two.

carlosflorencio:
Thanks for picking this up — I’ve closed #2009 in favor of this PR. I compared the latest head (`0ebee372`) against #2009, and this is the stronger base, especially because it includes real Windows validation and broader launch/reconciliation coverage.

A few items still need resolving before merge:

1. **Legacy-link ownership:** `RemoveSelfReferentialDirectoryLink` cannot distinguish a Kandev-created legacy link from a user-created or Git-tracked `<workspace>/<RepoName>` link that also points to the repository. Automatically deleting it can remove user content. The safe default is to prevent new self-links but preserve an existing entry and surface targeted manual-cleanup guidance unless durable Kandev ownership can be proven. See the [ownership finding](https://github.com/kdlbs/kandev/pull/2009#discussion_r3667433244).
2. **Check/remove race:** the `Lstat`/`Stat` checks followed by `os.Remove(link)` have a TOCTOU window. The entry can be replaced with an unrelated empty directory before removal, and `os.Remove` will delete it. If automatic cleanup remains, unlinking must be tied to the validated identity/handle; process-local locking alone would not protect against external filesystem changes. See the [TOCTOU finding](https://github.com/kdlbs/kandev/pull/2009#discussion_r3667427593).
3. **Latest Unix test failure:** at `0ebee372`, `go test ./internal/worktree ./internal/agent/runtime/lifecycle` fails in `TestIsWorkspaceEntryName` and `TestReconcileWorkspaceRepositories_RejectsTraversalRepoName`. The tests reject `\`, `C:`, and `a\b` unconditionally, but those are legal single-component names on Unix. The Windows-only cases need OS-specific expectations/tests.
4. **Cleanup errors:** `reconcileWorkspaceRepositories` currently discards `RemoveSelfReferentialDirectoryLink` errors. If cleanup remains, a failed unlink should not silently report success while leaving the recursive entry behind.

The five older unresolved bot threads appear addressed by the newer commits (directory validation, error wording, and fatal link-creation tests).

JnManso:
Thanks for the thorough pass, and for closing #2009 in favour of this — all four points addressed.

**1 & 2 & 4 — auto-cleanup is gone** (`61abc62a4`). You are right that ownership cannot be established: Kandev deliberately writes no marker into user-owned sources, so there is nothing to check against, and a link of that name and target is exactly what someone would create by hand as a shortcut. `RemoveSelfReferentialDirectoryLink` is replaced by the read-only `IsSelfReferentialDirectoryLink`, and `reconcileWorkspaceRepositories` now logs the entry together with the command to remove it. That also dissolves the discarded-error path in 4, since there is no longer a cleanup that can fail.

On the TOCTOU specifically — my earlier justification was wrong and worth correcting for the record. I argued `os.Remove` was safe because a real directory would fail with `ENOTEMPTY`; that only holds for a directory with content in it. An empty one would have been deleted, exactly as you describe. Preventing new self-links is what actually fixes the reported bug; clearing an existing one costs the user a single command and risks nothing.

**3 — Unix test failure** (`658686507`). Mine, and a plain oversight: I derived the rejected set from measurements taken on Windows and then asserted it unconditionally. `\`, `C:` and `a\b` are legal single-component names on Unix and the implementation accepts them correctly there — only the test was wrong. `TestIsWorkspaceEntryName` now splits its expectations on `runtime.GOOS`, and `TestReconcileWorkspaceRepositories_RejectsTraversalRepoName` is narrowed to names invalid on every platform.

Worth noting I cannot run the Unix suite locally, so that fix is verified by CI rather than by me.

The PR is now: prevent new self-links, report pre-existing ones, and match owned links by filesystem identity so Windows relaunches stop failing. Happy to split the `EnsureOwnedDirectoryLink` change out if you would still rather review it separately.

JnManso:
Both new findings addressed.

**Copilot — stale PR description.** Correct, and my miss: I changed the behaviour to report-only after @carlosflorencio's ownership point and left the description claiming the link is removed on the next launch. Description updated to match what the code does; the Validation section now also distinguishes what the manual Windows run actually proved from the behaviour that ships.

**CodeRabbit — shell-interpretable guidance** (`e0c7c4003`). The underlying point is right: `%q` is Go quoting, not shell quoting, so a path carrying `$()`, backticks or `%VAR%` would still be expanded by whatever shell the line was pasted into. Rather than add per-shell escaping I dropped the commands entirely — the entry path now goes out as a structured log field, which is never shell-interpreted, and the guidance describes the removal instead of spelling it out. That also removes the need for the metacharacter test cases, since there is no longer a command to escape.

On the severity label, same note as last time: I do not think this was Security & Privacy. Nothing was executed, and the value only reached a log field; the risk was a user pasting a line we had formatted badly. Worth fixing, but as output hygiene.

The two low-confidence notes are noted and skipped as cosmetic: the duplicated junction rationale across docblocks is deliberate, since each reader hits one of the two packages without the other, and the second `os.Stat` after `sameDirectory` is on a cold path once per repository per launch.

carlosflorencio:
Thanks for the contribution and for working through the safety and portability feedback. One last coverage gap before merging: the legacy self-link test passes a `nil` logger, so it verifies preservation but not that users receive the cleanup warning.

Could we pass an observer logger and assert the warning and structured entry path? Something along these lines:

```go
core, logs := observer.New(zapcore.WarnLevel)
log, err := logger.NewFromZap(zap.New(core))
if err != nil {
    t.Fatal(err)
}

err = reconcileWorkspaceRepositories(root, repositories, log)
if err != nil {
    t.Fatal(err)
}

entries := logs.FilterMessage(
    "workspace entry links to the workspace root; remove that entry to stop tools recursing into it",
).All()

if len(entries) != 1 {
    t.Fatalf("warnings = %d, want 1", len(entries))
}
if got := entries[0].ContextMap()["entry"]; got != filepath.Join(root, "api") {
    t.Fatalf("entry = %v", got)
}
```

Please keep the existing assertion that the link remains untouched as well. That would cover the complete “preserve and warn” behavior promised by the PR.

JnManso:
Done in `40726b30b` — good catch, the `nil` logger was a shortcut and it left untested exactly the half of "preserve and warn" that the user acts on.

`TestReconcileWorkspaceRepositories_PreservesAndReportsPreExistingSelfLink` now uses the observer, asserting one warning carrying the structured `entry` path, and keeps the assertions that the link and the repository content are untouched. I added the negative case alongside it — `..._DoesNotWarnWithoutSelfLink`, a repository that is the workspace root with no stale entry must stay silent. The pair is self-validating: one demands exactly one warning, the other demands none, so a warning that stopped firing or started firing spuriously fails one of them.

One deviation from your snippet: I extracted the message into a `selfReferentialEntryWarning` constant shared by the warning and the assertions, rather than repeating the literal in the test. A copy would have gone stale the moment anyone reworded the log line, and the test would have kept passing while asserting a message nobody emits.

11 of 11 reconcile tests pass locally on Windows; `vet` and `golangci-lint` clean.

PR #2014, opened 2026-07-28
fix(backend): return bad request on malformed sprites WS payloads

jcfs:
## Problem

Three sprites WebSocket handlers discarded the `ParsePayload` error:

- `wsStatus`
- `wsListInstances`
- `wsTest`

Their siblings in the same file (`wsDestroyInstance`, `wsGetNetworkPolicy`, `wsUpdateNetworkPolicy`) return a bad-request response instead. The three outliers carried on with a zero-valued `SecretID`, so a client sending a malformed frame got a confusing downstream failure (or a lookup against an empty secret ID) rather than a clear `BAD_REQUEST`.

## Change

All three now match the sibling shape:

```go
if err := msg.ParsePayload(&payload); err != nil {
    return ws.NewError(msg.ID, msg.Action, ws.ErrorCodeBadRequest, "invalid payload", nil)
}
```

These were the only `_ = msg.ParsePayload(...)` occurrences in `internal/sprites` — none remain in the package.

## Empty/absent payloads are preserved

`Message.ParsePayload` short-circuits on a nil payload:

```go
func (m *Message) ParsePayload(v interface{}) error {
    if m.Payload == nil {
        return nil
    }
    return json.Unmarshal(m.Payload, v)
}
```

So a frame with no payload still succeeds and the handler answers for the zero-valued secret ID exactly as before — only genuinely malformed JSON now errors. **No handler was deliberately left as-is.**

Frontend impact is nil: `grep -rn "sprites\." apps/web/lib/ws` finds nothing, and a repo-wide search for the `sprites.status` / `sprites.instances.*` / `sprites.test` / `sprites.network_policy.*` action strings shows the only references are the backend `RegisterFunc` calls. The web app talks to sprites over HTTP (`apps/web/lib/api/domains/sprites-api.ts` → `/api/v1/sprites/*`), which has its own separate handlers that are unchanged.

## Tests

`internal/sprites/handlers_test.go`:

- `TestWSHandlersRejectMalformedPayload` — all **six** WS handlers × three malformed payload shapes (truncated JSON, wrong field type, non-object) assert `MessageTypeError` + `BAD_REQUEST` + `"invalid payload"`, and that the request ID/action are echoed back. The three siblings are included as a pin so the shape can't drift apart again.
- `TestWSStatusAcceptsValidPayload` / `TestWSListInstancesAcceptsValidPayload` / `TestWSTestAcceptsValidPayload` — happy path for each changed handler, run with both a valid payload and an absent payload. `wsListInstances` asserts the failure it does surface is `INTERNAL_ERROR` ("API token not configured"), i.e. it got past parsing.

The handler under test is built with a nil secret store, so every path short-circuits before any network call.

Verified red before the fix: reverting `handlers.go` fails exactly the 9 `status` / `instances_list` / `test` subtests while the three sibling handlers stay green.

## Validation

- `make -C apps/backend lint` → **0 issues**
- `go test ./internal/sprites/...` → **ok**
- `make -C apps/backend test` → the sprites package passes. Four unrelated packages fail in this sandbox (`internal/gitlab`, `internal/notifications/service`, `internal/task/handlers`, `internal/task/service`) with errors like `invalid directory creation: parent directory cannot be accessed`. This diff touches only `internal/sprites`, so those files are byte-identical to `main` — pre-existing/environmental, not caused by this change.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2014?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/sprites/handlers.go
+++ b/apps/backend/internal/sprites/handlers.go
@@ -184,7 +184,9 @@ func (h *Handler) wsStatus(ctx context.Context, msg *ws.Message) (*ws.Message, e
 	var payload struct {
 		SecretID string `json:"secret_id"`
 	}
-	_ = msg.ParsePayload(&payload)
+	if err := msg.ParsePayload(&payload); err != nil {
+		return ws.NewError(msg.ID, msg.Action, ws.ErrorCodeBadRequest, "invalid payload", nil)
+	}
 	return ws.NewResponse(msg.ID, msg.Action, h.getStatus(ctx, payload.SecretID))
 }
 
@@ -192,7 +194,9 @@ func (h *Handler) wsListInstances(ctx context.Context, msg *ws.Message) (*ws.Mes
 	var payload struct {
 		SecretID string `json:"secret_id"`
 	}
-	_ = msg.ParsePayload(&payload)
+	if err := msg.ParsePayload(&payload); err != nil {
+		return ws.NewError(msg.ID, msg.Action, ws.ErrorCodeBadRequest, "invalid payload", nil)
+	}
 	instances, err := h.listInstances(ctx, payload.SecretID)
 	if err != nil {
 		return ws.NewError(msg.ID, msg.Action, ws.ErrorCodeInternalError, err.Error(), nil)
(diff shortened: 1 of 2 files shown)


jcfs:

Focus areas:
- `apps/backend/internal/sprites/handlers.go` — three WS handlers (`wsStatus`, `wsListInstances`, `wsTest`) now return `ws.ErrorCodeBadRequest` on a `ParsePayload` error instead of discarding it. Is the behaviour change safe for callers that send an absent/empty payload? (`Message.ParsePayload` returns nil early when `m.Payload == nil`.)
- `apps/backend/internal/sprites/handlers_test.go` — do the tests actually pin the behaviour, or would they pass against the old code?

Note: the automatic review on this PR completed without filing any findings, so this is a re-request.

PR #2015, opened 2026-07-28
fix(backend): return real results from Service.GetBlocking

jcfs:
`Service.GetBlocking` returned a hardcoded empty slice behind a TODO, so every caller asking "which tasks does this task block?" silently got a wrong answer; it now delegates to the reverse-lookup query that already existed on the blocker repository.

## Important Changes

- The TODO asked for a query that was already implemented: `ListTasksBlockedBy` on the office SQLite repository, already declared on the `BlockerRepository` interface in the same file as `GetBlocking`, and already used in production by the office scheduler (`cascadeBlockersResolved`), the office event subscribers (`queueBlockersResolvedRuns`), and `HandoffService`'s related-tasks builder. No new repository method, interface change, or test-double update was needed — `GetBlocking` just never called it.
- `ListTasksBlockedBy` now orders by `created_at, task_id` (mirroring `ListTaskBlockers`) and returns an empty non-nil slice rather than `nil`.
- Archived tasks are deliberately **not** filtered out. `GetBlockers` — the symmetric forward lookup directly above it — returns raw edges, and no existing `ListTasksBlockedBy` consumer filters on `archived_at`; filtering only the reverse direction would make the two asymmetric. Documented on the method and pinned by `TestGetBlocking_IncludesArchivedTasks`.
- `GetBlocking` has no call sites outside tests (verified repo-wide across Go, TS/TSX and docs), so the change from "always empty" to real data cannot break a consumer. The `BlockedBy` data the UI does render already went through `ListTasksBlockedBy` directly and is unchanged.

## Validation

- `make -C apps/backend test lint` — lint reports `0 issues`.
- New service coverage in `service_office_test.go`: `TestGetBlocking_ReverseLookup` (none / one / several blocked tasks, edge removal, and that the forward and reverse directions aren't confused), `TestGetBlocking_NoRepositoryConfigured`, `TestGetBlocking_IncludesArchivedTasks`.
- New repository coverage in `blockers_test.go`: `TestListTasksBlockedBy` (empty non-nil result, one, several with deterministic ordering, a task that only appears on the blocked side, and delete removing the edge).
- Three local test failures — `internal/gitlab` `TestEnvironmentTokenAllowsImmutableStartupHost` and two `internal/notifications/service` delivery-occurrence tests — reproduce identically on a clean `HEAD` worktree and pass in CI, so they are environment-specific and unrelated to this change.

## Possible Improvements

Low risk. Worth noting separately: `task_blockers` has no foreign key to `tasks` and nothing removes its rows on task deletion, so dangling edges are possible in both lookup directions — pre-existing behaviour, not addressed here.

## Checklist

### DIFF
--- a/apps/backend/internal/office/repository/sqlite/blockers.go
+++ b/apps/backend/internal/office/repository/sqlite/blockers.go
@@ -40,12 +40,18 @@ func (r *Repository) DeleteTaskBlocker(ctx context.Context, taskID, blockerTaskI
 }
 
 // ListTasksBlockedBy returns task IDs that are blocked by the given task.
+// This is the reverse direction of ListTaskBlockers; results are ordered by
+// insertion time so callers see a stable list.
 func (r *Repository) ListTasksBlockedBy(ctx context.Context, blockerTaskID string) ([]string, error) {
 	var ids []string
 	err := r.ro.SelectContext(ctx, &ids, r.ro.Rebind(
-		`SELECT task_id FROM task_blockers WHERE blocker_task_id = ?`), blockerTaskID)
+		`SELECT task_id FROM task_blockers WHERE blocker_task_id = ? ORDER BY created_at, task_id`),
+		blockerTaskID)
 	if err != nil {
 		return nil, err
+	}
+	if ids == nil {
+		ids = []string{}
 	}
 	return ids, nil
 }
--- a/apps/backend/internal/office/repository/sqlite/blockers_test.go
+++ b/apps/backend/internal/office/repository/sqlite/blockers_test.go
@@ -2,6 +2,7 @@ package sqlite_test
 
 import (
 	"context"
+	"slices"
 	"testing"
 
 	"github.com/kandev/kandev/internal/office/models"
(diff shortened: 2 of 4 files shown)


jcfs:
@codex review

jcfs:
@claude review this PR. The auto-review run posted only "I'll analyze this and get back to you" without any findings, so please do a real pass.

Focus areas:
- `GetBlocking` now returns the raw reverse blocker edges without filtering archived tasks, for symmetry with `GetBlockers`. Is that the right call, or should it filter?
- `ListTasksBlockedBy` gained `ORDER BY created_at, task_id` and now normalizes an empty result to a non-nil slice. It has three existing production callers (office scheduler `cascadeBlockersResolved`, office event subscribers `queueBlockersResolvedRuns`, and `HandoffService`) — does either change affect them?
- Test coverage adequacy in `service_office_test.go` and `blockers_test.go`.

jcfs:
Thanks — agreed on points 1, 2 and 4. Pushing back on **point 3**: the `nil` check in `GetBlocking` is not dead code, and removing it breaks a test.

The reasoning assumes `ListTasksBlockedBy` has one implementation. It's an interface method on `BlockerRepository` (`service_office.go:20`), and `s.blockers` is that interface — not the concrete SQLite repo. The non-nil guarantee added in this PR is a property of `*office/repository/sqlite.Repository`, not of the interface. The other two implementations both return `nil` on empty:

```go
// mockBlockerRepo, service_office_test.go:341
var ids []string
for _, b := range m.blockers { if b.BlockerTaskID == blockerTaskID { ids = append(ids, b.TaskID) } }
return ids, nil   // nil when nothing matches

// fakeBlockerRepo, handoff_workspace_test.go:198 — same shape
```

`TestGetBlocking_ReverseLookup` exercises exactly that path: it asserts a non-nil empty slice on the first call, against `mockBlockerRepo`, before any blocker exists. Removing the guard:

```
--- FAIL: TestGetBlocking_ReverseLookup (0.05s)
    service_office_test.go:388: expected non-nil empty slice, got nil
FAIL    github.com/kandev/kandev/internal/task/service   0.145s
```

So the guard is what makes `GetBlocking`'s non-nil contract hold for *any* `BlockerRepository`, rather than only for the SQLite one. That's deliberate: the service returns `[]string{}` because `GetBlockers` directly above it does (`make([]string, len(blockers))` is never nil), and callers shouldn't have to care which implementation is wired in.

One clarification on the supporting sentence — "`GetBlocking` propagates the nil result before reaching this check" — the only early return above it is the `err != nil` path, which returns `nil, err`. On the success path nothing propagates before the check; that's the branch under test.

Keeping it as-is. The repository-level normalization stays too: it makes the SQLite path self-consistent so a future direct caller of `ListTasksBlockedBy` gets the same shape.

PR #2017, opened 2026-07-28
fix(backend): bound the launcher health probe with a timeout

jcfs:
## Problem

`waitForHealth` in `apps/backend/internal/launcher/health.go` probed the backend with:

```go
resp, err := http.Get(healthURL) //nolint:gosec,noctx
```

`http.Get` uses `http.DefaultClient`, which has **no timeout**, and carries no context. If the backend accepts the TCP connection but never writes a response, that call blocks forever: the poll loop never reaches its next deadline check, the 45s health timeout never fires, and the launcher hangs with no way to cancel it. That is precisely the failure the suppressed `noctx` lint was pointing at.

## Change

- **Bounded client.** New package-level `healthProbeClient` with an explicit `Timeout` (2s), so a single stalled probe fails fast and the loop retries on the next 300ms tick.
- **Context-carrying request.** `http.NewRequestWithContext(ctx, http.MethodGet, healthURL, nil)` replaces `http.Get`. Each request inherits the overall wait deadline, so a probe never outlives the wait.
- **Context threaded from the caller.** `launcher.Run` creates the root context (the launcher's process entry point) and passes it through `runStart` / `runInstalled` → `runManagedApp` → `waitForHealth`. The exported `launcher.Run` signature is unchanged. `waitForHealth` derives the overall deadline with `context.WithTimeout` instead of comparing wall-clock times, and the poll sleep now selects on `ctx.Done()` — so cancellation ends the wait promptly and reports distinctly from a timeout.
- **Body drained and closed.** `io.Copy(io.Discard, resp.Body)` before `Close()` so the keep-alive connection is reused across poll iterations rather than being dropped on every non-2xx response.
- **`//nolint:gosec,noctx` removed** — neither suppression is needed any more.

## Tests

New `health_test.go` covers:

| Test | Asserts |
| --- | --- |
| healthy response | returns nil; also exercises one non-2xx iteration first, so body-drain + connection reuse across polls is on the path |
| non-200 response | keeps polling, then fails with a timeout error and calls `onFailure` once |
| **hanging server** | server accepts the connection and never responds — the probe returns via timeout instead of blocking (guarded by a 10s watchdog that fails the test if it hangs) |
| context cancellation | a one-minute wait against a hanging server ends immediately on `cancel()`, with a cancellation-flavored error |
| backend already exited | fails fast with the exit code, without probing |
| `healthTimeout` env handling | default on unset/non-numeric/zero/negative; override honored |

## Validation

```
make -C apps/backend lint    # 0 issues
go test -race ./internal/launcher/...   # ok
make -C apps/backend test
```

The full backend suite has pre-existing failures in `internal/gitlab`, `internal/notifications/service`, `internal/task/handlers`, and `internal/task/service` (e.g. `parent directory cannot be accessed`). They reproduce unchanged on the base commit `e5237b26c` in a clean worktree, and `go list -deps` confirms none of those packages import `internal/launcher`. `internal/launcher` and `internal/launcher/cli` both pass.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2017?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/launcher/health.go
+++ b/apps/backend/internal/launcher/health.go
@@ -1,12 +1,25 @@
 package launcher
 
 import (
+	"context"
+	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"strconv"
 	"time"
 )
+
+const (
+	healthPollInterval = 300 * time.Millisecond
+	healthProbeTimeout = 2 * time.Second
+)
+
+// healthProbeClient bounds every individual health request. http.DefaultClient
+// has no timeout, so a backend that accepts the connection but never answers
+// would otherwise block the launcher forever.
+var healthProbeClient = &http.Client{Timeout: healthProbeTimeout}
 
 type childState interface {
 	Exited() (bool, int)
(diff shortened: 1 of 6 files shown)


jcfs:
@coderabbitai review

PR #2018, opened 2026-07-28
refactor(backend): share the native-binary ACP agent scaffold

jcfs:
## What

Eight ACP agent definitions in `apps/backend/internal/agent/agents/` were near-identical ~110-160 line copies of the same `Agent` implementation. Seven opened with a file-level `//nolint:dupl` (two also `goconst`) whose comment said the quiet part out loud:

> Native-binary ACP agents (Cursor, Kimi, Kiro, Qoder, Trae, Omp, Devin) follow the same minimal scaffold; differences are the binary name, argv, and auth surface.

This extracts that scaffold.

- **`native_acp.go`** — `nativeACPSpec` declares everything that varies (identity, `bin` + `args`, logos, install script, remote auth, permission settings, plus a `nativeACPRuntimeSpec` for the `RuntimeConfig` fields that differ: skill dirs, session dir, MCP overrides, `StripEnv`, `RequiredEnv`). `nativeACPAgent` derives the rest — `WorkingDir`, resource limits, `ProtocolACP`, native session resume, `IsInstalled` via `WithCommand(bin)`, `InferenceConfig`, billing — and `BuildCommand`, `Runtime().Cmd` and `InferenceConfig().Command` all come from one argv instead of three hand-kept-in-sync copies.
- **Passthrough** is a separate embed (`nativeACPPassthroughAgent`). Grok stays on the bare `nativeACPAgent` so it still does **not** satisfy `PassthroughAgent` (ADR 0014) — a pre-existing test pins that, and the new table does too.
- **Shared literals** — `"CLI Passthrough"`, `"Show terminal directly instead of chat interface"` and `"{workspace}"` now live in one place, which is what the `goconst` suppressions were complaining about.
- **All `//nolint:dupl` / `//nolint:goconst` directives removed.**

`claude_acp.go`, `codex_acp.go`, `copilot_acp.go` and `opencode_acp.go` are untouched — they're substantially different and not part of the cluster.

Net: **-786 / +310** lines across the eight agent files.

## Proving it's a pure refactor

Before touching anything I dumped the complete registered surface of all eight agents — identity, all three launch argvs, full `RuntimeConfig` (including `RequiredEnv` nil-vs-empty and `ProjectMCPStrategy` concrete type), `SessionConfig`, `PassthroughConfig`, six `BuildPassthroughCommand` option combinations, `RemoteAuth`, install script, `PermissionSettings`, `CatalogPermissionSettings`, `BillingType`, skill dirs, and which optional interfaces each type satisfies. The dump is **byte-identical before and after** (modulo `*bool` addresses).

That scaffolding was temporary. The permanent safety net is **`native_acp_test.go`**, a table-driven pin over all eight agents covering identity, the three launch surfaces, runtime, session config, passthrough config, permission settings, install script and remote auth. Mutation-checked: swapping `projectSkillDir`/`userSkillDir` in the shared `Runtime()` fails it on cursor, kiro and omp.

## Validation

- `make -C apps/backend lint` — 0 issues, with the suppressions gone.
- `make -C apps/backend test` — the agents packages pass. Four unrelated packages fail (`internal/gitlab`, `internal/notifications/service`, `internal/task/handlers`, `internal/task/service` — local-repository-init and notification-occurrence tests). I confirmed the **identical** failure set on a clean `HEAD` worktree, so they're pre-existing/environmental, not from this change.

## Notes for review

**The `//nolint` directives were already stale.** I checked: at `HEAD`, with the directives stripped, `dupl` and `goconst` both report **0 issues** on this package under the repo config. `dupl` at `threshold: 150` only starts flagging these files around threshold 60 — and when it does, it points at `qwen_acp.go` / `kilocode_acp.go`, not the suppressed cluster. So removing them is safe, but the refactor stands on readability rather than on un-suppressing a live warning. (There is still a real `dupl` cluster among the *npm-launched* ACP agents — `qwen`, `iflow`, `droid`, `kilocode`, `pi` — that a follow-up could give the same treatment.)

**Per-agent quirks preserved verbatim, not "fixed"** (flagging rather than changing, per the refactor's scope):
- Kiro, Qoder and Trae have no `SessionDirTemplate`, so `NativeSessionResume` has no persistence across container restarts. Their TODOs moved with them.
- Devin is the only passthrough agent with no `--model` flag.
- Grok sets `RequiredEnv: []string{}` where every peer leaves it nil. The distinction is preserved and asserted (`reflect.DeepEqual`, not `slices.Equal`) since it reads as deliberate documentation of "auth is optional".
- Cursor's `PermissionSettings` are the cluster's only non-empty ones, and the only `cli_flag`.

## Docs

`docs/add-agent-cli.md` gains a "Native-Binary ACP Agent" section — this is now the shortest path for a non-npm ACP CLI, and new agents should add a `nativeACPCases` entry.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2018?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/agent/agents/cursor_acp.go
+++ b/apps/backend/internal/agent/agents/cursor_acp.go
@@ -1,14 +1,9 @@
-//nolint:dupl // Native-binary ACP agents (Cursor, Kimi, Kiro, Qoder, Trae) follow the same minimal scaffold; differences are the binary name and subcommand.
 package agents
 
 import (
-	"context"
 	_ "embed"
-	"time"
 
 	"github.com/kandev/kandev/internal/agent/mcpconfig"
-	"github.com/kandev/kandev/internal/agent/usage"
-	"github.com/kandev/kandev/pkg/agent"
 )
 
 //go:embed logos/cursor_acp_light.svg
@@ -18,6 +13,18 @@ var cursorACPLogoLight []byte
 var cursorACPLogoDark []byte
 
 const cursorACPBin = "cursor-agent"
+
+// cursorACPInstallScript installs cursor-agent, which isn't on npm. The
+// official installer drops the binary into ~/.local/bin; make sure that dir is
+// on PATH for the rest of the prepare script and for future shells on the
+// sprite.
+const cursorACPInstallScript = `set -e
+tmp="$(mktemp)"
+curl -fsS https://cursor.com/install -o "$tmp"
+bash "$tmp"
+rm -f "$tmp"
+export PATH="$HOME/.local/bin:$PATH"
+grep -qxF 'export PATH="$HOME/.local/bin:$PATH"' "$HOME/.bashrc" 2>/dev/null || echo 'export PATH="$HOME/.local/bin:$PATH"' >> "$HOME/.bashrc"`
 
 // cursorPermSettings maps a curated CLI flag to cursor-agent's --force switch.
 // In ACP mode Cursor emits session/request_permission for commands off its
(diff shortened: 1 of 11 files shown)


jcfs:
@coderabbitai review

PR #2048, opened 2026-07-29
fix(github): stop the PR panel from showing the previous task's stale feedback

ClemDNL:
Switching from one task's PR tab to another task's PR tab could briefly show the previous task's stale reviews/comments/checks — including whatever "Bot comments" disclosure was left expanded — because `usePRFeedback` kept the old PR's feedback in state while the new PR's fetch was still in flight.

## Important Changes

- `usePRFeedback` now keys its cached feedback by `<workspaceId>/<owner>/<repo>/<prNumber>` and masks it via `resolvePRFeedbackView` whenever the requested key changes (mirrors the existing `resolvePRDiffView` masking pattern used for the diff viewer), while still preserving cached feedback during a same-PR manual refresh so that flow doesn't flicker.
- The `{feedback && (...)}` gate in `PRDetailContent` then naturally unmounts/remounts the reviews/comments/checks subtree on a real identity change, resetting any stale local disclosure state (e.g. the bot-comments toggle) along with the data.

## Validation

- `hooks/domains/github/use-pr-feedback.test.ts` (new): unit tests for `resolvePRFeedbackView`, plus a hook test proving PR 1's feedback is masked immediately on switching to PR 2 and a late PR 1 response can't resurrect stale data. Confirmed failing before the fix, passing after.
- `components/github/pr-detail-panel.test.tsx` (new): renders `PRDetailPanelComponent`, expands "Bot comments" on task 1's PR, switches the store to task 2 while reusing the same panel instance, and asserts task 2 renders with the toggle collapsed and no leaked text. Confirmed failing before the fix, passing after.
- Added a Playwright case to `e2e/tests/pr/pr-detail-auto-show.spec.ts` and ran it against a real Chromium + real Go backend — it passes with the fix. That harness gives each task its own isolated env, so it doesn't reproduce the exact reused-panel-instance race (documented in the test's own comment, matching a precedent already in that file for a sibling bug); the two tests above are the discriminating regression coverage.
- `make fmt-web` (clean), `make typecheck` (pass), `make lint-web` (pass, 0 warnings), `make test-web` (7237/7242 pass; the 1 failure, `storage-maintenance-settings.test.tsx`, is an unrelated pre-existing timeout confirmed passing in isolation).
- No screenshots: this is a data-timing fix, not a visual/layout change — every screen it touches renders identically once the (now-correct) data has loaded.

## Possible Improvements

Low risk: behavior change is scoped to one hook's internal state machine and the render gate that already existed; no API or type surface changed for the hook's callers.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2048?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/web/components/github/pr-detail-panel.test.tsx
+++ b/apps/web/components/github/pr-detail-panel.test.tsx
@@ -1,0 +1,184 @@
+import { act, cleanup, fireEvent, render, screen, waitFor } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { StoreApi } from "zustand";
+import { StateProvider, useAppStoreApi } from "@/components/state-provider";
+import { ToastProvider } from "@/components/toast-provider";
+import { TooltipProvider } from "@kandev/ui/tooltip";
+import type { AppState } from "@/lib/state/store";
+import type { PRFeedback, TaskPR } from "@/lib/types/github";
+import { getPRFeedback } from "@/lib/api/domains/github-api";
+import { PRDetailPanelComponent } from "./pr-detail-panel";
+
+vi.mock("@/lib/api/domains/github-api", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("@/lib/api/domains/github-api")>();
+  return {
+    ...actual,
+    getPRFeedback: vi.fn(),
+  };
+});
+
+vi.mock("@/lib/ws/connection", () => ({
+  getWebSocketClient: () => null,
+}));
+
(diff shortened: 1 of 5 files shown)


ClemDNL:
While this was a real issue, the real trigger was network issue :sweat:

PR #2057, opened 2026-07-30
fix(web): stop long GitLab row titles drawing over the row actions

nova28:
## Problem

In the My GitLab panel, a long issue or MR title escapes its column and paints across the **Unsubscribe** and **Task** buttons, leaving both visually unreachable.

## Cause

The title link carried `inline-flex` together with `truncate`. An inline-flex box sizes itself to its content, so the link grew past the row's `min-w-0 flex-1` column regardless of `truncate` — `truncate` sets `overflow: hidden` on the element it is applied to, but that element had already been allowed to size beyond its parent. Hidden overflow on a box that is itself too wide clips nothing useful.

## Fix

Make the link block-level (`flex min-w-0`) so it participates in the column's width, and move truncation onto the inner text span, whose `overflow-hidden` zeroes its automatic minimum size so it can actually shrink. The external-link icon keeps `shrink-0` so it is never clipped.

Both lists carried the same markup, so it is extracted into `RowTitleLink` rather than fixed twice. The comment there records why the element must stay `flex` and not revert to `inline-flex`, since that is the part a future edit would plausibly undo.

The full title stays available on hover via the `title` attribute.

## Verification

```
vitest run components/gitlab/my-gitlab/row-title-link.test.tsx
Test Files  1 passed (1)
      Tests  2 passed (2)

tsc --noEmit   (clean)
```

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2057?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/web/components/gitlab/my-gitlab/issue-list.tsx
+++ b/apps/web/components/gitlab/my-gitlab/issue-list.tsx
@@ -1,7 +1,6 @@
 "use client";
 
-import Link from "@/components/routing/app-link";
-import { IconCircle, IconCircleCheck, IconExternalLink } from "@tabler/icons-react";
+import { IconCircle, IconCircleCheck } from "@tabler/icons-react";
 import { Badge } from "@kandev/ui/badge";
 import { Spinner } from "@kandev/ui/spinner";
 import { cn, formatRelativeTime } from "@/lib/utils";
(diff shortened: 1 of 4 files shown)


carlosflorencio:
Thanks @nova28 for the smaller PRs, i've approved the CI to run. 

After you fix the CI & Address/Resolve the review comments I'll merge them.

carlosflorencio:
FYI I'll push some commits to your PRs to address the CI, because we want to cut a new release today and include your fixes.

PR #2068, opened 2026-07-30
refactor(web): reuse stats-utils formatDuration in stats-sections

jcfs:
`stats-sections.tsx` carried a private `formatDuration` that was a byte-for-byte copy of the exported one in its sibling `stats-utils.ts`, so the two stats surfaces could silently drift apart. It now imports the shared helper instead.

## Validation

- `pnpm --filter @kandev/web lint` — clean
- `pnpm run typecheck` (apps/web) — clean
- `pnpm vitest run` (apps/web) — 964 files, 7368 passed / 4 skipped

No new test needed: `stats-utils.test.ts` already covers `formatDuration` across all four buckets (`0` → `—`, seconds, minutes, hours). No UI behaviour changes, so no new Playwright coverage.

## Possible Improvements

Negligible risk — the deleted body and the kept one were character-identical apart from `"—"` vs a literal `—`.

The other ~8 `formatDuration` implementations in `apps/web` were deliberately left alone: they genuinely differ (some render `450ms`, some `1.5s`, some `2m 30s`, and some take `(startedAt, completedAt)` strings rather than a number of ms), so collapsing them would be a real behaviour change across many surfaces. One further genuinely-identical pair is a clean follow-up — `components/task/simple/components/session-timeline-entry.tsx:14` and `components/task/simple/components/agent-turn-panel.tsx:21` are byte-for-byte the same.

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have manually tested my changes and they work as expected.
- [ ] My changes have tests that cover the new functionality and edge cases.
- [ ] If my change touches UI files (`apps/web/`), I have added or updated Playwright e2e tests in `apps/web/e2e/` and verified them with `make test-e2e`.
- [ ] I checked whether this affects public docs in `docs/public/**` and updated them or noted why no docs change is needed.

PR #2069, opened 2026-07-30
refactor(web): consolidate formatRelativeTime into lib/utils

jcfs:
Five copies of `formatRelativeTime` had drifted across `apps/web` into three different output vocabularies, so the same timestamp read differently depending on which surface rendered it — a 30-second-old item showed `30s ago` in the task list but `just now` in the sidebar, and a commit said `5 minutes ago` next to a task saying `5m ago`. `lib/utils.ts` is now the single implementation and every surface reads the same.

## Important Changes

- Deleted the private copies in `task-item.tsx` (no seconds bucket, no `yesterday`), `automations/format-utils.ts` (whole file; same divergence), `commit-detail-panel.tsx` (long form, 30-day cutoff), and `kanban/graph2-task-pipeline.tsx` (30-day cutoff plus a `Nmo ago` bucket nothing else had). All call sites now import from `@/lib/utils`.
- `commit-detail-panel.tsx` adopts the shared **short** form rather than keeping its long-form wording as a separate exported helper. It renders inside a dense one-line metadata row — `Author · 5 minutes ago · a1b2c3d` at `text-xs text-muted-foreground` — directly alongside task surfaces already using the short form, so the long wording bought nothing and was the most visible instance of the inconsistency this change exists to remove.
- The original brief listed four implementations; the fifth (`kanban/graph2-task-pipeline.tsx`) turned up during the sweep. Consolidating it was one line each way and is what "single implementation" means — leaving it would have reproduced the same bug in a fifth place. Its `Nmo ago` bucket is gone: items older than 7 days now show an absolute date like every other surface.

Out of scope by design: `formatPreciseTime` is untouched, and `formatBytes` / `formatDuration` are handled separately in #2071 and #2068.

## Validation

- `pnpm --filter @kandev/web lint` — passed
- `cd apps/web && pnpm run typecheck` — passed
- `cd apps/web && pnpm vitest run` — 964 files, 7375 passed, 4 skipped

New bucket-boundary coverage in `apps/web/lib/utils.test.ts` uses fake timers against a frozen clock and pins the 10s / 60s / 60m / 24h / `yesterday` / 48h / 7d boundaries plus the `toLocaleDateString()` fallback.

`apps/web/e2e/tests/task/sidebar-layout.spec.ts:273` asserts each sidebar row matches `/now|ago/`. The shared implementation's `yesterday` bucket matches neither, but that spec creates its tasks inline, so they render `just now` / `Ns ago` and the assertion still holds.

## Possible Improvements

Low risk — behavior-only consolidation with no API change. The visible effect is that four surfaces now use vocabulary they did not use before: the sidebar gains `Ns ago` and `yesterday`, commits switch from `5 minutes ago` to `5m ago`, and the Kanban pipeline loses `Nmo ago` in favour of an absolute date past 7 days. Any future assertion on a rendered timestamp needs to account for `yesterday`, which contains neither "now" nor "ago".

## Checklist


jcfs:
The automated review pass no-opped on this PR (`Claude finished ... in 0s` with no findings posted), so requesting a real review explicitly.

@codex review

Context for reviewers: this consolidates five drifted copies of `formatRelativeTime` into the shared one in `apps/web/lib/utils.ts`. The one genuine judgment call is `commit-detail-panel.tsx`, which previously used long-form wording (`5 minutes ago`) and a 30-day cutoff and now adopts the shared short form — rationale is in the PR description. Worth a second opinion on whether that reads correctly in the commit metadata row.

PR #2071, opened 2026-07-30
refactor(web): use the shared formatBytes in chat attachments

jcfs:
`apps/web/lib/utils/format-bytes.ts` is the canonical byte formatter and its doc comment asks new code to import it, yet two chat-attachment modules each shipped a byte-identical local copy. Both copies are deleted and every call site now uses the shared helper.

## Important Changes

- **Deleted both local copies** (`components/task/chat/image-attachment-preview.tsx`, `components/task/chat/file-attachment.ts`). Both were exported, so six importers were repointed at `@/lib/utils/format-bytes`.
- **Size labels gain a decimal.** The copies used `parseFloat(x.toFixed(1))`, which strips a trailing zero; the canonical helper keeps it. Over-limit toasts and attachment chips now read `10.0 MB` instead of `10 MB`. Unit and e2e expectations updated to match.
- **Bad input now renders `-` instead of garbage.** The copies had no upper clamp (any value >= 1 TB indexed past the end of the unit array and rendered `1 undefined`) and no `NaN`/negative guard (`NaN` rendered `NaN undefined`). The canonical helper returns `-` for null/undefined/non-finite and `0 B` for `<= 0`.

On severity: attachment sizes are capped at 10 MB per file, so the missing TB clamp is theoretical, not a real user-facing bug. The plausible bad input is a missing or malformed `size` reaching a preview chip, which previously rendered `NaN undefined`. The honest justification for this PR is the duplication plus the canonical helper asking to be used.

Both call sites render `-` sensibly: the preview chip shows `-` in place of the size, and the context chip shows `filename.txt (-)`.

Scope is `formatBytes` only. `formatRelativeTime` (#2069) and `formatDuration` (#2068) are separate PRs. One further local copy remains in `app/office/agents/[id]/components/instruction-file-list.tsx` — unexported, outside this task's call sites, left for a follow-up.

## Validation

- `pnpm --filter @kandev/web lint` — passed
- `pnpm run typecheck` (apps/web) — passed
- `pnpm vitest run` (apps/web) — 964 files, 7370 passed, 4 skipped, 0 failed
- `lib/utils/format-bytes.test.ts` extended with the above-TB clamp (`1024**5` -> `1024.0 TB`) and non-finite input (`NaN`, `+/-Infinity` -> `-`); null, negative, `0`, and KB/MB/GB/TB cases were already covered
- E2E specs `create-task-attachment-warning.spec.ts` and `mobile-attachment-size-warning.spec.ts` had expected strings updated for the decimal change; verified by the CI E2E shards on this head rather than locally

## Possible Improvements

Low risk. The one user-visible consequence is the extra decimal in size labels; if that reads worse than the old `10 MB`, the fix belongs in the canonical helper so every caller changes together, not in a re-added local copy.

## Checklist

### DIFF
--- a/apps/web/components/task-create-dialog-selectors.test.tsx
+++ b/apps/web/components/task-create-dialog-selectors.test.tsx
@@ -4,6 +4,7 @@ import { afterEach, describe, expect, it, vi } from "vitest";
 import { TooltipProvider } from "@kandev/ui/tooltip";
 import { ToastProvider } from "@/components/toast-provider";
 import { MAX_FILES, MAX_TOTAL_SIZE, processFile } from "@/components/task/chat/file-attachment";
+import { formatBytes } from "@/lib/utils/format-bytes";
 import { TaskFormInputs } from "./task-create-dialog-selectors";
 import type { TaskFormInputsHandle } from "./task-create-dialog-types";
 
@@ -292,7 +293,7 @@ describe("TaskFormInputs attachment feedback", () => {
     const warning = await screen.findByTestId(TOAST_MESSAGE_TEST_ID);
     expect(warning.textContent).toContain("Attachment limit reached");
     expect(warning.textContent).toContain(
-      `Attachments can total up to ${MAX_TOTAL_SIZE / 1024 / 1024} MB.`,
+      `Attachments can total up to ${formatBytes(MAX_TOTAL_SIZE)}.`,
     );
   });
 
@@ -312,7 +313,7 @@ describe("TaskFormInputs attachment feedback", () => {
     const warning = await screen.findByTestId(TOAST_MESSAGE_TEST_ID);
     expect(warning.textContent).toContain("Attachment is too large");
     expect(warning.textContent).toContain(
-      "copied-image.png is 14 MB. The maximum file size is 10 MB.",
+      "copied-image.png is 14.0 MB. The maximum file size is 10.0 MB.",
     );
   });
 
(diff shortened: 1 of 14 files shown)


jcfs:
@coderabbitai review

jcfs:
Thanks — acted on one of the two pre-merge warnings.

**Description check — fixed.** The body now follows `.github/pull_request_template.md`: prose summary with no heading, `Important Changes`, `Validation`, `Possible Improvements`, and the Checklist kept as-is and unfilled. The auto-generated attribution footer is removed; the template forbids it explicitly, so that rule wins over my default footer.

**Docstring Coverage — not acting on this one.** The threshold is measured against a diff that adds no functions to document. It deletes two `formatBytes` copies, adds import statements at seven call sites, and adds test cases. The one function that survives, `lib/utils/format-bytes.ts`, already carries a doc comment explaining its units, its `-` and `0 B` return values, and that it is the shared helper. Adding docstrings to `it(...)` blocks to move the percentage would be noise. Happy to reconsider if there's a specific undocumented symbol I've missed.

jcfs:

@codex review

Context for reviewers: this deletes two byte-identical local `formatBytes` copies from the chat-attachment modules and repoints seven call sites at the canonical `apps/web/lib/utils/format-bytes.ts`. The change worth scrutinising is behavioural, not structural — the deleted copies used `parseFloat(x.toFixed(1))` (strips a trailing zero) while the canonical helper uses `.toFixed(1)` (keeps it), so user-visible size labels move from `10 MB` to `10.0 MB`. The canonical helper also returns `-` for null/undefined/non-finite where the copies returned `NaN undefined`. Please check the `-` path renders acceptably in the attachment preview chip and the context chip.

PR #2076, opened 2026-07-30
fix(web): keep transcript scroll position across nav when auto-scroll is on

ClemDNL:
## What

Fixes the transcript scroll position silently resetting to the bottom
when navigating away from a task and back, while the auto-scroll toggle
is left at its default (enabled) state.

## Root cause

Both message-list renderers (`native`, `virtuoso`) already capture the
scroll offset continuously, independent of the auto-scroll toggle:

- Native: `useAutoScroll`'s `captureScrollTop` writes
  `scrollTopBySessionId[sessionId]` on every scroll event.
- Virtuoso: `captureSnapshot` writes `virtuosoStateBySessionId[sessionId]`
  on unmount.

But both only *restored* that captured offset at mount time when
auto-scroll was disabled:

- `resolveNativeInitialScrollTop` returned `scrollHeight` (bottom)
  unconditionally when `enabled: true`.
- `message-list-virtuoso-auto-scroll.ts`'s `restoreStateFrom` lazy
  initializer skipped the snapshot when `enabled: true`.

Since auto-scroll is enabled by default, any navigate-away-and-back
while scrolled up snapped back to the bottom. Reproduced live via a
throwaway e2e spec: scrolled to `scrollTop=989` (`scrollHeight=1979`),
navigated to the kanban board and back, landed at `scrollTop=1939`
(bottom).

## Fix

Drop the `enabled` gate on the mount-time restore in both renderers.
Always prefer the captured offset; fall back to the bottom only when
nothing was ever captured, or a dockview layout-restore is pending
(unchanged precedence). No capture-side changes. Live in-session
auto-follow/freeze behavior is untouched — this only changes what a
*remount* opens with.

## Docs

- New repair spec: `docs/specs/keep-transcript-position/spec.md`
- Plan + tasks: `docs/plans/keep-transcript-position/`

## Tests

- `transcript-auto-scroll.test.ts`: updated `resolveNativeInitialScrollTop`
  unit cases for the enabled+saved-offset path.
- `auto-scroll-toggle.spec.ts`: two new e2e regressions — native and
  virtuoso renderers, auto-scroll left enabled, scroll mid-transcript,
  navigate away/back, assert restored (not bottom) position.
- All pre-existing disabled-state tests in the same spec pass unmodified.

Verified: unit tests, `pnpm run typecheck`, `pnpm run lint`, and the full
`auto-scroll-toggle.spec.ts` (9/9) all green.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2076?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ClemDNL:
Closing: the original diagnosis was wrong. The requirement is that the transcript scroll position is preserved across navigating away and back **only while auto-scroll is disabled** — which is already the shipped, correct behavior on main. My fix mistakenly made position-preservation apply when auto-scroll is enabled too, which is not wanted (enabled means the transcript should snap to the bottom to catch up with new output). Reverted; this branch is now byte-identical to main. No merge needed.

PR #2113, opened 2026-07-31
fix(agentctl): demote unwatched workspace trackers to slow polling

JnManso:
Workspace trackers are constructed in `PollModeFast` on the assumption that the gateway will push a slower mode once it knows whether anyone is watching, but for a workspace nobody focuses or subscribes to that push never arrives — so an unwatched tracker scanned at the 2s/3s cadence for the life of the process, costing ~18.6 `git.exe` spawns/second and ~50% of a core on Windows. A 60s grace timer now demotes any tracker the gateway never speaks about, while trackers created after launch inherit the mode it already pushed.

## Important Changes

- `newWorkspaceTracker` keeps the fast default, and `Start` arms a 60s fallback that demotes to slow when no explicit `SetPollMode` arrives. Fixed at the tracker rather than on the push path because `manager_rescan.go` and `GetWorkspaceTrackerFor` create trackers without assigning any mode, so a push-side fix would miss them.
- Any push disarms the fallback permanently, including a push matching the current mode: it changes no cadence but proves the gateway is managing the workspace, which is what the demotion exists to detect the absence of.
- `Manager` remembers the last pushed mode and applies it to trackers created after launch, so a repository added by rescan to an already-focused workspace is not demoted while the user is looking at it.

## Validation

- `go build ./...`, `go vet ./internal/agentctl/server/process/` — clean.
- `go test ./internal/agentctl/server/process/ -run 'TestPollModeGrace|TestRescanTracker|TestManager_SetWorkspacePollMode' -race` — 7 tests pass, covering demotion, explicit push, no-op push, unstarted tracker, inheritance into a focused workspace, and grace still owning a tracker when the gateway never pushed.
- Verified on a live Windows instance: unwatched workspaces log `no poll mode pushed within grace period, demoting to slow` 60s after startup while focused ones stay fast. With no client attached the backend now goes quiet — 0.1 `git.exe`/s and ~0% agentctl CPU, against 15+/s and ~50% of a core before.

## Possible Improvements

Low risk; the blast radius is the polling cadence of workspaces nobody is watching. Two larger issues found while investigating are not addressed here: agentctl instances are created per execution and never deleted in the standalone path, so trackers accumulate; and a focused task pays the fast cadence once per repository, which on a 12-repo task is 12 trackers at ~2.33 git commands/second each. Process spawn on Windows costs ~134ms against 3-5ms on macOS, so the same cadence is roughly 30x more expensive there.

## Checklist

### DIFF
--- a/apps/backend/internal/agentctl/server/process/manager.go
+++ b/apps/backend/internal/agentctl/server/process/manager.go
@@ -126,13 +126,14 @@ type Manager struct {
 	// per-field repoTrackersMu still allows concurrent subscribe/unsubscribe
 	// readers while a rescan is in flight.
 	rescanMu sync.Mutex
-	// lastPollMode is the most recent mode the gateway pushed for this
-	// workspace, empty until it pushes at all. Trackers created after that
-	// point — rescan adding a repository to a workspace that is already
-	// focused, or a lazily-created subpath tracker — inherit it, because no
-	// further push arrives unless the session's own mode changes.
-	lastPollMode   PollMode
-	lastPollModeMu sync.RWMutex
+	// workspacePollMode is the most recent mode the gateway pushed for this
+	// workspace. Trackers created after that point — rescan, rebind or
+	// reconcile adding a repository to a workspace that is already focused —
+	// inherit it, because no further push arrives unless the session's own
+	// mode changes.
+	workspacePollMode    PollMode
+	workspacePollModeSet bool
+	workspacePollModeMu  sync.RWMutex
 	// workspaceTrackersBySubpath caches per-subpath trackers for multi-repo
 	// task roots. Key is the cleaned subpath (relative to cfg.WorkDir). The
 	// root tracker lives in workspaceTracker above.
@@ -614,7 +615,6 @@ func (m *Manager) GetWorkspaceTrackerFor(subpath string) (*WorkspaceTracker, err
 		return t, nil
 	}
 	t := NewWorkspaceTracker(full, m.logger)
-	m.inheritPollMode(t)
 	t.SetBaseBranch(lookupBaseBranch(m.getBaseBranches(), cleaned))
 	m.workspaceTrackersBySubpath[cleaned] = t
 	return t, nil
@@ -745,20 +745,20 @@ func (m *Manager) RepoSubpaths() []string {
 // 60s later, while the user is looking at it.
 func (m *Manager) newTrackerForRepo(path, repositoryName string) *WorkspaceTracker {
 	t := NewWorkspaceTrackerForRepo(path, repositoryName, m.logger)
-	m.inheritPollMode(t)
+	m.configurePollMode(t)
 	return t
 }
 
-// inheritPollMode applies the last mode the gateway pushed, if it ever pushed
-// one. SetPollMode also disarms the tracker's grace demotion, which is correct:
-// the gateway has spoken for this workspace, so the fallback must not override
-// it. With no prior push the tracker keeps its own default and the grace timer
-// stays in charge.
-func (m *Manager) inheritPollMode(t *WorkspaceTracker) {
-	m.lastPollModeMu.RLock()
-	mode := m.lastPollMode
-	m.lastPollModeMu.RUnlock()
-	if mode == "" {
+// configurePollMode applies the last mode the gateway pushed, if it ever
+// pushed one. SetPollMode also disarms the tracker”s grace demotion, which is
+// correct: the gateway has spoken for this workspace, so the fallback must not
+// override it. With no prior push the tracker keeps its own default and the
+// grace timer stays in charge.
+func (m *Manager) configurePollMode(t *WorkspaceTracker) {
+	m.workspacePollModeMu.RLock()
+	mode, ok := m.workspacePollMode, m.workspacePollModeSet
+	m.workspacePollModeMu.RUnlock()
+	if !ok {
 		return
 	}
 	t.SetPollMode(mode)
@@ -772,9 +772,10 @@ func (m *Manager) inheritPollMode(t *WorkspaceTracker) {
 // after a focus event, since the agent's initial pushes happen at boot and
 // no replay path exists for clients that subscribe later.
 func (m *Manager) SetWorkspacePollMode(ctx context.Context, mode PollMode) {
-	m.lastPollModeMu.Lock()
-	m.lastPollMode = mode
-	m.lastPollModeMu.Unlock()
+	m.workspacePollModeMu.Lock()
+	m.workspacePollMode = mode
+	m.workspacePollModeSet = true
+	m.workspacePollModeMu.Unlock()
 
 	root, trackers := m.snapshotTrackers()
 	root.SetPollMode(mode)
--- a/apps/backend/internal/agentctl/server/process/manager.go
+++ b/apps/backend/internal/agentctl/server/process/manager.go
@@ -785,6 +785,18 @@ func (m *Manager) applyWorkspacePollModeLocked(trackers ...*WorkspaceTracker) {
 // after a focus event, since the agent's initial pushes happen at boot and
 // no replay path exists for clients that subscribe later.
 func (m *Manager) SetWorkspacePollMode(ctx context.Context, mode PollMode) {
+	// Reject before storing, not just before applying. Every tracker already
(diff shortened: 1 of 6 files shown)


carlosflorencio:
The grace-period idea looks reasonable, but I think there’s one blocker around tracker replacement.

`SetWorkspacePollMode` only updates the trackers that exist at that moment. If a focused workspace later creates replacement/new trackers through rescan, rebind, or reconciliation, the gateway doesn’t emit another mode transition. Those trackers therefore receive no explicit mode and demote to slow after 60 seconds, even though the workspace is still focused.

I’d suggest retaining the last explicit mode on `Manager` and applying it before starting every newly created tracker. Roughly:

```go
type Manager struct {
    // Protected by an appropriate manager-level mutex.
    workspacePollMode    PollMode
    workspacePollModeSet bool
}

func (m *Manager) configurePollMode(tracker *WorkspaceTracker) {
    if m.workspacePollModeSet {
        tracker.SetPollMode(m.workspacePollMode)
    }
}
```

Then all rescan/rebind/reconcile creation paths would call `configurePollMode(tracker)` before `tracker.Start(ctx)`. A manager-level test covering `fast → rescan/rebind → grace expires` would pin this behavior.

Two smaller improvements:

- Please replace the fixed `time.Sleep(300 * time.Millisecond)` assertions with `testing/synctest` or deterministic timer-state synchronization, and add coverage for `Stop()` during the grace window.
- The constructor comment says trackers from `GetWorkspaceTrackerFor` previously scanned forever, but those trackers are explicitly never started, so that part of the rationale should be removed.

Thanks again!

JnManso:
Thanks — good catch, and you were right that tracker replacement was the blocker. I'd pushed a fix just after your comment (`c942b5797`); it's essentially what you sketched, and I've now adopted your naming and the explicit set flag — `workspacePollMode` / `workspacePollModeSet` applied through `configurePollMode` before every post-launch tracker starts. `RescanRepositories`, `RebindWorkspace` and `ReconcileRepositories` all create trackers in `manager_rescan.go`, so all six sites are covered, and there's a manager-level test driving the real `RescanRepositories` path for focused → rescan → new tracker, plus its converse where the gateway never pushed and the fallback must stay in charge.

You were right about `GetWorkspaceTrackerFor` too — those trackers are cached in `workspaceTrackersBySubpath` and never started, so they run no polling loops. Removed that claim from the constructor comment and dropped the `configurePollMode` call I'd put there, which was a no-op.

On the sleeps: `testing/synctest` doesn't fit here — these trackers run real git subprocesses, so their goroutines never become durably blocked and the bubble can't advance. I took the deterministic timer-state option you also offered: once a push is recorded and the timer cleared, no elapsed time can demote the tracker, so those cases assert on `pollModePushed` + `pollModeGraceTimer` and wait for nothing. The package's poll-mode tests went from 137s to 61s. Added `Stop()`-during-grace coverage as well.

Rebased on main. Measured on Windows after: one focused task with 11 repositories sits at **27.1 `git.exe`/s** — exactly the 12 trackers × 2.33/s the design implies, so it's behaving correctly, just expensively. Notably this is *worse* than before the inheritance fix, because trackers that were previously being demoted by mistake are now correctly kept fast. This PR only makes unwatched workspaces cheap; the cost of a watched one is unchanged.

Which makes your batching/coalescing point the whole ballgame, and I found two concrete things while measuring:

- `git status --porcelain=v2 --branch` returns `branch.oid` + `branch.head` + the file entries in one invocation, replacing `rev-parse HEAD`, `rev-parse --abbrev-ref HEAD`, `symbolic-ref` and `status --porcelain`. That alone takes `pollGitChanges` from 4 spawns to 1.
- `git log` and `cumulative-diff` have no dedup, and the multi-repo `git log` fan-out is sequential (`handleGitLogMultiRepo`, one repo at a time). I logged `duration_ms 60081` returning 3958 bytes, with the client timing out and retrying — so a second full 11-repo fan-out starts while the first is still running. `getGitStatus` already uses `singleflight.DoChan` for exactly this shape. One wrinkle: the handlers use the request context, so coalescing needs the shared work decoupled from whichever caller started it, otherwise one client giving up kills the work for everyone that joined.

Worth knowing why this is invisible on macOS: process spawn on Windows is ~134ms against 3-5ms there — Git for Windows spawns a shim plus the real binary, `CreateProcess` is far costlier than `fork/exec`, and Defender inspects each one. Twelve trackers at 2.33 commands/second is ~3.7 cores of pure spawn overhead before any useful work, versus ~11% of a core on macOS. Same code, same number of tasks.

Happy to take the coalescing as a follow-up PR if you want it.

JnManso:
Both addressed in 57df4dc.

**Non-determinism** — you were right, and the failure mode was worse than flakiness: these tests would have kept *passing* while testing the wrong thing. With the 50ms grace applied to every tracker, a stall between `Start` and `SetPollMode` demotes first, so `TestPollModeGrace_NoOpPushStillDisarms` exercises a slow-to-fast transition and its assertion still holds. It would never have caught the regression it exists to catch.

The grace duration is now a parameter: only the demotion test uses a short one, the rest use an hour, so a push is the only thing that can change the mode. `TestPollModeGrace_ExplicitPushDisarmsDemotion` now pushes `PollModeSlow` — as you noted, neither test covered the mode-changing path — and since the fallback cannot fire, the slow mode it asserts can only have come from the push. The no-op test keeps its fast push and now also asserts the mode was left alone.

**Goroutine rationale** — correct, the comment was wrong. `time.AfterFunc` registers a runtime timer and only spawns a goroutine when it fires, so an armed-but-pending timer holds no goroutine. Reworded to what that test actually guards: an unstarted tracker allocating a fallback timer that nothing would disarm, since only `Stop` clears it.

JnManso:
**@coderabbitai — the registration race (cb5452c).** Confirmed and fixed. You were right that this reintroduced the very failure the PR exists to prevent: a tracker adopting a stale mode, with `SetPollMode` marking it pushed and disarming the grace timer, so nothing could ever correct it — wrong cadence for the life of the process.

The fix follows your suggestion of one synchronization boundary rather than two. `workspacePollMode` now lives under `repoTrackersMu`, the lock already guarding the tracker set, and `workspacePollModeMu` is gone. The rescan paths apply the mode inside the same critical section that publishes the trackers (four publication points), and `SetWorkspacePollMode` holds that lock across both the store and the fan-out. A tracker is therefore either registered before a push and receives it, or registered after and reads the stored mode on the way in — no window in between.

Holding the lock across the fan-out is cheap: `SetPollMode` is a mutex update plus a non-blocking wake-up. `RefreshGitStatus` is the slow part and stays outside it, in the goroutine that was already there.

**@chatgpt-codex-connector — focus mode for rescan-created trackers.** This one was already addressed in 561cb39, before your comment's diff was superseded: `Manager` records the last pushed mode and trackers created afterwards inherit it, which is what stops a rescan-added repository from being demoted while the user is looking at it. The race above was a defect in *that* fix rather than in the original grace timer.

Verified with `go test -race` over the poll-mode, rescan and tracker suites: no data races. One unrelated pre-existing failure on this machine, `TestManager_StartAllWorkspaceTrackers_StartsRootAndRepoTrackers`, which fails identically on the base commit.

carlosflorencio:
The current head addresses the tracker-registration/focus-mode races from the earlier reviews. One small case remains: `SetWorkspacePollMode` stores an invalid mode even though each tracker ignores it. That can replace the last valid inherited mode for trackers created later.

Could we validate before storing it?

```go
func (m *Manager) SetWorkspacePollMode(ctx context.Context, mode PollMode) {
    if !mode.IsValid() {
        return
    }

    m.repoTrackersMu.Lock()
    // ...
}
```

A focused test showing that an invalid update does not replace the last valid inherited mode would cover it.

The other unresolved threads appear addressed by the current locking and inheritance changes, so they can be resolved. Thanks for working through those races.

JnManso:
Good catch — validating only at the tracker left the Manager storing garbage that later trackers would inherit, which is the same class of bug as the one this PR set out to fix: a workspace whose stored mode no longer reflects anything the gateway said.

Fixed in 02d164bc. The test drives the real rescan path — push slow, push garbage, add a repository, assert the new tracker still inherits slow. Without the guard it reports `stored mode = turbo` and the tracker falls back to its construction default.

Thanks for staying on these; the locking and inheritance threads look settled to me too.

PR #2133, opened 2026-08-01
perf(agentctl): collapse the git poll tick into a single spawn

JnManso:
On Windows the workspace tracker's git poll tick costs far more than the work it does, because each of its four git invocations pays a fixed ~146ms spawn penalty; collapsing them into a single `git status --porcelain=v2 --branch` takes the tick from a 582ms median to 176ms.

## Important Changes

- The tick's `rev-parse HEAD`, `symbolic-ref` and `status --porcelain` reads become one `status --porcelain=v2 --branch` — the `branch.oid` and `branch.head` headers carry the SHA and branch name alongside the entry lines.
- The `rev-parse --git-dir` health probe moves off the happy path. On a successful read its exit code was already implied; it now runs only when the read fails, where it still decides whether git is broken or the failure was transient. Only "broken" counts against `maxConsecutiveGitFailures`, as before.
- `checkGitChanges` is split into a router plus one handler per case. **This is not gratuitous, and it is worth a word of explanation**: the function's cyclomatic complexity was already 21 against a limit of 15 on `main`. `cyclop` reports at the function declaration line, and this repo lints with `new: true`, so adding the snapshot parameter to that line made a pre-existing violation newly visible and blocked the commit. The alternative was a `//nolint`; the split seemed the better answer since it also removes the shared mutable locals that made the original hard to follow. It is mechanical — same branches, same order, each case owning its own cache update — and can be reviewed independently of the performance change above.

## Validation

Measured on a Windows dev machine, n=25, medians, same repository:

| | |
|---|---|
| before, 4 spawns | 582ms |
| after, 1 spawn | 176ms |
| `git --version` (spawn floor, zero repo I/O) | 146ms |

So of the new tick only ~30ms is real git work — the old one was almost entirely spawn overhead. On macOS and Linux a spawn is 3-5ms and none of this is visible, which is why it went unnoticed.

- `go build ./...` and `go vet ./internal/agentctl/server/process/` — clean.
- `golangci-lint run ./internal/agentctl/server/process/` — 0 issues.
- `go test ./internal/agentctl/server/process/` — passes. One unrelated pre-existing failure on this machine, `TestWorkspaceFileMutationsRejectDescendantSymlinkSwap`, which needs the Windows symlink privilege (Developer Mode); confirmed identical with these changes stashed.
- New tests cover the parsing directly: the `(initial)` and `(detached)` sentinels, that `branch.ab` moving does not perturb the index hash, and that entry lines do. The two existing `getGitStatusHash` tests now exercise the production read path instead of a helper.
- `TestWorkspaceTracker_StopsWhenGitBroken` still passes, confirming the relocated probe stops polling after the same 5 failures.

## Possible Improvements

Low risk — no exported API and no persisted state; the index hash is only ever compared against itself within a process lifetime, so changing the status format cannot affect anything durable. Two behaviours are preserved explicitly rather than incidentally, and are the first places to look if something is off: the `(initial)` and `(detached)` sentinels map to the empty string (matching what `rev-parse` and `symbolic-ref` returned when they failed in those states), and the `# ` headers stay out of the index hash (`branch.oid` and `branch.ab` move on every commit and fetch, and would otherwise report index changes the HEAD and branch comparisons already cover).

Separately, #2113 addresses the same CPU symptom from the other direction — it demotes unwatched trackers out of fast polling, reducing how many ticks happen, where this reduces what each tick costs. The two are independent and complementary: no file overlap, and neither depends on the other. The multiplier neither of them touches is the per-repository one — a focused task with 11 repositories runs 12 trackers, and this change takes that from roughly 28 to 16 git spawns per second.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2133?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/agentctl/server/process/workspace_git_poll.go
+++ b/apps/backend/internal/agentctl/server/process/workspace_git_poll.go
@@ -36,18 +36,21 @@ func (wt *WorkspaceTracker) pollGitChanges(ctx context.Context) {
 	}
 	defer timer.Stop()
 
-	// Initialize cached HEAD SHA, branch name, and index hash
+	// Prime cached HEAD SHA, branch name, and index hash. A read failure here
+	// is non-fatal: the caches stay empty and the first successful tick fills
+	// them, which is what the previous per-value getters did on error too.
+	snap, _ := wt.readGitPollSnapshot(ctx)
 	wt.gitStateMu.Lock()
-	wt.cachedHeadSHA = wt.getHeadSHA(ctx)
-	wt.cachedBranchName = wt.getCurrentBranchName(ctx)
-	wt.cachedIndexHash = wt.getGitStatusHash(ctx)
+	wt.cachedHeadSHA = snap.headSHA
+	wt.cachedBranchName = snap.branch
+	wt.cachedIndexHash = snap.indexHash
 	wt.gitStateMu.Unlock()
 
 	wt.logger.Info("git polling started",
 		zap.Duration("fast_interval", wt.gitPollInterval),
 		zap.String("initial_mode", string(wt.GetPollMode())),
-		zap.String("initial_head", wt.cachedHeadSHA),
-		zap.String("initial_branch", wt.cachedBranchName))
+		zap.String("initial_head", snap.headSHA),
+		zap.String("initial_branch", snap.branch))
 
 	var consecutiveFailures int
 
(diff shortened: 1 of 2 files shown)


JnManso:
Good catch, and it was worse than a missed optimisation — it was a regression.

I measured it on a branch with real divergence (1799 commits behind its upstream), since with `+0 -0` there is nothing to walk and the difference hides in the spawn noise:

| | median, n=25 |
|---|---|
| `--branch` | 646ms |
| `--branch --no-ahead-behind` | 334ms |

So without the flag this PR would have been *slower* than the four-spawn version it replaces (582ms) for anyone working off a stale branch — the exact opposite of the point.

Fixed in 1d206c2. git prints `# branch.ab +? -?` when it skips the count; the parser already ignored it as a header, and there is now a test pinning that so removing the flag has to be a deliberate act rather than a silent revert to the graph walk.

PR #2138, opened 2026-08-01
perf(agentctl): parallelise the multi-repo git fan-outs

JnManso:
`handleGitLogMultiRepo` and `handleGitCumulativeDiffMultiRepo` walked their repositories serially, making request latency the sum over repositories instead of the slowest one — `handleGitStatusMulti` in the same file already fans out in parallel for exactly this reason, and these two never followed.

## Important Changes

- Both fan-outs now run one goroutine per repository, writing results by index so merge order stays deterministic. Merging itself stays sequential and in subpath order, since it writes into one shared map.
- `runGitCumulativeDiffForRepo` no longer writes its own HTTP response. This was the actual blocker: it emitted `c.JSON(...)` and returned `nil` to mean "already handled", which cannot survive a fan-out — gin's `ResponseWriter` is not safe for concurrent use, and several failing repos would each try to write. It returns `(result, status, error)` and the caller writes, so the single-repo path emits exactly the status codes it did before (400 on operator lookup, 500 on diff failure).
- `resolvePerRepoBase` and `runGitLogForRepo` took a `*gin.Context` only to reach `Request.Context()`. Both now take a plain `context.Context`, which is what makes them callable off the request goroutine.

**Behaviour is preserved, not improved.** A hard per-repo failure still aborts the whole request, as it did serially. The fan-out reports the first failure in *subpath order* rather than the first to occur, so the response stays deterministic now that the repositories race.

## Validation

- `go build ./...` and `go vet ./internal/agentctl/server/api/` — clean.
- `golangci-lint run ./internal/agentctl/server/api/` — 0 issues.
- `go test -race ./internal/agentctl/server/api/` — no race detector warnings. `TestMultiRepoReviewEndpointsUseStoredBaseBranches` and `TestMultiRepoReviewEndpointsCorrectStaleBases`, which exercise both changed fan-out paths, pass.
- Pre-existing unrelated failures on this machine: the `TestWorkspaceRemoveMaterializedRepository*` and `TestMaterializeRepository_RejectsSymlinkedGitMetadata` family, which need the Windows symlink privilege (Developer Mode). Confirmed identical with these changes stashed — the baseline fails the same set.

**No latency measurement.** I did not have a multi-repository task to measure against, and would rather state the reasoning than invent a number: the change takes wall-clock from the sum of per-repository times to roughly the slowest one, bounded by the shared git throttle (cap 12 in `common/subproc`). The motivating case is Windows, where a git spawn costs ~150ms before doing any work, so an 11-repository task multiplies that on every commits-panel or diff refresh — against the gateway's 60s `singleflightComputeTimeout`, past which the client retries and the user sees a timeout rather than a slow answer.

## Possible Improvements

Medium risk — this is a concurrency change in request handlers, and the part worth reviewing closely is the error protocol rather than the goroutines. One judgement call made deliberately: on a hard per-repo failure I kept "abort the whole request", because changing it belongs in a behaviour PR rather than a performance one. The other multi-repo handlers instead return partial results and let the frontend render an incomplete warning (`handleGitStatusMulti` packs failures into each repo's `Status.Error`; the log fan-out uses `PerRepoErrors`), which is arguably the better contract — worth aligning separately if you agree.

Context: this is the third of three changes addressing high CPU and slow git operations, and the only one targeting request latency rather than background polling. #2113 demotes unwatched workspace trackers out of fast polling; #2133 collapses the poll tick from four git spawns to one. All three are independent — no file overlap, and none depends on the others.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2138?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


carlosflorencio:
I found one cancellation regression to fix before merging.

The single-repo log path now passes Gin’s context directly:

```go
result, err := s.runGitLogForRepo(c, req, limit, req.Repo)
```

Because this server uses `gin.New()` without `ContextWithFallback`, cancellation from the HTTP request won’t propagate through `c`. This should use the request context, like the multi-repo path already does:

```go
result, err := s.runGitLogForRepo(c.Request.Context(), req, limit, req.Repo)
```

Could we also add a small regression test showing that request cancellation reaches the Git operation? A deterministic two-repo gate test proving the new fan-out runs concurrently would also give this performance behavior direct coverage.

The process-wide Git throttle already limits active subprocesses, so bounding the goroutine fan-out can be a follow-up unless we expect unusually large repository lists.

JnManso:
Good catch, and it was a regression I introduced rather than a pre-existing gap — worth spelling out because of how quietly it happened.

Before this PR, `runGitLogForRepo` took a `*gin.Context` and reached for `c.Request.Context()` itself. Changing the parameter to `context.Context` should have broken the single-repo call site, but `*gin.Context` satisfies `context.Context`, so `runGitLogForRepo(c, ...)` kept compiling — now handing over a context whose `Done()` is nil, since this server is built with `gin.New()` and does not enable `ContextWithFallback`. The merge-base, rev-parse and log subprocesses simply stopped being cancelled on client disconnect, with nothing in the type system to say so.

Fixed in 44b2157, with a comment at the call site recording why it must be `c.Request.Context()` and not `c`. I audited the other call sites in the file at the same time: `resolvePerRepoBase` and `runGitCumulativeDiffForRepo` were already correct on both paths.

JnManso:
Done — `t.Cleanup(cleanup)` in place of `defer cleanup()`, so it also runs for subtests and for a `t.Fatal` raised inside a helper.

JnManso:
Thanks — you and the Codex reviewer found the same cancellation regression independently, which says something about how quietly it happened. Fixed in 44b2157, with a note at the call site recording *why* it must be `c.Request.Context()`: `*gin.Context` satisfies `context.Context`, so changing the parameter type left the single-repo call site compiling while silently handing over a context whose `Done()` is nil. I audited the other call sites in the file at the same time — `resolvePerRepoBase` and `runGitCumulativeDiffForRepo` were already correct on both paths.

Both tests are in 40b05f2.

**Cancellation reaches the git operation.** One test per path. Each issues the same request twice — once live, to prove the fixture really produces commits, then with an already-cancelled context, which must not. Without the live half a broken fixture would look like a passing test. Verified they fail with the bug restored: *"returned 1 commits for a cancelled request; cancellation is not reaching the git operation"*.

**Deterministic fan-out gate.** I extracted the two identical fan-out loops into `fanOutRepos` first, so the concurrency has one place to be tested rather than two copies to be trusted. The test is a gate, not a timing measurement: every collect call blocks until all of them have arrived, so serial execution cannot pass — the first would wait forever and the rest would never start. The timeout is a failure detector rather than a threshold, so machine speed doesn't enter into it. Verified it fails with the helper made serial: *"not every repository entered the fan-out; the work is running serially"*.

One note on placement: `git_status_fresh_concurrency_test.go` is `//go:build !windows` because its shim needs `mkfifo`, so these use plain channels and run everywhere. The trade-off is that the gate covers the helper rather than the full HTTP stack.

`handleGitStatusMulti` still has its own copy of the loop — folding it into the helper felt like a separate change rather than something to slip into this one.

Agreed on not bounding the goroutine fan-out here. Related: the throttle turns out to be the binding constraint on tasks with many repositories, and it is shared with background polling — I opened #2150 with the measurements, since the fix looked like a design call on your component rather than something to patch unilaterally.

carlosflorencio:
Thanks for fixing the request-context cancellation path. I think two things remain before merging.

The new per-repository `-n` bound can change which commits survive the final merge: Git truncates using traversal order, while `mergeGitLogResults` later sorts by author timestamp. With skewed author/committer dates, a discarded commit could have ranked inside the merged result. `--author-date-order` still respects topology, so it would not fully match the in-memory sort. Since the issue measurements say this optimization did not affect the timeout count, I suggest removing `boundBaseRange` from this PR and handling it separately if approximate ordering is acceptable.

The fan-out also still starts one worker per repository, allowing one request to occupy the complete shared Git budget. Could we use a bounded group or worker pool, ideally aligned with the admission design in #2150?

```go
group, ctx := errgroup.WithContext(ctx)
group.SetLimit(maxGitFanout)

for i, sub := range subpaths {
    i, sub := i, sub
    group.Go(func() error {
        outcomes[i] = s.collectLogForRepo(ctx, req, limit, sub)
        return nil
    })
}
```

Please add a channel-gated test proving the limit, cancellation, and actual parallel execution without timing sleeps.

Container E2E shard 5 also reached its 20-minute timeout, so that check needs a rerun after the update. If it times out again, we should inspect which container test is hanging. Thanks for the follow-up changes here.

JnManso:
You are right about the `-n` bound, and my equivalence argument was wrong. I reasoned as if git returned the newest N under the same ordering the merge applies; it truncates by traversal order while `mergeGitLogResults` sorts by author timestamp, so with skewed author/committer dates a dropped commit could have ranked inside the merged result. Reverted in c56b11f5 — and since the #2150 measurements showed it did not move the timeout counts, it was carrying a correctness risk for no demonstrated gain.

**Bounded fan-out** via `errgroup.SetLimit`, with the collect callback now receiving the group context so cancellation reaches every repository rather than only those not yet started. `maxGitFanout` is a judgement call rather than a derived number — `subproc` exposes no public accessor for its cap, and adding one felt like it belongs with the admission design in #2150 rather than here. Happy to align it to whatever that lands on.

**Three gate tests, no sleeps:** parallelism (workers block until all have arrived, so serial execution deadlocks and the timeout reports it), the limit, and cancellation propagation. The limit test was verified to fail without `SetLimit`: `peak concurrency = 12, want at most 8`. Also added `goleak.VerifyTestMain` for the package, per the other review comment.

I cannot rerun the container E2E shard from here — could you kick it off, or point me at how to trigger it?

carlosflorencio:
@JnManso can you address the failing ci?

JnManso:
@carlosflorencio the failing check was mine, but not from the fan-out — it was the `goleak.VerifyTestMain` I added in the previous commit at coderabbitai's request. Removed in aa93852, so this should go green.

Worth recording why rather than just reverting it. The leaked goroutines are HTTP/2 client streams reading a gzipped tar over the network, from a download path in this package that predates this PR — Windows CI passed, Linux did not, so it is platform-dependent too. `goleak` did not create the problem; it made a pre-existing one visible, inside a PR that is about something else.

@coderabbitai I did consider the escape hatch you mentioned. `IgnoreTopFunction` would need to cover `net/http.(*http2clientStream).writeRequest` and the flate decompressor, and those are broad enough to mask real leaks in any HTTP path — including the ones this PR adds. That would leave the check present and not doing its job, which seemed worse than not having it.

Enabling `goleak` for this package belongs with the change that fixes that download. Happy to do both together as a follow-up if you want it. In the meantime the three gate tests already cover what this PR introduces: each waits on the group returning, so a worker left blocked fails them rather than passing quietly.

PR #2149, opened 2026-08-01
fix(agentctl): abandon instance creation when the caller has gone

JnManso:
`CreateInstance` is serialised on a single mutex and the control client gives up after 30s, so under load the lock grows a queue of requests whose callers have already timed out — and each one still built a complete instance that nobody could ever stop, which is what turned a transient slowdown into a permanent collapse.

## Important Changes

- `CreateInstance` now checks the request context twice: on entry, which covers the queued case, and again after `StartAllWorkspaceTrackers`, which is the slow part a caller can time out during.
- `abandonPartialInstance` unwinds what was built when the second check fires — tracker goroutines, bound listener, allocated port. It has to do this by hand: `StopInstance` takes the same mutex still held here, and the instance was never registered in the map to begin with.

The context was already threaded in — `handleCreateInstance` passes `c.Request.Context()`. It was simply never consulted.

## Validation

Measured on a Windows dev machine with one task open, before the fix. `POST /api/v1/instances` latency over a single session:

```
9ms  18ms  7ms  6ms  6ms  →  5541ms  →  34329ms  36215ms  50144ms
                                        64698ms  99775ms  149126ms  192104ms
```

Monotonically increasing, which is the signature of a queue rather than of slow work: the first five completed in single-digit milliseconds. 14 creations, **zero** deletions, 16 instances still listening at the end. By that point plain `git ls-files` was exceeding its 10s timeout, terminals returned 503, and session resume failed outright — each failure prompting a retry that lengthened the queue further.

- `go build ./...` and `go vet ./internal/agentctl/server/instance/` — clean.
- `golangci-lint run ./internal/agentctl/server/instance/` — 0 issues.
- `go test -race ./internal/agentctl/server/instance/` — passes.
- Two new tests: a cancelled caller must produce an error wrapping `context.Canceled` and register no instance; and the unwind path must return the port to the allocator and close the listener. I verified both fail without the fix — with the context checks disabled, the first reports `got instance &{ID:34a23e49-... Port:0}`, which is precisely the orphan.

## Possible Improvements

Low risk in isolation — no exported signature changes, no persisted state, and the new behaviour only triggers on an already-cancelled context, where the previous code produced an instance nobody could reach.

What this deliberately does **not** address, both of which deserve their own change:

- **Why a creation became slow in the first place.** The trigger is contention on the process-wide git throttle (cap 12 in `common/subproc`), which background workspace polling shares with interactive request fan-outs. On a task with many repositories a single fan-out saturates it, and `CreateInstance` starts every repository's trackers while holding the lock. This fix stops a slowdown from compounding; it does not prevent the slowdown.
- **The serialisation itself.** One slow creation still blocks every other. Callers now fail cleanly instead of leaking, which is strictly better, but they still fail.

Related and independent: #2113 demotes unwatched workspace trackers out of fast polling, #2133 collapses the poll tick from four git spawns to one, and #2138 parallelises the multi-repo fan-outs. No file overlap between any of the four.

## Checklist

### DIFF
--- a/apps/backend/internal/agentctl/server/instance/manager.go
+++ b/apps/backend/internal/agentctl/server/instance/manager.go
@@ -26,6 +26,11 @@ type ServerFactory func(cfg *config.InstanceConfig, procMgr *process.Manager, lo
 
 const instanceHTTPShutdownGrace = 250 * time.Millisecond
 
+// ErrManagerShuttingDown is returned by CreateInstance once Shutdown has begun.
+// Callers get a clear refusal instead of an instance that the shutdown sequence
+// has already walked past and will never stop.
+var ErrManagerShuttingDown = errors.New("instance manager is shutting down")
+
 // Manager manages multiple agent instances.
 // It handles creation, tracking, and removal of agent instances,
 // each with their own HTTP server on a dedicated port.
@@ -50,7 +55,22 @@ type Manager struct {
 	// caller vanishes mid-creation. They run off the creation mutex so a slow
 	// tracker stop cannot block the queue; Shutdown drains them so a process
 	// exit does not leave half-torn-down trackers behind.
-	abandonWG sync.WaitGroup
+	//
+	// shuttingDown, guarded by mu, closes the window where Shutdown could
+	// observe the counter at zero and a CreateInstance already past its own
+	// checks could then Add(1) behind it. Setting the flag under the same mutex
+	// CreateInstance holds means every creation either completes before the
+	// flag is set — so its Add is visible to the Wait below — or sees the flag
+	// and refuses.
+	abandonWG    sync.WaitGroup
+	shuttingDown bool
+
+	// afterTrackerStart runs immediately after StartAllWorkspaceTrackers and is
+	// nil in production. It exists so a test can cancel the caller's context at
+	// exactly that point and assert the abandonment branch ran, rather than
+	// racing cancellation against tracker startup and accepting whichever
+	// outcome it happens to get.
+	afterTrackerStart func()
 }
 
 // NewManager creates a new instance manager.
@@ -87,6 +107,12 @@ func (m *Manager) SetServerFactory(factory ServerFactory) {
 func (m *Manager) CreateInstance(ctx context.Context, req *CreateRequest) (*CreateResponse, error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
+
+	// Refuse once Shutdown has begun, so no creation can register an
+	// abandonment teardown after Shutdown drained the wait group.
+	if m.shuttingDown {
+		return nil, ErrManagerShuttingDown
+	}
 
 	// The caller may already be gone. Creation is serialised on m.mu, the
 	// control client gives up after 30s, and callers that give up retry — so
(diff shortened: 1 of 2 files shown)


JnManso:
**@chatgpt-codex-connector — teardown under the create lock (6936a1f).** Confirmed, and worse than described. `stopWorkspaceTrackers` stops the root tracker and then each repository tracker in sequence, and `WorkspaceTracker.Stop` takes no context at all — it falls back to its own `stopTimeout` of 5s. So a workspace with a dozen repositories could pin the creation mutex for over a minute. The `abandonPartialInstanceTimeout` I passed into `StopForTeardown` only bounded the admission wait; it never reached the stops. My comment claiming the timeout was "deliberately short because m.mu is held" was reasoning about a bound that did not exist.

That made the fix capable of worsening exactly the queue it exists to drain. Teardown now runs on its own goroutine, tracked by a wait group so `Shutdown` drains it rather than exiting on half-torn-down trackers. It is safe off the lock because the instance was never added to `m.instances` and `PortAllocator` carries its own mutex.

**@coderabbitai — test for cancellation after startup begins.** Added, with one deliberate difference from what you suggested. Asserting that the *second* check specifically fired would mean racing the cancellation against `StartAllWorkspaceTrackers`, and a test that has to win a race to be meaningful stops being meaningful without anyone noticing — which is the failure mode you caught in the grace tests on #2113. So it asserts the invariant instead: whichever check fires, the call must error with a wrapped `context.Canceled`, return no response, register no instance, and lose no port. Both outcomes are legitimate, including the caller winning the race and getting a live (owned, non-leaked) instance.

The test waits on the abandon wait group before asserting, since teardown is now asynchronous by design. Ran 20x under `-race` without flaking.

carlosflorencio:
Moving abandoned teardown off the creation mutex addresses the earlier queue-blocking concern. There is still a shutdown ordering race to close: `Shutdown` can call `abandonWG.Wait()` while the counter is zero, then an in-flight `CreateInstance` can call `Add(1)` afterward.

I suggest adding a shutdown guard under `m.mu`, using that lock as the barrier before draining abandonment work:

```go
func (m *Manager) Shutdown(ctx context.Context) error {
    m.stopReaperOnce()

    m.mu.Lock()
    m.shuttingDown = true
    ids := snapshotInstanceIDs(m.instances)
    m.mu.Unlock()

    if err := waitGroupWithContext(ctx, &m.abandonWG); err != nil {
        return err
    }

    // Stop registered instances...
}
```

`CreateInstance` should reject new work after that flag is set:

```go
m.mu.Lock()
defer m.mu.Unlock()

if m.shuttingDown {
    return nil, ErrManagerShuttingDown
}
```

It would also be better to close the unpublished listener and release its port before the potentially slow tracker teardown.

The current mid-creation test permits either success or cancellation, so it does not guarantee the abandonment branch ran. Could we add a deterministic barrier after startup work begins, cancel there, and assert that shutdown waits, no instance is registered, and the listener/port are released?

Thanks for tackling the leak and its cleanup paths.

JnManso:
Both addressed in ffbf566d.

**Shutdown ordering.** A `shuttingDown` flag under the same mutex `CreateInstance` holds, so every creation either finishes before the flag is set — making its `Add` visible to the `Wait` — or sees the flag and returns `ErrManagerShuttingDown`. Verified the test fails without it: `error = <nil>, want ErrManagerShuttingDown`.

**Listener and port first.** Moved ahead of the tracker stops. Nothing was ever published on that listener, and holding the port through a teardown that can take seconds shrinks the pool exactly when creations are already piling up.

**Deterministic barrier.** You were right that permitting either outcome does not prove the branch ran. Reaching it deterministically needed a seam, so there is now an `afterTrackerStart` hook — nil in production, documented as test-only — which cancels the caller's context at precisely the point the second check guards. The test asserts the post-startup branch specifically, that `Shutdown` waits for the teardown, and that the port is back afterwards. Verified to fail without the check: `got &{ID:51f0bbd1… Port:0}`.

If you would rather not carry a test-only field on the struct, say so and I will find another way to reach that branch.

PR #2193, opened 2026-08-03
feat(i18n): localize Settings → Agents

jcfs:
The agents list, the per-agent setup page and the agent-profile editor were the largest un-migrated Settings surface left, and the one densest in code that reads like copy. All three now route through `t()` / `<Trans>` in a new `agents` namespace, and the closure is added to the `i18nGuardFiles` ratchet.

## Important Changes

- **The closure was re-derived from the four routes' imports, not from the string counter.** Three files the counter attributed to this route are not reachable from it and are deliberately left alone: `components/settings/model-combobox.tsx` (reached only from the task-side CLI profile editor and Utility Agents — the profile page's model picker is `components/model-config-selector.tsx`), `components/settings/agent-card.tsx` (Office), and `components/settings/inference-agent-status.tsx` (Utility Agents). Conversely `components/settings/profile-status-panels.tsx` and `components/settings/pty-terminal-dialog.tsx` *are* in the closure and were not on the list; both are migrated.
- **`install-agent-card.tsx` was the predicted blind spot and it was real.** It counted zero because its four button labels (`Queued…`, `Installing…`, `Retry`, `Install`) live in `installButtonContent()`, a non-JSX function `mode: "jsx-only"` never inspects. `use-agent-update-dialog-state.ts` genuinely carries no copy — its `errorMessage` falls back to `String(error)`.
- **Six wire enums now carry `labelKey`s resolved at render**, never a translated value: the agent-update job `status` (`UPDATE_PHASE_KEYS`), the install-job `status`, the capability-probe `status` (`CAPABILITY_STATUS_KEYS`), the watcher `kind` (`WATCHER_KIND_LABEL_KEYS`), and the model/mode `id`s. `capabilityStatusMessage`, `updatePhase`, `installButtonContent`, `noAuthHint` and `watcherKindLabel` all take `t` as a parameter rather than calling it at module scope.
- **`agent-profile-delete-dialog.tsx` has no type-to-confirm gate.** I checked specifically — the destructive path is a plain `AlertDialogAction`, so nothing here can be made impossible to satisfy. `formatRoutingTier` title-cases the routing `tier` wire value for display and is left untranslated; the tier-mapping row is a `<Trans>` so the tier keeps its bold styling without the sentence being split.
- **`agent-profile-page-state.ts` is new.** The added copy pushed `agent-profile-page.tsx` past the 600-line cap; the draft state, save/delete actions and store sync move out so the component file stays presentation-only. `preserveNewerProfileDraft` is re-exported so the existing test's import is unchanged.
- **`ModelConfigSelector` and `PtyTerminalDialog` are migrated even though they are shared.** Both render inside this surface, so leaving them would have left plain English on a page this PR claims is done. The guard is per-file, so the chat model selector, session tabs, Utility Agents and the chat action-message shell are unaffected.

### Deliberately left in English

Each is an identifier the user reads or types verbatim, and each is interpolated as a **value** so the pseudo-locale cannot turn it into a dead pointer:

- **Substitution tokens.** `{prompt}` in the command preview and `{{model}}` in the TUI command — both are placeholders the backend replaces. `{{model}}` is passed through `t()` as a value, which i18next does not re-interpolate; there is a test pinning that, because a regression would silently blank the placeholder.
- **Code and shell fragments.** `--my-flag`, `value`, `greywall --`, `/init`, `superclaude`, `best`, the `mcpServers` JSON key, and the whole assembled command preview and update command. `cli-flags-field.tsx` and `command-prefix-field.tsx` only translate the labels around them.
- **MCP server product names** (`Playwright MCP`, `Chrome DevTools MCP`, `Context7 MCP`, `GitHub MCP`), `Kandev MCP`, and the Kandev tool list (`list_workspaces`, `create_task`, …). The `POPULAR_SERVERS` keys are the wire values the JSON is keyed by; the four buttons now render from that record instead of being spelled out, so the key and the label cannot drift.
- **Wire values.** Agent names, model ids, mode ids (`acceptEdits`, `bypassPermissions`, `plan` and friends all arrive from the agent's ACP capability payload and are never touched), the permission `apply_method`, the routing `tier`, and every `status` union.
- **Env var names and MCP server names**, which are user-entered data.

**Backend-owned copy is also left alone**, matching the contract in `docs/i18n.md`: `PermissionSetting.label` / `.description` (`Auto-approve all permissions`, `Skip permission prompts`), `PassthroughConfig.label` / `.description` (`CLI Passthrough`, `TUI Passthrough`), the available-agent `description` strings, and the agent-supplied mode and command descriptions. These render as plain English under the pseudo-locale by design — localizing them needs a key/value split in the Go API, not a frontend change.

`common:agents` is reused for the page heading and the sidebar nav label rather than duplicated into the new namespace; `common:cancel` is reused for every Cancel button.

## Validation

```
pnpm run typecheck                  # clean
pnpm lint --max-warnings 0          # clean
pnpm run i18n:check                 # 4/4 gates pass
pnpm run i18n:ratchet               # new-code clean; allowlist +20, intact
pnpm test                           # 1060 files, 8042 passed
```

English is byte-identical everywhere a test selects by accessible name, and the whole suite passes unchanged — `Watchers (will be disabled)`, `Linear:`, `GitHub Issues:`, `Jira:`, `Tasks:`, `Cannot delete agent profile`, `Workspace tier mappings:`, `Balanced tier in Office Workspace (ws-1) for Codex (codex-acp)`, `Delete Anyway`, and the `command-preview-card` / `command-prefix-field` dirty-state specs.

Five tests were added, four of them around `<Trans>`: its `<0>` index addresses the JSX children positionally, so a prettier reflow silently reassembles the sentence into fragments without failing anything. Each block now asserts its whole reconstructed sentence and the token inside its `<code>` — the `{prompt}` hint, the `{{model}}` command help, the command-prefix help, and the slash-command dialog. The fifth pins `{{model}}` surviving interpolation in the command placeholder. `add-tui-agent-dialog.test.tsx` and `profile-capability-helpers.test.tsx` are new files that also cover the validation paths those components had no coverage for.

**Pseudo-locale acceptance** (the completeness check lint cannot do) was run against a live instance on an isolated database, walking: the agents list with four installed agents and eighteen uninstalled ones, the Add-TUI-Agent dialog, the host-terminal dialog, an agent setup page in create mode, a saved profile editor, the model picker, the mode combobox, the slash-commands dialog, the command preview, the MCP config card with its built-in tooltip, the profile delete dialog, and the runtime-update dialog. Every string this PR owns renders accented. The only plain English left on those screens is the backend-owned copy listed above, the deliberate identifiers, the settings nav groups outside this route, and the Environment Variables card.

### One gap, reported rather than folded in

`components/settings/profile-edit/env-vars-card.tsx` and `profile-env-vars-section.tsx` render on the agent profile editor but belong to the **executor** profile tree, which a sibling task owns. They are the only plain-English section left inside a card this PR migrated. Pulling them in would have pulled the executor surface into this PR; the allowlist comment records why they are absent so the executor migration picks them up.

Neither route is added to `SCREENS` in `pseudo-coverage.spec.ts`, for the same reason #2182 and #2187 recorded: the settings nav (`settings-tree.tsx`, `workspaces-group.tsx`) is not migrated yet, so allowlisting those labels would hide real misses instead of proving anything.

## Possible Improvements

Low risk — copy-only, no behavior change. `agent-profile-page.tsx` gains a module split, which is mechanical and covered by the existing tests. The residual risk is a translator shipping a locale where a status or phase label drifts from the wire value; the values are a separate field from the labels precisely so that cannot happen.

## Checklist

- [ ] If my change touches UI files (`apps/web/`), I have added or updated Playwright e2e tests in `apps/web/e2e/` and verified them with `make test-e2e`. — no e2e change: this is a copy-only migration and no route is added to `SCREENS` while the settings nav is still un-migrated (see above). Acceptance was done manually against the pseudo-locale on a live instance.
- [x] I checked whether this affects public docs in `docs/public/**` and updated them or noted why no docs change is needed. — no public-docs impact; `docs/i18n.md` already documents the migration process and this PR changes no CLI, config key, flow or terminology.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2193?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jcfs:
Thanks — three of the four are fixed in 8f350ec, the fourth is deliberately out of scope.

**Fixed**

- **`pty-terminal-dialog.tsx` `Connection error`** (@chatgpt-codex-connector) — a real miss, and exactly the class the guard cannot see: the copy is `setError`'d from the WebSocket `onerror` callback, so it is neither a JSX literal nor something `mode: "jsx-only"` inspects. Allowlisting the file while that string was still English was claiming more than was true. It now resolves through the module-level `t` as `agents:ptyConnectionError`.
- **`agent-profile-page-state.ts` missing from `i18nGuardFiles`** (@claude, @coderabbitai) — correct, and the reason given is the right one: lint cannot enforce a `.ts` entry, but the listing is what records the file as reviewed, and its save/delete toast copy does reach the user. Added, with the existing `.ts` note extended to cover it and the `pty-terminal-dialog` handler.
- **Missing branch coverage** (@coderabbitai) — added the non-`Error` rejection path in `add-tui-agent-dialog.test.tsx`, and a case in `agent-profile-delete-dialog.test.tsx` asserting all four watcher kinds resolve to their labels plus an unknown kind echoing its raw value. The routing-tier row and the three known kinds were already covered by the pre-existing specs in that file, which pass unchanged.

**Not doing**

- **`SettingsGroup`'s `Expand ${label}` / `Collapse ${label}` aria-label** (@coderabbitai) — real, but `settings-nav-primitives.tsx` is the shared chrome for *every* settings group, not just Agents. Migrating it here would pull Workspaces, Executors, System and the rest into an Agents PR and change a file the sibling i18n tasks are also touching. It belongs with the settings-nav migration, which #2182 and #2187 already recorded as the blocker for adding any of these routes to `SCREENS` in `pseudo-coverage.spec.ts`. It shows as partly-English under the pseudo-locale today for Integrations and General too, so this is not a regression this PR introduces.

Revalidated: `typecheck`, `lint --max-warnings 0`, `i18n:check` (4/4), `i18n:ratchet` (allowlist +21, intact), `pnpm test` — 1060 files, 8044 passed.

PR #2196, opened 2026-08-03
fix(web): stop unrelated task.updated events from un-nesting a subtask

ClemDNL:
When both the unread-divider and anchored last-prompt bar settings are enabled, opening a task can scroll the transcript so the last prompt is already scrolled past at the same moment the New divider is placed, opening the pinned bar directly on top of it — reserving scroll room for the bar's real height fixes the overlap.

## Important Changes

- `AnchoredLastPromptBar` now measures its pinned content's real height (a normal-flow descendant of the CSS-grid-collapsed row, so the measurement stays stable regardless of open/closed state) and reports it up through `TaskChatPanel`.
- Native renderer: the height feeds a `--anchored-bar-h` CSS var consumed via `scroll-margin-top`; `useScrollToDividerOrBottom` re-asserts the divider's scroll position once the async-measured height arrives, reusing the existing settling-window/no-user-scroll gate.
- Virtuoso renderer: the height feeds a negative `scrollToIndex` offset (Virtuoso's documented sticky-header pattern); `useScrollToDividerOnceResolved` gained the same reassertion gate (previously a strict one-shot).

## Validation

- Unit tests (TDD): `message-list-shared.test.tsx` (`anchoredBarScrollOffsetPx`), `anchored-last-prompt-bar.test.tsx` (height reporting + unmount reset), `message-list-native.test.tsx` and `message-list-virtuoso.test.tsx` (hook-boundary tests mocking `scrollIntoView`/`scrollToIndex`, covering offset reassertion, no-divider-target, settling-window expiry, and a live-update-with-unchanged-offset case).
- `pnpm --filter @kandev/web test` — full suite: 1058+ files passed (one pre-existing, unrelated flaky timeout in `storage-maintenance-settings.test.tsx`, reproduced on the pre-change tree too).
- `pnpm --filter @kandev/web lint` (`eslint --max-warnings 0`) — clean.
- `tsc -p tsconfig.json --noEmit` — clean.
- Added `apps/web/e2e/tests/chat/unread-divider.spec.ts` case asserting the divider's bounding box never overlaps the anchored bar's rendered content.
- Manually verified against a real backend + built frontend on both sides of the fix (see screenshots): seeded a task where the divider lands right after the last prompt, confirmed the bar covers the divider before the fix and clears it after.

## Checklist

### DIFF
--- a/apps/web/lib/ws/handlers/tasks.test.ts
+++ b/apps/web/lib/ws/handlers/tasks.test.ts
@@ -596,3 +596,93 @@ describe("task.updated executor preservation", () => {
     });
   });
 });
+
+describe("task.updated parent preservation", () => {
+  it("preserves parentTaskId when a partial update omits parent_id", () => {
+    // A task.updated for an unrelated field change (e.g. a title rename) must
+    // not silently un-nest a live child — only an explicit detach (parent_id:
+    // null) should clear the link. See parentIDEventField in service_tasks.go.
+    const existingTask = {
+      id: "t1",
+      workflowStepId: "step1",
+      title: "Old title",
+      position: 0,
+      primarySessionId: "session-1",
+      parentTaskId: "parent-1",
+    };
+    const store = makeStore({
+      kanban: {
+        workflowId: "wf1",
+        steps: [],
+        tasks: [existingTask],
+      } as unknown as AppState["kanban"],
+      kanbanMulti: {
+        isLoading: false,
+        snapshots: {
+          wf1: { workflowId: "wf1", workflowName: "WF1", steps: [], tasks: [existingTask] },
+        },
+      } as unknown as AppState["kanbanMulti"],
+    });
+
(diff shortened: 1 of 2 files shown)


carlosflorencio:
A few additional points:

- Every rejected fetch is currently treated as a permanent missing parent. A temporary network error or 5xx will therefore leave the child unnested until the workspace changes. Only cache terminal not-found responses; transient failures should retry with a bounded policy. For example:

  ```ts
  if (error instanceof ApiError && error.status === 404) {
    failedRef.current.add(id);
  }
  ```

- This also creates one request per missing ancestor, potentially from both desktop and mobile hook instances. That’s probably fine for the immediate edge case, but a workspace-scoped batch endpoint or snapshot option returning minimal parent-anchor projections would avoid an N+1 pattern as the number of archived parents grows.

ClemDNL:
Closing — PR description got corrupted (unrelated content clobbered the body). Reopening as a new PR with the correct description.

PR #2197, opened 2026-08-03
fix(i18n): localize the two Agents hooks the lint count could not see

jcfs:
Follow-up to #2193. That PR walked the **component** import closure of Settings → Agents and checked the one hook the string count listed (`use-agent-update-dialog-state.ts`, which genuinely has no copy). It did not walk the other hooks those routes call — and two of them did.

## Important Changes

- **`hooks/domains/settings/use-agent-runtime-updates.ts`** threw `Agent installation is already in progress.` when the user approves a runtime update while an install job for the same agent is still running.
- **`hooks/domains/settings/use-dynamic-models.ts`** fell back to `Failed to fetch capabilities` when the capability-probe request throws without an `Error`.

Both now resolve through the module-level `t`, which is correct here: one is thrown from a callback and the other is set from an async `catch`, neither during render.

Both hooks are reached **only** from this surface — `use-agent-runtime-updates` from `app/settings/agents/page.tsx`, `use-dynamic-models` from `components/settings/profile-form-fields.tsx` — so both are added to `i18nGuardFiles` rather than left as shared files.

### Why the count said zero

Neither string is a JSX literal, so `mode: "jsx-only"` never inspected either file. This is the documented blind spot one step further out than the SCREAMING_CASE config table: **a lint count is a floor on the visible half, not a work-list.**

What makes this pair worth a follow-up rather than a shrug is where they land. Each is interpolated into a message that *is* translated:

```
agents:unableToStartUpdate  →  "Unable to start update: {{error}}"
agents:failedToRefresh      →  "Failed to refresh: {{error}}"
```

So under the pseudo-locale the sentence renders accented with a plain-English payload inside it — `Ữńàƀĺē ţō śţàŕţ ūþďàţē: Agent installation is already in progress.` That reads as "translated" at a glance, which is exactly the case the pseudo-locale oracle is weakest at catching. Worth knowing for the migrations still to come: skim the *interpolated values*, not just the sentence frames.

Credit to the sibling Settings → System → Storage migration, which hit the same class on `hooks/domains/system/use-storage-maintenance.ts` and flagged the pattern.

### Checked and clean

I re-swept the whole #2193 file list for the other two shapes in the same family, and both are clean:

- **Functions/factories returning object literals** with `label` / `description` / `title` / `hint` fields — the shape one step out from a SCREAMING_CASE table. `installButtonContent`, `capabilityStatusMessage`, `updatePhase`, `noAuthHint` and `watcherKindLabel` were all already converted to take `t` in #2193; `extractCuratedSettings` returns `label`/`description` read from the backend `PermissionSetting`, which is data, not copy.
- **Thrown `Error` messages, `toast.*`, `confirm()` and `alert()` arguments** anywhere in the migrated file list — all already routed through `t` in #2193.

So the gap was specifically the un-walked hooks, not the file list itself.

## Validation

```
pnpm run typecheck                  # clean
pnpm lint --max-warnings 0          # clean
pnpm run i18n:check                 # 4/4 gates pass
pnpm run i18n:ratchet               # new-code clean; allowlist +2, intact
pnpm test                           # 1063 files, 8066 passed
```

## Possible Improvements

Low risk — two strings, copy-only, no behavior change. Both are error paths that only fire on a conflicting install job or a failed capability probe.

## Checklist

- [ ] My changes have tests that cover the new functionality and edge cases. — no new logic: two literals become catalog lookups on existing error paths, and the existing specs for both hooks pass unchanged.
- [ ] If my change touches UI files (`apps/web/`), I have added or updated Playwright e2e tests in `apps/web/e2e/` and verified them with `make test-e2e`. — copy-only; no route is added to `SCREENS` while the settings nav is still un-migrated, same as #2193.
- [x] I checked whether this affects public docs in `docs/public/**` and updated them or noted why no docs change is needed. — no public-docs impact.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2197?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jcfs:
Good catch, all three of you — added in be90ad55c.

You're right that "no new logic" was the wrong frame for this one. The failure mode isn't a logic bug, it's that **a renamed catalog key makes `t()` return the key name silently**, so the user would see `agents:agentInstallAlreadyInProgress` in a dialog and no test would fail. That's precisely the drift class this PR exists to close, so leaving the branches uncovered was inconsistent.

- **`use-dynamic-models.test.ts`** — two cases. One asserts the catalog fallback on a non-`Error` rejection, one asserts a real `Error`'s message still wins over the fallback. Note for anyone writing a similar spec: the hook fetches from a mount `useEffect`, so a single `mockRejectedValueOnce` reaches the `catch` — my first attempt called `refresh()` as well, which let the mount fetch eat the mock and left the assertion chasing an exhausted mock. @claude's sketch had this right and I over-complicated it.
- **`use-agent-runtime-updates.test.tsx`** — new file; the hook had no coverage at all, as @claude noted. Four cases: the install-conflict throw (the localized one), the **update**-conflict branch that must *not* throw because it adopts the running job and returns it, a non-conflict rethrow, and the clean path. The sibling branch is worth pinning because the two 409 kinds differ only by `active_kind` and diverge into throw-vs-return.

Both assert the resolved English rather than the key, so a rename fails the test instead of reaching the user.

Revalidated: `typecheck`, `lint --max-warnings 0`, `i18n:check` (4/4), `i18n:ratchet` (allowlist +2, intact), `pnpm test` — 1064 files, 8072 passed.

PR #2233, opened 2026-08-03
feat(web): drive navigation surfaces from one manifest

Aulma:
Application destinations were independently hardcoded across the sidebar, mobile menu, command palette, and plugin navigation, allowing those surfaces to drift and leaving Stats unreachable on phones. A shared manifest and typed resolvers now keep navigation aligned while preventing unnecessary integration polling.

This is the first of five independently shippable navigation improvements identified in my research about the current UX UI state; The next PRs will be: breakpoint alignment, a shared page shell, Settings route cleanup, and the
navigation spec.

## Important Changes

- Added a central destination manifest for first-party and plugin navigation, including labels, icons, routes, surfaces, sections, and availability requirements.
- Added separate typed hooks for gated and static destinations, preventing static surfaces from accidentally resolving integration-gated sections.
- Migrated sidebar, mobile menu, command palette, integration, and plugin navigation to shared destination resolution and row components.
- Added guardrails for mobile reachability, top-level route coverage, availability gates, plugin ID collisions, and invalid hook usage.
- Added mobile Playwright coverage for navigating to Stats.
- Split the manifest into `types.ts`, `surface-policy.ts`, `core-destinations.ts`,
  `plugin-destinations.ts`, and `resolve-destinations.ts`, so surfaces depend on the
  resolution rules rather than the data.
- Namespaced plugin destination IDs by owning plugin
  (`plugin:<pluginId>:<itemId>`, both segments percent-encoded). Resolved IDs are
  React keys and `NavItem.id` is plugin-local, so the owner is part of the identity;
  raw item IDs still drive the public `plugin-nav-item-<id>` selectors.
- Recorded the boundaries and the four deliberately deferred decisions in
  `docs/decisions/2026-08-04-navigation-manifest-boundaries.md`.

## Validation

- Web TypeScript typecheck — passed
- ESLint on all changed files — clean
- Focused Vitest navigation and component suites — 80 passed
- Mobile Playwright coverage — 5 passed, covering Stats, configured integrations, plugin navigation, retry, and diagnostics
- Vite production build — passed
- i18n ratchet against the merge base — passed
- Manually verified the updated frontend at phone width: **Menu → Utilities → Stats** opens `/stats`
- Updated `docs/public/feature-status.md` — Stats is now reachable from
  **Menu → Utilities → Stats** on small screens, so the page that documents where
  Stats is opened needed the new path. `node scripts/validate-public-docs.mjs`
  passes (41 pages). No other published page enumerates navigation.

## Possible Improvements

Risk is low. Existing destination hrefs and compatibility test IDs are preserved. The primary user-facing change is that Stats is now reachable from the mobile menu.

Catalog array position remains shared across surfaces, so reordering the catalog currently reorders every surface. Plugin navigation paths also remain unvalidated links: plugins cannot serve or shadow first-party routes, but they can link to them. These constraints and their intended follow-ups are recorded in the navigation ADR.

## Checklist

- [x] If my change touches UI files (`apps/web/`), I have added or updated Playwright e2e tests in `apps/web/e2e/` and verified them with `make test-e2e`.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2233?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


carlosflorencio:
Two smaller improvements:

- The comment in `use-nav-availability.ts` says additional consumers do not add network traffic, but Jira, Linear, and Azure DevOps ultimately use `useIntegrationAuthed`, which creates a fetch and 90-second interval per mounted consumer. The static/gated split is still useful, but either soften the comment:

  ```ts
  // Each mounted gated consumer may own a polling subscription.
  // Use useStaticDestinations for sections that do not need availability.
  ```

  Or, if deduplication is intended, hoist availability into a shared store/provider.

- The manifest test enforces exactly one of `label` and `labelKey`, but the TypeScript type permits both or neither. This invariant could be encoded directly:

  ```ts
  type DestinationCopy =
    | { label: string; labelKey?: never }
    | { label?: never; labelKey: string };

  type Destination = DestinationBase & DestinationCopy;
  ```

carlosflorencio:
Could you sanity-check a few long-term boundaries before this becomes the foundation for more navigation work?

- `destinations.ts` is already 432 lines against the 600-line limit, with only nine first-party destinations. I’d keep one compact first-party catalog, but split the surrounding mechanics now: `types.ts`, `core-destinations.ts`, `plugin-destinations.ts`, `resolve-destinations.ts`, and `surface-policy.ts`.

- You already noted plugin ID collisions. I think this is more foundational than a low-risk follow-up because resolved IDs become React keys. The registry already knows `pluginId`, so could it expose owner-aware nav registrations and build an internal identity such as:

  ```ts
  `plugin:${encodeURIComponent(pluginId)}:${encodeURIComponent(item.id)}`
  ```

  The raw item ID can remain separately for compatibility test IDs.

- A single manifest order is currently shared by the sidebar, mobile menu, and command palette. Those surfaces will probably need different ordering as they grow. It may be worth supporting per-surface order/priority rather than making array position a global UI contract.

- Dynamic plugin paths should ideally be owner-namespaced or validated so a plugin nav item cannot point at a first-party route it does not own.

- The route-coverage test that parses `spa-routes.tsx` with a regex is a useful guardrail, but it is brittle as the permanent contract. Longer term, typed route IDs shared by routing and navigation would be safer.

- One documentation detail: `useNavAvailability` says additional callers do not add network traffic, but Jira, Linear, and Azure DevOps availability hooks create polling per mounted consumer. Either soften that claim or centralize availability in shared state if deduplication is intended.

I don’t think every future concern needs to be implemented here, but I’d consider addressing the module boundary and owner-aware plugin identity now, and recording the remaining decisions in the spec/ADR that defines this navigation contract.

Aulma:
Adjusted the PR body accordingly for these changes:

- Fixed command-palette cache invalidation so translated groups and keywords refresh after locale changes; added focused regression tests.
- Corrected availability documentation and encoded the [label](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html)/[labelKey](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html) XOR in TypeScript.
- Split navigation into focused catalog, types, policy, plugin mapping, and resolver modules.
Added owner-aware, collision-safe plugin destination IDs while preserving compatibility selectors.
- Added an ADR documenting the new boundaries and intentionally deferred ordering, route ownership, typed routes, and polling deduplication.

PR #2238, opened 2026-08-03
feat(web): allow reordering admitted cards within kanban columns

leanrob:
Kanban board cards could not be reordered inside a column even though the move API already accepts `position`. Admitted cards can now be dragged to a new slot in the same column and that order persists; dropping into a different column still appends (no insert-at-drop-index).

## Important Changes

- Sort admitted cards by `position`, then `createdAt`, then `id`, so densified positions control board order.
- Same-column drops densify admitted positions to `0..n-1` and persist via existing `POST /api/v1/tasks/:id/move`.
- Cross-column drops keep append-only behavior (target column length), even if the pointer lands between cards.
- Queued WIP overflow cards stay non-sortable and render after admitted cards.
- Extract swimlane DnD / persist helpers and add unit + Playwright coverage for same-column reorder.

## Validation

- `cd apps/web && pnpm exec vitest run lib/kanban/task-order.test.ts lib/kanban/reorder-admitted.test.ts lib/kanban/wip-limit.test.ts` (26 passed)
- `cd apps/web && pnpm exec tsc --noEmit`
- Pre-commit: prettier, web lint, i18n new-code ratchet
- Playwright: `apps/web/e2e/tests/kanban/card-reorder.spec.ts` (same-column reorder + queued overflow stays fixed)

## Behavior Notes

- **Same column:** dragging an admitted card to another slot in that column updates order and persists.
- **Different column:** the card moves to the destination step but is **not** inserted at the hovered slot; it is appended after existing admitted cards in that column.
- Pipeline board, queued overflow reorder, and cross-column insert-at-index are out of scope.

## Demo

Paste or attach short recordings here once captured:

### Desktop — same-column reorder

https://github.com/user-attachments/assets/abd6d4b3-7e03-47be-af96-ec3e000cca6f

## Possible Improvements

Low risk: mid-batch reorder failures roll the UI back while earlier server patches may remain until reload; a bulk reorder API would make that atomic.

## Checklist

- [x] If my change touches UI files (`apps/web/`), I have added or updated Playwright e2e tests in `apps/web/e2e/` and verified them with `make test-e2e`.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2238?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


carlosflorencio:
Nice @leanrob ! Let me know if you need help to progress with the PR.

carlosflorencio:
1. The branch needs a rebase onto current `main`. There are conflicts in the Kanban column/swipeable components and locale files. Recent `main` changes added the plugin task-menu `presentation` contract, so please preserve that while resolving:

```tsx
<SwipeableColumns presentation="mobile" ... />
<KanbanColumn presentation={presentation} ... />
<KanbanCard presentation={presentation} ... />
```

2. Same-column reorder currently persists each task independently:

```ts
for (const patch of result.patches) {
  await moveTaskById(patch.taskId, ...);
}
```

This can partially succeed or interleave with another reorder, leaving duplicate/mixed positions rather than one complete dense ordering. I think the ordering should be persisted atomically with a backend endpoint along these lines:

```http
PUT /api/v1/workflows/:workflowId/steps/:stepId/task-order

{ "task_ids": ["task-b", "task-a", "task-c"] }
```

The backend can validate the admitted task set, write `0..n-1` in one transaction, then publish updates after commit.

3. `rollbackSnapshot()` restores the complete old task array. If a WebSocket update or a later drag arrives while the request is pending, that newer state can be overwritten. With an atomic endpoint, a failed request could refetch the authoritative snapshot; otherwise the rollback needs an operation generation guard and should only revert positions owned by that operation.

4. The new Playwright coverage does not currently exercise the mobile path: `card-reorder.spec.ts` runs the desktop project and `dragKanbanCard()` only uses mouse input, so `TouchSensor` is never tested. Please add a `mobile-card-reorder.spec.ts` using a real touch gesture. The queued-card scenario should also attempt to drag the queued card within its column and assert that the order remains unchanged.

It would also be useful to add focused tests around persistence failure and overlapping reorder attempts using deferred promises.

One smaller localization cleanup: resolve `"Needs Reassignment"` through i18n at render time, and pass `count` for the WIP label so `_one`/`_other` pluralization works:

```ts
t("common:tasksOverWipLimit", {
  count: admittedTaskCount,
  countLabel: wipCountLabel,
});
```

PR #2248, opened 2026-08-04
feat(settings): add enable/disable toggle for agent profiles

Fclem:
## What

Adds an enable/disable toggle for agent profiles (kanban-flavour profiles).

- **Backend:** `enabled` column on `agent_profiles` (default `1`, backfilled by migration, preserved by the legacy-DB table recreation), carried through the model, DTO, PATCH API, and handler. New profiles are created enabled; `UpdateProfile` only touches `enabled` when the PATCH includes it.
- **Frontend:** `enabled` flows through `AgentProfile`/payload types, `normalizeAgentProfile` (legacy payloads default to enabled), and the store's `AgentProfileOption`.
- **Selection filtering:** disabled profiles are hidden from every new-work choice surface — task/subtask/session creation options, autopick (last-used / workspace-default / first fallback), handoff menu, new-session dialog default, and quick-chat workspace-default handling. The store keeps disabled profiles so existing sessions and labels are untouched.
- **Settings UI:** an `Enabled` switch in the profile editor header (dirty/save flow) and an immediate-save switch on each row in `/settings/agents` ("Manage existing profiles by agent"). Disabling the only profile shows the dialog's "No compatible agent profiles" state instead of a selector, blocking submit until re-enabled.

## Notable implementation detail

The list-row toggle lives **outside** the row's `Link` (sibling, not child). A switch nested inside an anchor still follows the href as the click's native default action — `stopPropagation` alone does not cancel it, and `preventDefault` would make Radix skip its own toggle handler. The sibling layout also avoids interactive-inside-interactive.

## Tests

- Backend: store round-trip + legacy-DB backfill migration tests; controller create-defaults-enabled / update tests.
- Frontend: normalize, store option mapping, option filtering, autopick-never-disabled, handoff filtering, new-session fallback, profile header dirty/payload, list toggle action.
- E2E (`tests/settings/agent-profile-disable.spec.ts`): header toggle persists across reload, disabled profile hidden from task creation (selector-absence with other profiles, no-compatible state when it's the only one), list toggle disables/re-enables without navigating.

## Verification

- `go test ./internal/agent/settings/...` — pass
- Web lint (`--max-warnings 0`), typecheck, full vitest suite — pass (3 pre-existing `lib/http-git-server.test.ts` failures reproduce at HEAD)
- `pnpm e2e tests/settings/agent-profile-disable.spec.ts` — 2 pass
- Adjacent E2E (`create-task`, `new-session-deleted-profile`, `workflow-agent-profile`, `agent-profile-acp`) — pass except 2 pre-existing plan-mode failures in `create-task.spec.ts` (`plan-panel` visibility; reproduce at HEAD with changes stashed)

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2248?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


carlosflorencio:
- Saving from the profile editor rebuilds `agentProfiles` manually without copying `enabled`. That makes a newly disabled profile selectable again until reload. Using the canonical mapper avoids this field drifting again:

  ```ts
  nextAgents.flatMap((agent) =>
    agent.profiles.map((profile) => toAgentProfileOption(agent, profile)),
  )
  ```

- The contract needs another pass across all new-work surfaces. Configuration Chat, automation configuration, and watcher dialogs still consume the raw profile list, and existing watchers/direct API callers can continue launching disabled profiles. If `enabled` means “out of rotation,” I’d centralize the rule:

  ```ts
  export const isSelectableAgentProfile = (profile: AgentProfileOption) =>
    profile.enabled !== false;
  ```

  Then use it in every new-work selector and define backend behavior for new launches versus resuming existing sessions. Otherwise, the spec and UI copy should describe the narrower “hidden from manual task/session pickers” behavior.

PR #2271, opened 2026-08-05
fix(i18n): restore zh-cn parity for the chat attachment keys

nova28:
## Problem

`pnpm run i18n:check` is failing on `main`. `en/chat.json` carries thirteen
attachment keys that `zh-cn/chat.json` never received, and the parity check
treats a missing real-locale key as an error:

```
✖ 13 real-locale catalog parity issue(s):
  zh-cn / chat: missing key: attachmentUploadPending
  zh-cn / chat: missing key: attachmentUploadPendingSubmit
  zh-cn / chat: missing key: attachmentUploadFailed
  ...
```

This is the "Run i18n checks" step of **Run Frontend Lint, Tests, and Build**,
so it fails on `main` itself ([run 30951568390](https://github.com/kdlbs/kandev/actions/runs/30951568390))
and on every open PR that carries `main`'s catalogs — #2241, #2262, #2267 and
#2268 all fail at the same step for this reason and no other.

## Change

Translate the thirteen keys. Locale data only — no UI source is added or
modified, so the new-code ratchet has nothing to judge.

## Verification

```
✓ i18n keys OK — 2803 key(s) referenced, 3313 en entr(ies), 4 orphan(s), pseudo and zh-cn in sync.
✓ <Trans> indices OK — 53 element(s) checked.
✓ no inline English plurals — count-bearing messages use _one/_other.
✓ no module-scope t() — all copy resolves at render.
✓ i18n new-code ratchet — no UI source added or modified
✓ guard allowlist intact — 248 entr(ies)
```

The four remaining orphan warnings (`turnGroupToolCalls_*`, `activeSubagents_*`)
are pre-existing plural variants the reference scanner does not match; they are
warnings, not failures.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2271?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/web/src/locales/zh-cn/chat.json
+++ b/apps/web/src/locales/zh-cn/chat.json
@@ -36,5 +36,18 @@
   "providerQuotaResetUnknown": "请在提供商恢复可用容量后重试。",
   "providerQuotaProviderFallback": "提供商",
   "turnGroupToolCalls_one": "{{count}} 次工具调用",
-  "turnGroupToolCalls_other": "{{count}} 次工具调用"
+  "turnGroupToolCalls_other": "{{count}} 次工具调用",
+  "attachmentUploadPending": "正在上传附件…",
+  "attachmentUploadPendingSubmit": "请等待附件上传完成后再发送。",
+  "attachmentUploadFailed": "附件上传失败",
+  "retryAttachmentUpload": "重试上传附件",
+  "attachmentPreviewAlt": "预览",
+  "attachmentMessageAlt": "附件 {{index}}",
+  "attachmentFallbackName": "附件",
+  "attachmentPreviewFullAlt": "全尺寸预览",
+  "attachmentDeliveryMode": "附件传递方式",
+  "attachmentDeliveryPrompt": "提示词",
+  "attachmentDeliveryFile": "文件",
+  "attachmentDeliveryPromptDescription": "作为提示词上下文发送，供智能体识别图像内容。智能体不会获得文件路径。",
+  "attachmentDeliveryPathDescription": "上传到工作区，使智能体可以读取或编辑该文件。"
 }


carlosflorencio:
Heyy @nova28 FYI, i'm fixing the CI. And will rebase the PRs afterwards, thanks for your recent contributions, will check the other PRs later.

Join the discord if you wanna talk about the PRs more directly.

jcfs:
Context, not a request — this is yours to decide.

`main` was red because `i18n:check` failed on the 13 `zh-cn` `chat` keys this PR adds. That is fixed independently in #2272 (`a2db334ad`): real-locale parity is now **advisory** rather than a merge gate, so a lagging translation catalog warns and no longer blocks anyone. Verified on merged `main` — `i18n:check` exits 0 with `pseudo in sync. 13 advisory zh-cn issue(s).`

So **this PR is no longer needed as a `main` fix**, but it is still the only thing that actually translates those 13 keys, and `i18n:check` will keep reporting them until something does. It stands on its own merits as a translation contribution.

I deliberately did **not** write these keys myself. The instruction was that Chinese is handled out of band, and the advisory warning is the signal meant to reach whoever owns it — filling it in would have removed the signal and left my guesses in your catalog. Leaving this open rather than closing it for the same reason.

Groundwork from investigating, in case it is useful — full version in #2272's body under "For whoever owns the `zh-cn` translation":

**Renderings already established elsewhere in the shipped catalogs**, worth matching rather than re-deciding: upload → 上传 (`agents:uploadYamlFile`), preview → 预览 (`agents:commandPreview`), retry → 重试 (`common:retry`), prompt → 提示词 (`azuredevops:prompt`), agent → 代理, workspace → 工作区, context → 上下文, `Failed to X` → `X失败`, `Xing…` → `正在X…`.

**No precedent anywhere in any shipped catalog — you would be setting one:** `attachment`, `file path`, `full size`. `attachment` appears in 9 of the 13 keys, so it is worth deciding deliberately rather than by whichever key lands first.

**Register**, from the call sites: `attachmentPreviewAlt` / `attachmentPreviewFullAlt` / `attachmentMessageAlt` are `alt` text (the last interpolates `{{index}}`); `attachmentDeliveryMode` and `retryAttachmentUpload` are `aria-label`; `attachmentDeliveryPrompt` / `attachmentDeliveryFile` are the two labels of one mutually exclusive toggle and read as a matched pair; `attachmentUploadPendingSubmit` is the reason a disabled submit button gives; `attachmentFallbackName` substitutes for a missing filename.

One note if you rebase: `i18n:check` now also verifies that a translated value keeps the same `{{placeholders}}` and `<n>` `<Trans>` tags as `en`, and that check still fails hard. It does not check that a value is actually translated — that part is structural only.

nova28:
Closing — superseded, and the remaining delta would be a regression.

**All 13 keys are already translated on `main`.** 10 of the 13 here are byte-identical to what's there.

**The 3 that differ fork the terminology for "agent."** `main` uses 代理; this PR uses 智能体:

| Key | `main` | this PR |
|---|---|---|
| `attachmentPreviewFullAlt` | 完整尺寸预览 | 全尺寸预览 |
| `attachmentDeliveryPromptDescription` | …代理不会获得文件路径。 | …智能体不会获得文件路径。 |
| `attachmentDeliveryPathDescription` | 以便代理读取或编辑该文件。 | 使智能体可以读取或编辑该文件。 |

The shipped `zh-cn` catalog uses 代理 in 203 places and 智能体 in 0 (this file itself is split: 代理 ×1, 智能体 ×3). Merging would overwrite three correct values with an inconsistent term, so there's nothing left to gain here.

Two notes for anyone finding this later:

- The failing E2E shards were environmental, not this change — Shard 2 (`git-changes-panel`) and Shard 6 (`session-resume`) failed identically on #2263, a backend-only PR that cannot touch either. A locale-file addition can't break them.
- The `zh-cn` advisory gap has since grown past these 13 keys to ~243 (`en` 3636 vs `zh-cn` 3393). That's the real signal now, and it belongs to whoever owns the `zh-cn` catalog. Contrary to earlier notes, `attachment` → 附件 is *not* an open question — it's already used 14× in `zh-cn/chat.json`. `file path` and `full size` are still unsettled, and `full size` is exactly where `main` and this PR disagreed.

PR #2285, opened 2026-08-05
fix(launcher): find the runtime bundle next to the launcher

JnManso:
The launcher refuses to start unless `KANDEV_BUNDLE_DIR` is set, even though it is sitting inside the bundle it is looking for. This adds a fallback that derives the bundle from the launcher's own path.

## Why

Every install channel works around this today, each in a different way:

| Channel | Workaround |
|---|---|
| Homebrew | `write_env_script` wraps the binary to export the variable |
| npm | `bin/native-shim.js` sets it before spawning |
| Scoop | `env_set` in the manifest |
| Manual `tar -xzf` | none — `kandev` exits with `KANDEV_BUNDLE_DIR is not set` |

That last row is a real papercut: download the release tarball, extract it, run `kandev`, and it fails with a message pointing at a variable nobody mentioned.

It also blocks winget and Chocolatey, which unpack an archive and put the binary on PATH with no mechanism for setting environment variables.

## What changes

`resolveRuntimeBundle` now tries two sources instead of one:

1. `KANDEV_BUNDLE_DIR`, exactly as before
2. failing that, `<bundle>/bin/kandev` → `<bundle>`, from `os.Executable()`

The environment variable keeps priority on purpose: `kandev service install` writes it into the systemd unit and launchd plist, and a self-update repoints it at the new bundle while the outgoing launcher is still running. Nothing that works today changes behaviour — the only case that behaves differently is the one that is currently a hard error.

The derived path is not trusted blindly. `bundleDirForLauncherPath` accepts only the `<bundle>/bin/<launcher>` layout, and the result still goes through `validateRuntimeBundle`, which already requires `agentctl` and the four remote helpers to be present. A launcher somewhere unrelated fails exactly as it does now.

## On symlinks

winget installs portable packages by unpacking to `WinGet\Packages\<id>\` and putting a symlink on PATH at `WinGet\Links\<name>.exe`, so it was worth checking whether the launcher would see the link or the target. Measured against a real winget symlink on Windows 11:

```
LINK   : ...\WinGet\Links\kubectl.exe
TARGET : ...\WinGet\Packages\Kubernetes.kubectl_.../kubectl.exe
Win32_Process.ExecutablePath = ...\WinGet\Packages\Kubernetes.kubectl_.../kubectl.exe
```

Windows resolves the link at process creation, and that path comes from the same source `os.Executable()` uses. `EvalSymlinks` is still attempted first for platforms that do not resolve it, and is best-effort — it fails on Windows junctions, and the unresolved path is worth trying anyway.

## Tests

Four new cases in `bundle_test.go`: the accepted layout, three rejected layouts, the environment variable winning over the executable-relative lookup, and the error mentioning both lookups rather than blaming the variable alone.

Note for anyone running the suite on Windows: five pre-existing `TestValidateRuntimeBundle*` cases fail there because they create `bin/kandev` without the `.exe` that `executableName` appends. They fail identically on `main` without this change, and pass in CI on Linux.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2285?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/launcher/bundle.go
+++ b/apps/backend/internal/launcher/bundle.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"strings"
 )
 
 type runtimeBundle struct {
@@ -73,9 +72,16 @@ func bundleDirFromExecutable() (string, bool) {
 // bundleDirForLauncherPath maps <bundle>/bin/kandev back to <bundle>. Any other
 // layout is refused, so a launcher sitting somewhere unrelated cannot make
 // validateRuntimeBundle probe an arbitrary directory.
+//
+// The directory name must be exactly "bin", matching what validateRuntimeBundle
+// joins onto the result. Accepting other casings here would derive a bundle on a
+// case-sensitive filesystem that validation then rejects, turning a clear "not a
+// bundle" into a confusing "launcher binary not found". Windows is unaffected:
+// EvalSymlinks above canonicalises the path to the casing on disk, so a launcher
+// invoked as BIN\kandev.exe still arrives here as bin\kandev.exe.
 func bundleDirForLauncherPath(exe string) (string, bool) {
 	binDir := filepath.Dir(exe)
-	if !strings.EqualFold(filepath.Base(binDir), "bin") {
+	if filepath.Base(binDir) != "bin" {
 		return "", false
 	}
 	dir := filepath.Dir(binDir)
--- a/apps/backend/internal/launcher/bundle.go
+++ b/apps/backend/internal/launcher/bundle.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 type runtimeBundle struct {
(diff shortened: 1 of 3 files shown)


carlosflorencio:
Thanks for the contribution! The launcher-level approach fits the existing boundary, and keeping `KANDEV_BUNDLE_DIR` as the explicit override is important.

I have two follow-ups that aren’t covered by the existing review:

1. Could we test the successful executable-relative lookup end to end? The package already has an `executablePath` seam, but `bundleDirFromExecutable` calls `os.Executable()` directly. Reusing the seam would avoid a second executable lookup boundary and make the core fallback deterministic to test:

   ```go
   exe, err := executablePath()
   ```

   The test can point it at a complete temporary `<bundle>/bin/kandev` and assert that `resolveRuntimeBundle()` returns the directory with `Source == "executable"`. The current tests cover path parsing, env precedence, and total failure, but not the successful path this PR introduces.

2. Since this explicitly makes extracted release archives usable, could we add a short manual archive section to `docs/public/cli.md`? It should mention selecting the correct platform archive, verifying its adjacent checksum, extracting it, and running `./kandev/bin/kandev`. The current install table only lists Homebrew and npm/npx.

JnManso:
> **Validate the launcher file name.** `bundleDirForLauncherPath` accepts any executable under `<bundle>/bin`. For example, `<bundle>/bin/diagnostic` derives `<bundle>` and then launches `<bundle>/bin/kandev`.

Not taking this one, for two reasons.

**The scenario cannot occur.** `internal/launcher` is imported by exactly one binary — `cmd/kandev`. `agentctl`, `acpdbg`, `mock-agent`, `plugin-pack`, `preview` and `winjob` do not call `resolveRuntimeBundle`, so no `<bundle>/bin/diagnostic` ever reaches this code path. `os.Executable()` here is always the launcher itself.

**Pinning the name would reject a currently valid case.** `validateRuntimeBundle` already requires `bin/kandev` to exist inside the bundle, so a launcher that a packager renamed still resolves a correct bundle today. Requiring `filepath.Base(exe)` to equal `executableName("kandev")` would break that, in exchange for guarding a path nothing can take.

The constraint that is load-bearing stays: only `<bundle>/bin/<exe>` is accepted, and the derived directory still has to satisfy `validateRuntimeBundle`, which checks for the launcher, `agentctl`, and all four remote helpers. A wrong guess fails closed with the same error as before this change.

JnManso:
Both applied, thanks — the seam was the better call and I had missed it.

**1. `executablePath()` instead of `os.Executable()`.** Swapped, and `TestResolveRuntimeBundleFallsBackToExecutable` now covers the path this PR actually introduces: a complete temporary `<bundle>/bin/kandev`, `KANDEV_BUNDLE_DIR` cleared, asserting `Source == "executable"` along with the resolved `Dir` and `Launcher`.

Worth reporting because it argues for the seam: the first version of that test failed on Windows. `t.TempDir()` returned a path under an 8.3 short name and `EvalSymlinks` expanded it, so the assertion compared `JOOMAN~1` against the long form. The production code was right; the test was comparing an unresolved path against a resolved one. It now resolves both sides, which also covers macOS, where `/var` is a symlink to `/private/var`. Without a deterministic seam that would have surfaced as a platform-specific flake rather than a clear failure.

**2. Archive install section in `docs/public/cli.md`.** Added under Install, after npm/npx: the per-platform archive naming, `shasum -a 256 -c` against the published `.sha256`, extraction, and `./kandev/bin/kandev --version`. It also states that the extracted directory can be moved anywhere, and that `KANDEV_BUNDLE_DIR` is now only for pointing the launcher at a bundle it is not part of.

One thing to flag: kdlbs/kandev#2269 also edits the Install section of `docs/public/cli.md`, adding a `### Scoop` entry and updating the support matrix. Whichever of the two merges second will need a small conflict resolution there — the sections are adjacent rather than overlapping.

I also left a separate comment declining the CodeRabbit suggestion to pin the launcher file name, since `internal/launcher` is imported only by `cmd/kandev` and `validateRuntimeBundle` already requires `bin/kandev` to exist. Happy to reverse that if you disagree.

JnManso:
> **Keep `bin` validation consistent with bundle validation.** `bundleDirForLauncherPath` accepts `bin` case-insensitively. `validateRuntimeBundle` later constructs `<bundle>/bin/...` with lowercase `bin`. On a case-sensitive filesystem, `<bundle>/BIN/kandev` passes derivation but fails validation.

Valid — applied. `strings.EqualFold` is now an exact comparison, which also drops the `strings` import, and the rejected-layout test covers `<bundle>/BIN/kandev`.

I had used the case-insensitive form because I assumed Windows would hand back whatever casing the caller typed, so rejecting it would break a legitimate invocation. That assumption was wrong, and it is worth recording why, since it is what makes the exact comparison safe rather than risky.

Measured on Windows 11, with a Go binary at `pkg\bin\prog.exe` printing `os.Executable()` and `filepath.EvalSymlinks` of it:

| Invoked as | `os.Executable()` | `EvalSymlinks` |
|---|---|---|
| `pkg\bin\prog.exe` | `...\pkg\bin\prog.exe` | `...\pkg\bin\prog.exe` |
| `pkg\BIN\prog.exe` | `...\pkg\BIN\prog.exe` | `...\pkg\bin\prog.exe` |

`os.Executable()` does preserve the caller's casing, so the concern behind my original choice was real — but `bundleDirFromExecutable` runs `EvalSymlinks` first, and that canonicalises to the casing on disk. A launcher invoked as `BIN\kandev.exe` therefore reaches the comparison as `bin\kandev.exe` and still resolves.

The exact comparison costs nothing on Windows, and on a case-sensitive filesystem it now refuses a `BIN/` directory up front with "no bundle was found next to the launcher", instead of deriving a bundle and failing later with "launcher binary not found", which pointed at the wrong thing.

PR #2311, opened 2026-08-06
fix(backend): enforce per-user workspace authorization in workflow sync

nova28:
## Summary

`internal/workflowsync` (workflow-sync config + force-sync) was the one per-workspace integration
service in this codebase without its own service-layer authorization boundary. GitHub, Jira, Linear,
Slack, Azure DevOps, and Automation all wire a `SetWorkspaceAuthorizer` + `authorizeWorkspaceAccess`
check into every user-facing service method, in addition to the global
`integrationWorkspaceScopeMiddleware`; `workflowsync.Service` had no check of its own.

**Severity note (important — read before reviewing):** an earlier draft of the linked spec claimed
this meant the HTTP endpoints (`/api/v1/workflow-sync/config`, `/api/v1/workflow-sync/sync`) were
directly exploitable by any authenticated member. That was wrong, and was caught by an independent
Codex review during this PR's own preparation: `origin/main`'s `integrationWorkspaceScopeMiddleware`
(`internal/backendapp/main.go:1807`) already path-matches `/api/v1/workflow-sync/` and authorizes
before any handler runs, exactly like every other integration prefix. So this PR is **not** closing a
currently-open HTTP vulnerability — it's a defense-in-depth and consistency fix:

- `workflowsync.Service`'s safety previously depended entirely on that one route-level middleware
  never being bypassed, misconfigured, or missing this path prefix, with no independent second layer
  — unlike every sibling integration.
- Any future non-HTTP caller of the service (a WS handler, an MCP tool, a plugin, a different mount
  point) would have inherited no protection at all.

Found while independently reviewing a sibling PR (#2289, GitLab workflow-sync support) — this gap
predates that PR and affects GitHub-only sync too, so it's fixed here as its own change.

## Changes

- `workflowsync.Service` gains a nil-safe `workspaceAuthorizer` field, `SetWorkspaceAuthorizer`
  setter, and `authorizeWorkspaceAccess` helper (mirrors `internal/github/service.go`), checked first
  in `GetConfigForWorkspace`, `SetConfigForWorkspace`, `DeleteConfigForWorkspace`, and `SyncWorkspace`.
- Wired at boot in `initWorkflowSyncService` via `taskSvc.AuthorizeWorkspaceAccess`.
- `handlers.go` maps the resulting `ErrWorkspaceNotFound` to a sanitized 404 in all four handlers,
  including `httpForceSync`'s second (post-sync) config lookup — a narrower race an independent
  review round caught, where access revoked between the sync and the follow-up read fell through to
  a generic 500.
- Design docs: `docs/specs/workflow-sync-workspace-authz/spec.md`,
  `docs/plans/workflow-sync-workspace-authz/{plan.md,task-01-*,task-02-*}` — spec/plan/task package
  written and committed before implementation, per this repo's spec-driven-development workflow, with
  the severity correction recorded in both.

## Test plan

- [x] `apps/backend/internal/workflowsync/service_authorization_test.go` (new): foreign-workspace
      denial with no store mutation across all four service methods; unscoped (no authorizer, no
      identity, synthetic identity) and authorized-allow paths unaffected.
- [x] `apps/backend/internal/workflowsync/handlers_test.go` (new — package had zero HTTP-level tests
      before this PR): table-driven across GET/POST/DELETE/sync — owner succeeds, synthetic identity
      succeeds, foreign member denied (404, no leak of repo owner/name/branch/path, full-struct
      before/after config equality), plus a dedicated regression test for the force-sync race above.
- [x] `apps/backend/internal/backendapp/workflowsync_wiring_test.go` (new): calls the real
      `initWorkflowSyncService` boot function against a real SQLite-backed `taskservice.Service` —
      verified by temporarily removing the production wiring line and confirming this test (and only
      this test) fails.
- [x] `go build ./...`, `go vet ./...`, `gofmt -l` — clean.
- [x] `go test -race ./internal/workflowsync/... ./internal/backendapp/...` — `ok`.
- [x] `golangci-lint run ./internal/workflowsync/... ./internal/backendapp/... --new-from-rev=main` —
      0 issues.
- [x] Full `go build ./... && go test -race ./...` (whole backend) — all packages pass except a
      pre-existing, unrelated flake in `internal/repoclone` (confirmed unrelated: this PR touches
      neither that package nor anything it depends on; the isolated test passes on rerun).
- No frontend/E2E surface — backend-only change, no UI/WS/client-visible contract changes.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2311?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/backendapp/services.go
+++ b/apps/backend/internal/backendapp/services.go
@@ -606,6 +606,7 @@ func initWorkflowSyncService(dbPool *db.Pool, githubSvc *github.Service, workflo
 		log.Warn("workflow sync service initialization failed (non-fatal)", zap.Error(err))
 		return nil
 	}
+	svc.SetWorkspaceAuthorizer(taskSvc.AuthorizeWorkspaceAccess)
 	return svc
 }
 
--- a/apps/backend/internal/backendapp/workflowsync_wiring_test.go
+++ b/apps/backend/internal/backendapp/workflowsync_wiring_test.go
@@ -1,0 +1,85 @@
+package backendapp
+
+import (
+	"context"
+	"errors"
+	"path/filepath"
+	"testing"
+
+	"github.com/jmoiron/sqlx"
+
+	"github.com/kandev/kandev/internal/auth/authn"
+	"github.com/kandev/kandev/internal/common/logger"
+	"github.com/kandev/kandev/internal/db"
+	"github.com/kandev/kandev/internal/github"
+	"github.com/kandev/kandev/internal/task/repository/repoerrors"
+	taskservice "github.com/kandev/kandev/internal/task/service"
+)
+
+// TestInitWorkflowSyncService_WiresRealWorkspaceAuthorization calls the real
+// initWorkflowSyncService boot function — not a hand-copied re-wiring — with a
+// real taskservice.Service backed by SQLite, so a regression that removes or
+// breaks the SetWorkspaceAuthorizer(taskSvc.AuthorizeWorkspaceAccess) line
+// inside it fails this test. The workflowsync package's own tests only ever
+// install a hand-written fake authorizer and cannot detect that class of
+// regression.
+func TestInitWorkflowSyncService_WiresRealWorkspaceAuthorization(t *testing.T) {
+	harness := newBootStateTestHarness(t)
+	ctx := context.Background()
+
+	ownerCtx := authn.WithIdentity(ctx, authn.Identity{UserID: "owner-1", Role: authn.RoleMember})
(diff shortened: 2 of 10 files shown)


carlosflorencio:
@claude review

Issue #2312, opened 2026-08-06
[Bug]: Windows Defender quarantines the unsigned agentctl.exe, and the running backend never says why


JnManso:
### Summary

On Windows, Microsoft Defender flagged the bundled `agentctl.exe` as `Trojan:Win32/Wacatac.B!ml` and quarantined it — including once while it was running. The backend process survived, but from that moment every agent start failed with `agentctl not ready: context deadline exceeded`, repeating roughly once a minute, with nothing in the UI or the log naming the real cause.

### Affected area

agentctl

### Kandev version

0.85.0

### Install mode

Other — Scoop on Windows

### Environment

- OS: Windows 11, amd64
- Install: Scoop (`scoop\apps\kandev\0.85.0\bin`)
- Runtime: standalone (no Docker)
- Defender: default settings, `CloudBlockLevel = 0` — not an aggressive configuration

### Kandev configuration

Standalone runtime, default ports. Nothing unusual; no config file. N/A otherwise.

### Task/session context

N/A — not task-specific. Any agent start fails once this happens.

### Steps to reproduce

Not deterministic; it depends on Defender's heuristics firing. What happened here:

1. Install Kandev on Windows via Scoop and start it.
2. Run a task so `agentctl.exe` is spawned.
3. Defender scans the binary (triggered here by unrelated processes touching the file) and quarantines it.
4. Every subsequent agent start fails with `agentctl not ready: context deadline exceeded` until Kandev is restarted — and even then it cannot start, because the binary is gone.

### Reproducibility

Once

### Evidence

From the Defender operational log, paths redacted:

```
22:47:42  2050  Filename: C:\Users\<user>\scoop\apps\kandev\0.85.0\bin\agentctl.exe
23:05:28  1116  Trojan:Win32/Wacatac.B!ml  (scan triggered by an unrelated process)
23:06:34  1116  Trojan:Win32/Wacatac.B!ml  process:_pid:78928
23:06:42  1117  Action: Quarantine
```

The Kandev backend log for the same window:

```
23:06:36  agentctl exited  pid=78928  exit_code=0
23:52:05  failed to relaunch agent for session
          error: failed to create execution: agentctl not ready: context deadline exceeded
```

The detection at 23:06:34 names the same PID the launcher reports exiting two seconds later. Afterwards `bin/` contains the four cross-platform helpers and `kandev.exe`; `agentctl.exe` is gone.

[...]

PR #2314, opened 2026-08-06
ci(release): sign the Windows binaries that ship in the release archive

JnManso:
Signs the Windows runtime binaries that ship in the release archive. Closes the gap described in #2312.

## The gap

`kandev.exe` and `agentctl.exe` already go through `signtool` — but only in `build-desktop`, and only on the copy under `apps/desktop/src-tauri/resources/kandev/bin`. Every signing step in the workflow sits inside that job (lines 1010–1077 and 1286–1322); `build-bundles` has none, and neither does `publish-release`.

So the desktop installer and the CLI archive carry byte-identical binaries — `build-desktop` downloads the very artifact `build-bundles` produced — and only one copy is signed. Scoop, npm and direct downloads get unsigned executables today, and would keep getting them even after a certificate is configured.

That is what #2312 ran into: Defender quarantined the unsigned `agentctl.exe` as `Trojan:Win32/Wacatac.B!ml` while it was running.

## The change

The Windows bundle now builds on `windows-latest` instead of cross-compiling from Linux, so `signtool` can sign the binaries in place, before `tar`/`zip`. No repackaging, no checksums to regenerate, and the archive that reaches users is the one that was signed.

It reuses the existing `desktop-signing-ready.sh` guard, so with no certificate configured it builds unsigned and logs a notice, exactly as `build-desktop` does today. Nothing changes until the secrets exist; then both paths sign without further work.

Knock-on changes, all consequences of the runner move:

- **`defaults: run: shell: bash`** on the job. The steps are POSIX shell and `windows-latest` would otherwise run them under pwsh. Git Bash is on the image and supplies `tar` and `shasum`.
- **`zip` → `7z a -tzip`.** Git Bash has no `zip`, so the Windows zip step added in #2286 would break. 7z is on both runner images, so this also removes a Linux-only assumption.
- **Cross-compilation removed.** Windows was the only entry with `cc`, so the mingw install step and the `CC:` passthrough are now dead and are dropped.

## Checked

- **The runner can still build it.** `windows-latest` ships gcc 14.2.0, so CGO with the required `fts5` tag compiles natively. Git Bash is present at `C:\Program Files\Git\bin\bash.exe`.
- **7z produces the same archive.** Verified the layout is unchanged — `kandev/`, `kandev/bin/…` — which is what Scoop's `extract_dir` and a winget `RelativeFilePath` both depend on. Integrity check and checksum verify.
- **`windows-sign.ps1` is reusable outside the desktop job.** It takes a path parameter, reads the certificate from the environment, materialises the `.pfx` under `RUNNER_TEMP` and resolves `signtool` itself. Nothing in it depends on Tauri or on `build-desktop` having run.
- **7-Zip on PATH is not guaranteed.** The runner image lists 7-Zip 26.02 but the documentation does not promise a `7z` on PATH, so the step resolves the command and falls back to the install path rather than failing a release on a lookup. This one is defended against rather than verified — I cannot check a runner's PATH from a branch.
- **No cross-compilation residue.** The only remaining `CC=` occurrences are the deliberate `CGO_ENABLED=0 CC=` on the four remote agentctl helpers, which are unrelated.

What cannot be checked from a branch: the job actually running. `release.yml` is `workflow_dispatch` only, so no PR CI exercises it. The first real test is the next release, and the risk is concentrated in the native Windows build rather than in the signing, which is inert without secrets.

## Timing, if a release is cut soon

Worth deciding deliberately rather than by accident:

- **Release before this merges** — publishes as today: unsigned, but carrying the `.zip` from #2286 and the launcher fallback from #2285, which is what a winget submission needs.
- **Release after this merges** — that release is the first to exercise the native Windows build. The signing itself stays inert without secrets, so the new risk is the compiler change, not the signature.

## Not in scope

The certificate itself. That is a purchase and an org decision; this only ensures a certificate would reach the archive users actually install. Also worth knowing: since March 2024 an EV certificate no longer clears SmartScreen immediately — EV and OV build reputation the same way — so signing improves the odds rather than switching the detection off.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -727,12 +727,19 @@ jobs:
             platform: macos-arm64
             goos: darwin
             goarch: arm64
-          - os: ubuntu-latest
+          # Windows builds natively instead of cross-compiling from Linux, so
+          # signtool can sign the binaries in place before they are packaged.
+          # The runner image ships gcc, which CGO still needs for the fts5 tag.
+          - os: windows-latest
             platform: windows-x64
             goos: windows
             goarch: amd64
-            cc: x86_64-w64-mingw32-gcc
     runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        # These steps are POSIX shell; windows-latest would otherwise run them
+        # under pwsh. Git Bash is on the image and provides tar and shasum.
+        shell: bash
     env:
       CGO_ENABLED: 1
       GOOS: ${{ matrix.goos }}
@@ -751,12 +758,6 @@ jobs:
           name: web-bundle
           path: apps/backend/internal/webapp/embedded/generated
 
-      - name: Install cross-compilation tools
-        if: matrix.cc
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y gcc-mingw-w64-x86-64
-
       # The `fts5` build tag is REQUIRED on the kandev binary. The office
       # task search migrations create a `tasks_fts` virtual table using
       # `USING fts5(...)`; without the tag, mattn/go-sqlite3 lacks the
@@ -768,7 +769,6 @@ jobs:
       # reports the released tag instead of the build-time default "dev".
       - name: Build backend binaries
         env:
-          CC: ${{ matrix.cc || '' }}
           KANDEV_VERSION: ${{ needs.nightly-prepare.result == 'success' && needs.nightly-prepare.outputs.tag || needs.prepare.outputs.tag }}
         run: |
           test -f apps/backend/internal/webapp/embedded/generated/index.html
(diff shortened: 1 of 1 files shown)


JnManso:
@carlosflorencio if you are cutting a release today, my suggestion is to merge this first and release after — but the choice depends on how much appetite there is for the risk below, so it is genuinely yours.

**Why after:** a release cut before this merges publishes unsigned binaries again, and the next Windows user installing via Scoop or npm can hit #2312 exactly as it happened here. Merging first means the archive is ready to carry signatures the moment a certificate exists, and this release becomes the one that proves the build change.

**Why you might prefer before:** this release would be the first to build the Windows bundle natively rather than cross-compiling from Linux, and that is the part I cannot test from a branch.

## What I could not validate

Being explicit rather than implying this is proven:

- **The job running at all.** `release.yml` is `workflow_dispatch` only, so no PR CI exercises it. Everything below is reasoning plus the runner image documentation, not a green run.
- **The native Windows build itself.** I confirmed the runner ships gcc 14.2.0, so CGO with the required `fts5` tag should compile. I could not confirm the build succeeds, or that a natively built binary behaves identically to the mingw cross-build. This is the real risk in the change; the signing is inert without secrets.
- **7-Zip on PATH.** The image lists 7-Zip 26.02 but the docs do not promise a `7z` on PATH. The step resolves the command and falls back to the install path, failing loudly if neither works. Defended, not verified.
- **`shasum` in Git Bash on the runner.** Git for Windows ships it, and the existing steps rely on it, but I have not run it there.

## One I did chase down and it is fine

Moving the build to Windows means `tar` no longer records a Unix exec bit for the four cross-platform `agentctl` helpers. That turns out not to matter: the SSH executor uploads them via SFTP with an explicit mode and chmods after writing, and the Windows bundle is only ever extracted on Windows, where the bit does not exist either way. Noting it so a reviewer does not have to re-derive it.

If you would rather de-risk it, a `desktop_validation_only` run would exercise the bundle build without publishing anything.

JnManso:
Moving this to draft. @carlosflorencio has applied for free OSS code signing through the SignPath Foundation, and if that is granted this PR is the wrong shape.

SignPath does not hand you a certificate to put in `WINDOWS_CERTIFICATE`. Their model is that the build uploads an artifact, an action submits it, and the signed artifact comes back — signing happens on their infrastructure. Two consequences for what is here:

- **The Windows runner move is unnecessary.** `signtool` no longer runs in CI, so cross-compiling the Windows bundle on Linux can stay exactly as it was. That also removes the one risk in this PR I could not test from a branch.
- **The `zip` → `7z` change goes with it**, since it only existed because Git Bash on `windows-latest` has no `zip`.

What survives is the underlying point from #2312: the archive that Scoop, npm and direct downloads use is not signed, and only `build-desktop` ever signs anything. A SignPath integration fixes that at the same place — right after `build-bundles` uploads `bundle-windows-x64`, which already satisfies their requirement that the artifact be a GitHub workflow artifact produced on GitHub-hosted runners.

Holding rather than closing, because if the application is declined this PR is still the right answer and can go back to ready as-is.

JnManso:
Superseded by the SignPath approach in #3579: the Foundation certificate lives in SignPath's HSM, so the signtool path can't use it.

PR #2344, opened 2026-08-06
refactor(plugins): move Slack out to kandev-plugin-slack and list it in the marketplace

jcfs:
Moves the Slack integration out of the monorepo and into a standalone plugin, and lists that plugin in the marketplace.

Plugin: **[kdlbs/kandev-plugin-slack](https://github.com/kdlbs/kandev-plugin-slack)** · [v0.1.0](https://github.com/kdlbs/kandev-plugin-slack/releases/tag/v0.1.0)

## Why

Slack shipped in-tree as `internal/slack`: a workspace-scoped config table, an unofficial browser-session (`xoxc` + `d` cookie) client, a 90s auth-health poller, a search-based `!kandev` trigger, HTTP/WS routes, a settings page, and six `{{Slack*}}` utility-agent template variables. `docs/public/feature-status.md` listed it as Experimental for exactly the reasons that made it awkward to own here — a nonstandard credential and a polling trigger, neither of which the other integrations need.

As a plugin it can move at its own pace, and it gained what the in-tree version could not justify: a real Slack app over **Socket Mode**, so events are pushed over a WebSocket with no public URL and nothing polled. `@Kandev …` and `/kandev …` replace the `!kandev` text marker. The browser-session pair survives as an explicit fallback, because the original spec's reason for it — workspaces whose admins forbid app installs — is unchanged.

## Removed

`internal/slack` and its `backendapp` wiring; `/api/v1/slack/*` and the `slack.config.*` WS actions; the `/settings/integrations/slack` page, route, sidebar entry, discovery-catalog entry, copy-config arm, API client, hooks, types and locale catalogs; and the `{{Slack*}}` template variables.

## Kept

The office channel relay's `slack` platform. It is one arm of a generic telegram/slack/discord/webhook notifier, not part of this integration, and removing only its Slack arm would leave a feature with no replacement.

## Upgrade path

The retired `slack_configs` table is dropped and the `slack:<workspace>:token` / `slack:<workspace>:cookie` vault entries are deleted on first boot (`dropRetiredSlackIntegration`). Orphaning them was the alternative, but those entries are live Slack credentials — leaving them encrypted at rest forever is a credential-retention problem, not merely dead data. **Anyone using the in-tree integration re-enters credentials in the plugin after upgrading**; the plugin's config shape differs anyway.

The spec is kept at `docs/specs/integrations/slack.md`, marked `archived` with a pointer to the plugin, as the record of what shipped in-tree.

## Marketplace

Second commit adds the catalog pointer. The v0.1.0 release exists and its manifest `id` is `kandev-plugin-slack`, matching the entry — which is what registry CI validates against.

## Verification

- `go build ./...` clean; `golangci-lint run --new-from-rev=origin/main` → 0 issues
- `go test ./internal/backendapp/... ./internal/utility/... ./internal/task/repository/sqlite/...` pass, including new fresh-DB, replay, and secret-scoping tests for the retirement migration
- `pnpm typecheck`, `pnpm lint`, `pnpm test` (1155 files / 8966 tests), `i18n:check`, `i18n:ratchet`, guard-allowlist all pass
- `validate-public-docs` (41 pages) and its 58 tests pass
- Booted a backend from this branch against an isolated DB: `/api/v1/slack/config` returns 404, no `slack_configs` table, `drop_slack_configs` migration applied. Installed the packaged plugin and exercised it through the real install path.

Two guard-list notes: the `i18nGuardFiles` entries for `app/settings/integrations/slack/**` and `components/slack/**` are removed because `check-guard-allowlist.mjs` rejects entries matching no file, and deletion is the one legitimate reason to remove one. The `integrationDescriptionSlack` locale key added on `main` while this branch was open is removed with the entry it described.

## Checklist

- [x] Tests added/updated for changed logic
- [x] Public docs updated (`integrations.md`, `feature-status.md`, `websocket-api.md`, `architecture.md`, `authentication.md`, `index.md`, `extending-kandev.md`, `coverage.json`)
- [x] Spec updated (archived with a pointer to the plugin)
- [x] Migration is idempotent and replay-safe, with tests
- [x] No breaking change to unrelated surfaces

## Not done here

No live Slack workspace has exercised the plugin end to end. Its behaviour is covered by unit and end-to-end tests against a stub Slack and a fake Host, and against Slack's real auth endpoint rejecting invalid tokens — but a real mention-to-task round-trip is untested.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2344?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jcfs:
Thanks both — all three findings were valid and are fixed in 386f7ea.

**Blocker — Postgres coverage for `secretsTableExists`:** agreed, and it was the right thing to block on. Added `slack_retirement_postgres_test.go` covering the populated case, the fresh-database case (absent `secrets` table, which is the normal fresh-boot ordering), and replay. I ran them against a real Postgres 16 rather than leaving them to skip, then broke the `information_schema` lookup and confirmed the suite fails — so the coverage is real, not decorative.

**Suggestion — `id = name` masking the filtered column:** correct, and a sharper catch than it first looks. Both suites now seed divergent values (`id = "slack:ws-1:token"`, `name = "Slack token"`, as `secretadapter.Set` actually stores them) plus a decoy row whose *label* starts with `slack:` but whose key does not, asserted to survive.

**Suggestion — stale archive banner:** correct, the banner predated the Socket Mode redesign. Rewritten to state the plugin's real contract (Socket Mode, `@Kandev` / `/kandev`) and to mark everything below it as the historical in-tree record. `!kandev` and polling now apply only to the browser-session fallback.

---

**Declining @coderabbitai's first comment** ("delete by `name`, not `id`"), with evidence — the premise is inverted:

- `secretadapter.Set(ctx, id, name, value)` stores `secrets.Secret{ID: id, Name: name}` (`internal/integrations/secretadapter/secretadapter.go:47`).
- The retired code called `secrets.Set(ctx, SecretKeyForToken(ws), "Slack token", …)` where `SecretKeyForToken` returns `"slack:<ws>:token"`.

So `id` is the vault key and `name` is the human label. `WHERE name LIKE 'slack:%'` would match **zero** rows in production — the label is `"Slack token"`, which has no colon — leaving exactly the credentials this migration exists to delete on disk indefinitely. Same namespacing convention as `pluginSecretID` in `internal/plugins/config.go:57`.

The second half of that comment was still useful and is applied: the test seeding did need to prove which column is filtered. I verified the direction by switching the query to `name` and confirming both the SQLite and Postgres tests now fail — that regression is caught either way now.

PR #2389, opened 2026-08-07
refactor(make): report the real dev ports and document every override

JnManso:
`make dev` announced "auto ports" even when PORT or WEB_PORT pinned them, and `make help` documented only PORT. Both were raised by CodeRabbit on #2368 and went in unaddressed.

## Important Changes

- `DEV_PORT_FLAG` was a byte-for-byte copy of `SERVICE_PORT_FLAG`; both are now `PORT_FLAG`, used by the `dev` and `service-install` targets alike.
- The banner has to test whether any flags were supplied, and `$(if …)` does not strip its condition — an all-whitespace `DEV_FLAGS` reads as true and prints `(  )` instead of `(auto ports)`. Stripping `DEV_FLAGS` itself is not the fix either: that collapses whitespace inside a quoted `DEV_ARGS` before the CLI sees it. So `DEV_FLAGS` is forwarded verbatim and a separate stripped `DEV_FLAGS_DISPLAY` drives both the condition and the banner.

## Validation

Ran for real in both PowerShell and Git Bash, against a checkout with an installed Kandev already serving on 38429 and `KANDEV_BACKEND_PORT=38429` exported:

- `make dev PORT=38430 WEB_PORT=37430` → `Launching via CLI (--port 38430 --web-internal-port 37430)...`, backend ready on 38430, SPA renders with no console errors.
- `make dev PORT=38431 WEB_PORT=37431 DEV_ARGS=--verbose` → all three reported and forwarded, backend ready on 38431, SPA renders.
- `make dev` → `Launching via CLI (auto ports)...`, port selection unchanged.
- `make -n dev DEV_ARGS='--label "alpha  beta"'` → the recipe keeps both spaces; only the banner collapses them.
- `make -n service-install PORT=3000 HOME_DIR=/tmp/x` → `service install --port 3000 --home-dir "/tmp/x"`, unchanged by the `PORT_FLAG` rename; likewise with no variables.
- `make help` lists PORT, WEB_PORT and DEV_ARGS.

The banner is the one line where a dry run is not enough: `$(if …)` expands parentheses into an `@echo`, which cmd.exe could have parsed. It survives both shells.

## Possible Improvements

Low risk: `make help` output and a launch banner. No change to what the dev target actually runs.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,11 @@ SERVICE_HOME_DIR_FLAG := $(if $(HOME_DIR),--home-dir "$(HOME_DIR)",)
 SERVICE_NO_BOOT_START_FLAG := $(if $(filter 1 true yes,$(NO_BOOT_START)),--no-boot-start,)
 SERVICE_INSTALL_FLAGS := $(PORT_FLAG) $(SERVICE_HOME_DIR_FLAG) $(SERVICE_NO_BOOT_START_FLAG)
 DEV_WEB_PORT_FLAG := $(if $(WEB_PORT),--web-internal-port $(WEB_PORT),)
-DEV_FLAGS := $(strip $(PORT_FLAG) $(DEV_WEB_PORT_FLAG) $(DEV_ARGS))
+DEV_FLAGS := $(PORT_FLAG) $(DEV_WEB_PORT_FLAG) $(DEV_ARGS)
+# $(if …) does not strip its condition, so an all-whitespace DEV_FLAGS reads as
+# true. Test and print this instead, and forward DEV_FLAGS itself — stripping
+# what reaches the CLI would collapse whitespace inside a quoted DEV_ARGS value.
+DEV_FLAGS_DISPLAY := $(strip $(DEV_FLAGS))
 DESKTOP_BUNDLES ?= dmg
 
 # Phase headers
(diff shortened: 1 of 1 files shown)


JnManso:
@coderabbitai Good catch on `Makefile:61` — applied in 29b772a31. I measured it first, because the two review bots disagreed on whether the `$(strip …)` belonged there at all:

|  | `$(if …)` condition | value |
|---|---|---|
| unstripped, no variables set | TRUE | `[  ]` |
| stripped, no variables set | FALSE | `[]` |
| stripped, `DEV_ARGS='--label "a  b"'` | TRUE | `[--label "a b"]` |

Both halves hold. `$(if …)` does not strip its condition, so the guard is required for `(auto ports)` to print at all — without it an all-whitespace value reads as true and the banner shows `(  )`. And stripping what reaches the CLI does eat whitespace inside a quoted value, exactly as you flagged.

So the split you proposed is the right shape: `DEV_FLAGS` is forwarded verbatim, `DEV_FLAGS_DISPLAY` drives the condition and the banner.

Verified with `make -n dev DEV_ARGS='--label "alpha  beta"'` in both PowerShell and Git Bash — the recipe keeps both spaces, only the banner collapses them — plus a real `make dev PORT=38430 WEB_PORT=37430 DEV_ARGS=--verbose` run that reached `backend ready` with the SPA rendering.

JnManso:
On the `SERVICE_INSTALL_FLAGS` suggestion — declining it, with the measurement that decided it.

The reasoning for `$(strip …)` on `DEV_FLAGS` is right, and I confirmed it on a scratch Makefile: `$(if …)` does not strip its condition, so an all-whitespace value reads as true and the banner would print `(  )` instead of `(auto ports)`.

But the same measurement shows `$(strip …)` also collapses whitespace *inside* a quoted value — which is what CodeRabbit flagged on `DEV_FLAGS`, and what 29b772a31 fixes by forwarding `DEV_FLAGS` verbatim and printing a separate `DEV_FLAGS_DISPLAY`. Applying `$(strip …)` to `SERVICE_INSTALL_FLAGS` would move that bug into `HOME_DIR`: a path like `/my  project` would reach `service install` as `/my project`.

`SERVICE_INSTALL_FLAGS` is never tested by an `$(if …)`, so the guard buys nothing there while costing that. Leaving it as-is. If a conditional is ever added around it, the fix is the same raw/display split now used for `DEV_FLAGS`.

carlosflorencio:
@claude review

PR #2390, opened 2026-08-07
fix(terminal): tell the client when a terminal will never reconnect

JnManso:
## Problem

#2325 fixed the server side of a reconnect loop: a shell terminal left open on
a session that had ended made the backend build and tear down a runtime every
five seconds. The client half was left alone, and it is the half users see.

A terminal on an ended session is refused with `503` on the WebSocket upgrade.
A failed handshake reaches the browser as a bare `1006` close — no status
code, no body, no reason. That is exactly what a transient failure looks like
too, so `ws-reconnect.ts` cannot tell "this will never work" from "agentctl is
still booting". It backs off to a flat 5s and retries for as long as the tab
stays open, and the pane shows nothing at all: no error, no explanation, just a
terminal that stopped responding.

## Fix

**Server** — when `GetOrEnsureExecutionForEnvironment` returns
`ErrSessionTerminal`, complete the handshake and then close with `4001`
(`closeCodeSessionTerminal`) and a short reason, instead of rejecting the
upgrade. Completing the handshake is the point: a close code is the only
channel the browser gives us for saying *why*.

Other failures keep their `503`. They are transient by nature, and the client
should keep retrying them.

**Client** — `ws-reconnect.ts` stops the loop on a permanent close code rather
than scheduling the next attempt, and hands the reason to a new optional
`onPermanentClose`. `use-passthrough-terminal.ts` writes it into the xterm
pane, so the terminal says what happened instead of going quiet.

The close code is a small shared contract between the two sides, so both
constants carry a comment pointing at the other.

## Testing

**Backend** — `TestEnvironmentTerminalClosesPermanentlyForTerminalSession`
dials the real handler over `httptest`, asserts the handshake *succeeds*
(regressing to a 503 fails the test on the dial), and that the close frame
carries `4001` with the expected reason.

**Client** — two cases, and the second is the control for the first:

- a `4001` close stops the loop: one connect attempt, still one after 60s of
  simulated time, and `onPermanentClose` receives the reason
- a `1006` close still retries: a second connect after the 300ms backoff, and
  `onPermanentClose` untouched

Without the pairing, the first test would also pass if the loop were broken
outright.

`./internal/gateway/websocket/` passes, the web suite passes, and eslint,
`tsc --noEmit` and `pnpm i18n:check` are clean.

## Notes

The user-facing string goes through `t()` with a new `task:terminalSessionEnded`
key; `pt-pt` and `zh-cn` fall back to `en` as they do for the other ~650 keys
they do not carry yet.

Scoped to the environment shell route, which is where the reported loop was.
The agent passthrough route has its own readiness path and is left alone.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2390?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/gateway/websocket/terminal_handler.go
+++ b/apps/backend/internal/gateway/websocket/terminal_handler.go
@@ -50,6 +50,17 @@ var terminalUpgrader = gorillaws.Upgrader{
 	CheckOrigin:     checkWebSocketOrigin,
 }
 
+const (
+	// closeCodeSessionTerminal tells the client this terminal will never come
+	// back, so it must stop reconnecting. 4000-4999 is the range RFC 6455
+	// reserves for application close codes.
+	closeCodeSessionTerminal = 4001
+	// sessionTerminalCloseReason travels in the close frame. Keep it under the
+	// 123-byte control-frame payload limit, or the frame is rejected and the
+	// client sees an anonymous 1006 again.
+	sessionTerminalCloseReason = "session has ended"
+)
+
 type terminalRoute struct {
 	kind string
 	id   string
@@ -138,6 +149,17 @@ func (h *TerminalHandler) handleAgentTerminalRoute(c *gin.Context, sessionID str
 func (h *TerminalHandler) handleEnvironmentUserShellWS(c *gin.Context, environmentID, terminalID string) {
 	execution, err := h.lifecycleMgr.GetOrEnsureExecutionForEnvironment(c.Request.Context(), environmentID)
 	if err != nil {
+		// A terminal session never gains an execution, so retrying cannot help.
+		// Rejecting the upgrade with 503 does not say that: a failed handshake
+		// reaches the browser as a bare 1006 close with no status and no body,
+		// indistinguishable from "agentctl is still booting". The client backs
+		// off to a flat 5s and then retries for as long as the tab stays open.
+		// Complete the handshake instead, and close with a code that carries
+		// the verdict.
+		if errors.Is(err, lifecycle.ErrSessionTerminal) {
+			h.closeTerminalAsPermanent(c, environmentID, terminalID, err)
+			return
+		}
 		h.logger.Warn("environment terminal execution not ready",
 			zap.String("task_environment_id", environmentID),
 			zap.String("terminal_id", terminalID),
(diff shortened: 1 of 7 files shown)


jcfs:
**Requesting changes. This shouldn't land in any form close to this one.**

**1. It doesn't do the thing it says it does.**

The description's entire user-facing claim is "the terminal says what happened instead of going quiet." It doesn't. `PassthroughTerminal` renders an opaque, full-bleed overlay whenever the socket isn't connected:

```tsx
{!isConnected && <PanelLoadingState className="absolute inset-0 bg-background" label={t("task:connectingTerminal")} />}
```

`apps/web/components/task/passthrough-terminal.tsx:280`

Trace the new path: handshake succeeds → `onopen` → `onConnected` → `4001` arrives → `onSocketClose` calls `onDisconnected()` at `ws-reconnect.ts:129`, *before* the new permanent-close branch → `connectedTargetId = null` → overlay returns → `onPermanentClose` writes the yellow line into the xterm **underneath the opaque overlay** → the loop stops.

Final state for the user: a spinner reading "Connecting terminal…", forever. `task:terminalSessionEnded` is never visible to anyone. Before this PR the user got a permanent "Connecting terminal…" spinner; after it, they get a permanent "Connecting terminal…" spinner that is now actively false, because nothing is connecting anymore. That is the whole delivered change.

**2. The tests are shaped to the implementation, and that's why nobody caught #1.**

The backend test asserts a close code. The client test asserts the loop stops. Both pass. Neither asserts the outcome the PR exists for — that the pane shows the reason. The description makes a point of pairing the `4001` test with a `1006` control so a broken loop can't fake a pass; good instinct, applied one layer too low. The control that mattered was at the UI: "the ended-session message is visible in the pane." It would have failed, and it would have failed for the right reason.

**3. The cost/benefit is upside down.**

#2325 already removed the expensive half — the workspace reconciliation, the activity lease, the runtime create/destroy cycle. What's left is one early-rejected `503` every 5s per open dead tab, plus a `WARN` line. For that residue this PR introduces a cross-stack wire contract — a magic `4001` duplicated in Go and TypeScript, each side carrying a comment begging the reader to go check the other — plus a new public export, a new optional callback threaded through the loop, a new i18n key in two catalogs, and a handshake-then-immediately-close dance in the gateway. That is a protocol-level commitment bought for a log line.

**4. The client already knows.**

`PassthroughTerminal` holds `useSession(sessionId)` and therefore the session state. `computeCanConnect` returns `true` unconditionally for shell mode (`passthrough-terminal.tsx:140`). Gate it on a non-terminal session and render a placeholder — the codebase already has `ArchivedPanelPlaceholder` and a `task:terminalUnavailableNoActiveSession` key doing exactly this shape of job. That kills the loop *and* fixes the blank pane, in one file, with no new protocol and no server change. If a client-side gate is deemed too racy to be authoritative, say that explicitly and defend it, because it's the load-bearing assumption behind choosing the expensive design over the cheap one, and the PR doesn't mention it.

**5. You're hardening a behavior that is probably wrong.**

#2325's own "Not in this PR" section named it: should a user shell be refused on an ended session at all? The worktree is still on disk. Immediately after an agent fails is precisely when someone wants a shell in it. This PR doesn't answer that question — it writes the refusal into a wire-level close code and ships user-facing copy asserting it as final. If the answer turns out to be "no, allow the shell," every line here is deleted plus a translated string that has to be retired from the catalogs. Answer the product question first.

**6. Minor, but the guarantee is weaker than the test suggests.**

`closeTerminalAsPermanent` writes the close frame and then `conn.Close()` via defer, with no write deadline and no lingering read for the peer's echo. Over `httptest` loopback that always works. Through a real proxy the frame can be dropped and the browser sees `1006` — i.e. it silently degrades to exactly the loop this PR removes. Safe direction, but it means the mechanism is best-effort, and the test can't tell you when it isn't working.

**What I'd take instead:** fix the overlay so a dead terminal actually renders a terminal state rather than a lying spinner. That's the only part a user would ever notice, and it needs neither the close code nor the server change. If you want a case for the server half afterward, make it on its own merits — with a UI test that fails when the pane stays blank.

JnManso:
Closing in favour of #2423.

@jcfs — I checked each point before acting on it rather than taking the review at face value, and the central one holds exactly as described. `PassthroughTerminal` renders `PanelLoadingState` with `absolute inset-0 bg-background` whenever `isConnected` is false; `onDisconnected()` runs at `ws-reconnect.ts:129` *before* the permanent-close branch, clearing `connectedTargetId`. So the yellow line went into the xterm underneath an opaque overlay, and the user was left with a "Connecting terminal…" spinner that this PR had made actively false. The delivered change was worse than nothing.

The diagnosis of *why* nobody caught it is the part I've taken most from: the tests asserted the mechanism — close code sent, loop stopped — and never the outcome. I had paired the 4001 test with a 1006 control specifically so a broken loop could not fake a pass, and still applied that instinct a layer below where it mattered.

#2423 keeps none of this. No close code, no cross-stack contract, no server change, no handshake-then-close. It gates `computeCanConnect` on the session the client already holds, as you suggested, and renders the reason in the overlay's slot instead of behind it. The e2e test asserts both that the message is visible and that the spinner is not — the second assertion being the one that would have failed here.

Point 5 is deliberately left open there rather than quietly resolved: whether a shell should be refused on an ended session at all is a product question, and #2423 only makes the existing refusal honest instead of silent.

PR #2407, opened 2026-08-07
fix(backend): detect WSAEADDRINUSE so Windows port-bind retries actually fire

### LINKED ISSUE #2371: [Bug]: on Windows the agentctl instance-port retry never fires — the EADDRINUSE check does not match WSAEADDRINUSE (10048)
### Summary

`allocatePortAndListener` is written to retry on the next port when a bind hits "address already in use", but neither of its two checks matches what Go returns on Windows. The first occupied port in the 41001–41100 range becomes a hard failure instead of a retry, so every agent instance creation fails while any other process holds 41001.

### Affected area

agentctl

### Kandev version

0.85.0 (also on `main` at 828213a)

### Install mode

Local development, with an installed release holding 41001

### Environment

- OS: Windows 11, amd64
- Go toolchain as pinned by the repo's mise config
- Reproduced with a second Kandev instance already bound to 41001

### Kandev configuration

Default instance port range (`AGENTCTL_INSTANCE_PORT_BASE=41001`, `AGENTCTL_INSTANCE_PORT_MAX=41100`).

### Task/session context

N/A — fails at instance creation, for every agent type.

### Steps to reproduce

1. On Windows, have any process listening on 41001 — for example an installed Kandev release while also running a dev checkout.
2. Start an agent instance.
3. Every agent type logs `failed to bind instance port 41001` instead of moving on to 41002.

### Reproducibility

Always

### Clean-state check

Reproduces on clean state

### Expected behavior

The allocator marks 41001 unavailable and retries at 41002, as the surrounding loop intends.

### Actual behavior

The bind error falls past both guards, the loop returns immediately, and the whole 100-port range effectively collapses to its first entry.

### Evidence

```go
// apps/backend/internal/agentctl/server/instance/manager.go:304-315
ln, err := net.Listen("tcp", fmt.Sprintf("%s:%d", m.config.ListenHost(), allocated))
if err != nil {
    if errors.Is(err, syscall.EADDRINUSE) || strings.Contains(err.Error(), "address already in use") {
        m.portAlloc.MarkUnavailable(allocated)
        ...
        continue
    }
    m.portAlloc.Release(allocated)
    return 0, nil, fmt.Errorf("failed to bind instance port %d: %w", allocated, err)
}
```

Measured on Windows 11 with a standalone Go program doing two `net.Listen` calls on the same address:

```
err.Error() = "listen tcp 127.0.0.1:58538: bind: Only one usage of each socket
               address (protocol/network address/port) is normally permitted."

errors.Is(err, syscall.EADDRINUSE)              = false
strings.Contains(err, "address already in use") = false
syscall.Errno actually returned                 = 10048        (WSAEADDRINUSE)
syscall.EADDRINUSE on windows/amd64             = 536870914
```

On Windows the runtime returns `WSAEADDRINUSE` (10048); `syscall.EADDRINUSE` in the Windows `syscall` package is a distinct synthetic value, and the English error text differs from the Unix wording. Both guards miss.

The same test appears at `apps/backend/internal/gateway/websocket/port_tunnel.go:263` and is blind in the same way.

### Suggested fix

A small shared helper — something like `IsAddrInUse(err)` — checking `errors.Is(err, syscall.EADDRINUSE)` plus `errors.Is(err, windows.WSAEADDRINUSE)` under a `//go:build windows` file, and drop the string match entirely. `golang.org/x/sys` is already a direct dependency (`apps/backend/go.mod:39`), so `golang.org/x/sys/windows.WSAEADDRINUSE` needs no new module.

A test that double-binds a real socket on the host platform and asserts the helper returns true would lock this in.

Worth adding `./internal/common/portutil/...` to the `Test windows-sensitive packages` step of the existing `Backend (windows)` job (`.github/workflows/backend-tests.yml`). That job already runs `go build ./...` and `go vet ./...` on `windows-latest`, but neither catches an `errors.Is` that silently returns `false`, and the affected packages are outside its current test scope — which is why this survived.

JnManso:
On Windows the "address already in use" guards in `allocatePortAndListener` and the websocket port tunnel never match, so the first occupied port in the 41001–41100 range is a hard failure instead of a retry and every agent instance fails to start while anything else holds 41001. Both call sites now go through a shared `portutil.IsAddrInUse` that also recognizes the winsock code.

## Important Changes

- `portutil.IsAddrInUse` centralizes the check; the `//go:build windows` variant tests `windows.WSAEADDRINUSE` (10048), which is what a real bind failure carries — `syscall.EADDRINUSE` there is a synthetic 536870914 that no bind ever produces.
- The `strings.Contains(err, "address already in use")` fallback is gone: the Windows wording shares no substring with the Unix one, and both are localized by the OS.
- `./internal/common/portutil/...` joins the `Backend (windows)` job's test scope. `go build` and `go vet` already run there and pass on every platform — neither notices an `errors.Is` that silently returns `false`, which is why this survived.

## Validation

Measured on Windows 11 (amd64) with two `net.Listen` calls on the same address, before the change:

```
err.Error() = "listen tcp 127.0.0.1:58538: bind: Only one usage of each socket
               address (protocol/network address/port) is normally permitted."

errors.Is(err, syscall.EADDRINUSE)              = false
strings.Contains(err, "address already in use") = false
syscall.Errno actually returned                 = 10048
syscall.EADDRINUSE on windows/amd64             = 536870914
```

- `go build -tags fts5 ./...` — pass
- `go vet -tags fts5` on the three touched packages — pass
- `go test -tags fts5 -race ./internal/common/portutil/...` — pass, including the new real-double-bind regression
- `gofmt -d` on every touched file — clean

The two `mcp_validation_test.go` failures visible locally in `internal/agentctl/server/instance` are pre-existing and environmental (`exec: "sh": executable file not found`), unrelated to this change.

## Possible Improvements

Low risk: the helper only widens what counts as "in use", so a missed match degrades to today's behaviour rather than introducing a new failure. `internal/agentctl/server/instance` and `internal/gateway/websocket` were deliberately left out of the Windows CI job — they carry unrelated shell-dependent tests that would redden it.

Closes #2371

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2407?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -377,8 +377,14 @@ jobs:
       # in source with a documented reason. The shell-fixture tests now use
       # a self-hosting Go helper (see testmain_test.go) so they run on
       # every platform.
+      # portutil carries the cross-platform "address already in use" check.
+      # It builds and vets clean everywhere, so only a real double-bind on
+      # Windows tells WSAEADDRINUSE apart from syscall.EADDRINUSE — without
+      # this entry the regression is invisible to CI. Its callers
+      # (agentctl/server/instance, gateway/websocket) only delegate to it and
+      # are deliberately left out: they carry unrelated shell-dependent tests.
       - name: Test windows-sensitive packages
-        run: go test -race -v ./internal/agentctl/server/process/... ./internal/agent/runtime/agentctl/launcher/...
+        run: go test -race -v ./internal/agentctl/server/process/... ./internal/agent/runtime/agentctl/launcher/... ./internal/common/portutil/...
 
       # Keep this targeted: the service package has broad subprocess coverage,
       # while this native test is the Windows path-semantics regression.
--- a/apps/backend/internal/agentctl/server/instance/manager.go
+++ b/apps/backend/internal/agentctl/server/instance/manager.go
@@ -10,13 +10,13 @@ import (
 	"os/exec"
 	"strings"
 	"sync"
-	"syscall"
 	"time"
 
 	"github.com/google/uuid"
 	"github.com/kandev/kandev/internal/agentctl/server/config"
 	"github.com/kandev/kandev/internal/agentctl/server/process"
 	"github.com/kandev/kandev/internal/common/logger"
+	"github.com/kandev/kandev/internal/common/portutil"
 	"github.com/kandev/kandev/pkg/agent"
 	"go.uber.org/zap"
 )
@@ -303,7 +303,7 @@ func (m *Manager) allocatePortAndListener(id string) (int, net.Listener, error) 
 		// all interfaces so Docker/remote executors can reach the instance.
 		ln, err := net.Listen("tcp", fmt.Sprintf("%s:%d", m.config.ListenHost(), allocated))
 		if err != nil {
-			if errors.Is(err, syscall.EADDRINUSE) || strings.Contains(err.Error(), "address already in use") {
+			if portutil.IsAddrInUse(err) {
 				m.portAlloc.MarkUnavailable(allocated)
 				m.logger.Warn("port already in use; retrying",
 					zap.String("instance_id", id),
--- a/apps/backend/internal/common/portutil/addrinuse.go
+++ b/apps/backend/internal/common/portutil/addrinuse.go
@@ -1,0 +1,30 @@
+package portutil
+
+import (
+	"errors"
+	"syscall"
+)
+
+// IsAddrInUse reports whether err is the operating system's "address already
+// in use" bind failure, on every platform kandev builds for.
+//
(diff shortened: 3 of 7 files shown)


JnManso:
Superseded by #2394, which landed the same `IsAddrInUse` consolidation as `netutil` along with the port-preflight and health-token work. Closing.

PR #2414, opened 2026-08-08
fix(workflow): resolve configure_session agent families so per-step models apply

nova28:
Workflow steps that set a per-step model silently had no effect: `configure_session` rules name an agent family the way a person writes it (`Claude`) while a session stores the canonical agent ID (`claude-acp`), so the byte-for-byte comparison in rule selection never matched and every step ran on the task profile's model. Rule selection now resolves both sides against the agent registry, and refuses rather than guesses when a reference could mean more than one installed agent.

The reported symptom was `configure_session` being applied and then clobbered by the profile write during ACP session initialization. It was never applied at all — the two look identical in a log only because the Claude ACP provider default happens to be `opus[1m]`, the same value the workflow asked for.

## Important Changes

- **`Registry.ResolveFamilyIDs(name) []string`** returns the canonical IDs of *every* registered agent answering to a reference — ID, name, or display name, case-insensitively and whitespace-trimmed — sorted. An exact canonical ID is unambiguous by construction and resolves alone. Returning the candidate set rather than a count is what lets a caller distinguish an ambiguity that concerns it from one naming only other agents.
- **Rule selection is classified against the session's own family.** A rule for a known but different agent stays a silent no-op. A reference that is ambiguous *for this session*, one that names no known agent, and two rules resolving to the same family each apply nothing and raise a session warning saying which — a session setting that goes missing without explanation is the other half of the same failure mode.
- **Both refusals are scoped to the session being configured.** A step routinely carries rules for agents this session is not running, so one colliding custom TUI agent cannot switch per-step model selection off for every agent at once.
- **`provideOrchestrator` wires the registry as the resolver.** Every orchestrator test installs a resolver directly into the `Service` struct, so deleting that line leaves the whole suite green while the shipped binary silently reverts to the old comparison. An AST guard test pins it.
- **The runtime session layer now logs `set runtime model on ACP session`,** like the profile layer above it. Without that line a trace shows only the profile write, and a runtime layer that never ran is indistinguishable from one that ran and was overwritten — which is exactly how this was originally misdiagnosed.

Deliberately unchanged: profile application for sessions with no workflow rule, the `processOnEnter` ordering, and the `configure_session` ⊕ step `agent_profile_id` exclusivity in `ValidateStepEvents`.

## Validation

Screenshots do not apply — this is a backend and spec change with no `apps/web/` paths.

Gauntlet at `9541537ef`, rebased onto `6393d2df8`:

| Command | Result |
| --- | --- |
| `make fmt` | clean (only pre-existing gofmt drift in `lifecycle/executor_backend.go`, untouched by this diff, reverted) |
| `make typecheck` | exit 0 |
| `make test` | 213 packages `ok`; 7 failures across 4 packages, all pre-existing — see below |
| `make lint` | `0 issues.` backend · web eslint clean · `All 113 harness file(s) passed.` · architecture clean |
| `make lint-format` | `All matched files use Prettier code style!` |
| `pnpm run i18n:ratchet` | `no UI source added or modified`; `guard allowlist intact — 452 entr(ies)` |

`make test-e2e` was not run, and is not owed: no changed path is under `apps/web/`, so the `apps/web/**` row of the verify impact matrix is never reached.

**Pre-existing test failures, established not assumed.** Six of the seven (`agentctl/server/config`, `agentctl/server/process`, `internal/gitlab`) were already baselined against the merge base in a detached worktree and reproduce there identically. The seventh, `repoclone.TestEnsureWorkspaceClonedWithBasicAuthKeepsCredentialScopedToGitChild`, is a full-suite-load flake: `git diff origin/main...HEAD -- apps/backend/internal/repoclone` is empty, `go list -deps -test ./internal/repoclone/` imports none of the changed packages, and the package passes on this branch when run on its own.

**Regression proof.** `TestProcessOnEnterConfigureSessionMatchesAgentDisplayName` is the primary receipt. Re-derived this round in a throwaway worktree by forcing `classifyRuleAgentFamily` onto its pre-fix verbatim branch — it and six siblings go red:

```
--- FAIL: TestProcessOnEnterConfigureSessionMatchesAgentDisplayName
--- FAIL: TestProcessOnEnterConfigureSessionUnknownAgentFamilyWarns
--- FAIL: TestProcessOnEnterConfigureSessionAmbiguousAgentFamilyWarns (both subtests)
--- FAIL: TestProcessOnEnterConfigureSessionUnrelatedAmbiguousRuleStillApplies
--- FAIL: TestProcessOnEnterConfigureSessionRelevantAmbiguityRefusesDespiteExactRule
--- FAIL: TestProcessOnEnterConfigureSessionConflictingRulesWarn
--- FAIL: TestProcessOnEnterConfigureSessionDistinctFamiliesStillApply
```

`configure_session` coverage went 12 → 14 tests; registry family resolution has 3, including one asserting the whole ambiguous candidate set.

**End-to-end on an isolated instance**, because this cannot be validated from inside a Kandev session whose own model handling is what changed. Throwaway `KANDEV_HOME_DIR`, fresh SQLite, mock agent, binary stamped `-X main.Commit=ed2dac479` — this branch's content before the rebase onto `6393d2df8`, which changed no file in the diff. Profile model `mock-fast`; start step given a two-rule `configure_session` through the real API so `ValidateStepEvents` ran.

*Run A — an ambiguity naming only other agents must not suppress the matching rule:*

```
set profile model on ACP session  {'model': 'mock-fast'}
set runtime model on ACP session  {'model': 'mock-smart'}
sending prompt to agent           {'prompt_length': 4282}
```

```json
"runtime_config":      {"model": "mock-smart"},
"acp_model_state":     {"current_model_id": "mock-smart"},
"acp_config_baseline": {"model": "mock-fast"}
```

The profile write still fires first and is unchanged, so profile application is not regressed.

*Run B — same binary, same workflow, one extra installed agent so the ambiguity now names this session's family:* the model stays at the profile's `mock-fast` and the user is told why — `This step's session settings were not applied because "Mock" matches more than one agent installed here. Name the agent by its exact ID in the workflow rule.`

**Docs.** `configure_session` appears nowhere in `docs/public/**`, so no public-docs change is owed. `docs/specs/workflow-session-settings/spec.md` is updated with the scoping rule and the corrected silence taxonomy.

**Out of scope, noted not fixed:** step-level `agent_profile_id` appears resolved only via `resolveEffectiveAgentProfile` from `startTask`, and `start_agent: true` on `POST /tasks` skips `processOnEnter` entirely, so a directly-launched session receives no step configuration at all. Related, but separate questions.

## Possible Improvements

Low risk — behaviour changes only for steps that carry a `configure_session` action, and degrades to the previous exact comparison when no resolver is wired. The known gap is `KANDEV_MOCK_PROVIDERS` (dev/E2E only, opt-in, never honoured in production): it replaces an agent's display name with `Mock Claude`, so `Claude` resolves to nothing there and the rule warns instead of applying. That mode behaved the same way before this change, and naming the agent by its canonical ID works.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2414?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


nova28:
Pushed `67e4e37d` — three of the four review items actioned, one declined with evidence. Also rebased onto current `main` (`cba3838c8`) and re-ran the affected packages and lint on the new base: `internal/orchestrator`, `internal/agent/registry`, `internal/backendapp` all `ok`, `golangci-lint` `0 issues.`

### Actioned

**Claude review #3 — `resolveSessionAgentFamily`'s ambiguous-session path was untested.** Correct, and it was the one item with real teeth. `TestProcessOnEnterConfigureSessionAmbiguousSessionFamilyRefuses` now covers it: session stored as the alias `"Claude"` instead of a canonical ID, with a colliding custom TUI agent installed, and an unambiguous rule on the other side. Proven load-bearing by deleting the guard — the warning disappears and the step silently falls back to the profile's model:

```
--- FAIL: TestProcessOnEnterConfigureSessionAmbiguousSessionFamilyRefuses
    warnings = []string{}, want one ambiguity warning
```

**Claude review #1 — the ambiguity early-return is order-dependent and only one ordering was tested.** Correct. `…RelevantAmbiguityRefusesDespiteExactRule` is now table-driven over both orderings. Being straight about what it is: the two behave identically today, so this is a guard rather than a regression test — it exists so a later refactor that accumulates matches before checking ambiguity cannot silently make array order decide which model a step runs on.

**CodeRabbit — canonical-ID precedence is case-sensitive but the doc comment implied otherwise.** Reproduces exactly: `r.agents[trimmed]` is an exact map lookup while `agentAnswersToFamily` uses `EqualFold`, so `"claude-acp"` takes the fast path and `"Claude-ACP"` falls through to the alias scan and can come back ambiguous. Comment clarified. Resolution behaviour deliberately unchanged — making the precedence case-insensitive would change which references resolve versus refuse, which is a behaviour change and does not belong in a review fixup.

### Declined

**CodeRabbit — add a regression test asserting the `set runtime model on ACP session` log line.** Not actioning this, for three reasons rather than one:

1. The diff adds a `logger.Info` call inside an `else` that already existed. No new branch, no changed return value. The repo convention asks for tests covering new or changed *logic*; a log statement is observability.
2. No test anywhere in the repo asserts any of these ACP layer log strings — `grep -rn "set profile model\|set runtime model\|set profile mode" internal/ --include=*_test.go` returns nothing. This would be the first, and it would pin a log message string, which is an implementation detail rather than behaviour.
3. `applyProfileSessionLayers` and `applyRuntimeSessionLayers` have no direct tests at all, so this needs new harness for an otherwise-untested function in order to cover one of its two symmetric log lines. If that coverage is wanted it should cover both layers and their failure paths, which is a bigger piece of work than a fixup round should smuggle into a branch that is about to merge.

Happy to do it as a follow-up if a maintainer disagrees.

### Unrelated red X

`label-allowlisted-fork` fails on this PR and on every fork PR. It is not this diff:

```
data: { message: 'Resource not accessible by integration', status: '403' }
x-accepted-github-permissions: 'issues=write; pull_requests=write'
```

`.github/workflows/claude-code-review.yml` declares only `permissions: issues: write` for that job, but adding labels to a pull request needs `issues: write` **and** `pull_requests: write`. It gated nothing here — `claude-review-fork` ran and passed without the labels. Left out of this PR on purpose to keep it to one logical change; worth a one-line fix separately.

carlosflorencio:
Nice!

PR #2380, opened 2026-08-07
ci: label allowlisted fork pull requests

carlosflorencio:
Allowlisted fork contributors currently require separate maintainer label actions before their pull requests can use the review and preview automation. This change makes `CLAUDE_REVIEW_ALLOWLIST` add both approval labels and directly authorize preview deployment while preserving per-commit `safe-to-test` cleanup.

## Important Changes

- Adds `safe-to-review` and `safe-to-test` to newly opened allowlisted fork pull requests through a base-controlled workflow job.
- Extends fork preview authorization to `CLAUDE_REVIEW_ALLOWLIST`; direct gates remain because `GITHUB_TOKEN` label events do not launch another workflow run.
- Adds workflow contract coverage, CI registration, and records the new security boundary in the spec and ADR.

## Validation

- `python3 .github/scripts/claude-code-review-workflow-contract_test.py` — 9 tests passed.
- `python3 .github/scripts/preview-env-workflow-contract_test.py` — 1 test passed.
- `python3 .github/scripts/lint-action-pinning.py` — all 18 workflow files use SHA-pinned action refs.
- `git diff --check` — passed.
- Live GitHub validation with an allowlisted fork PR remains after merge.

## Possible Improvements

Medium risk: the allowlist now authorizes fork preview code with deployment credentials, so its membership must remain limited to trusted contributors.

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have manually tested my changes and they work as expected.
- [ ] My changes have tests that cover the new functionality and edge cases.
- [ ] If my change touches UI files (`apps/web/`), I have added or updated Playwright e2e tests in `apps/web/e2e/` and verified them with `make test-e2e`.
- [ ] I checked whether this affects public docs in `docs/public/**` and updated them or noted why no docs change is needed.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2380?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #2458, opened 2026-08-08
docs(make): name the env var WEB_PORT actually overrides

JnManso:
The `make help` line added in #2389 credited both `PORT` and `WEB_PORT` with beating `KANDEV_BACKEND_PORT`/`KANDEV_PORT`. Only `PORT` does: `WEB_PORT` maps to `--web-internal-port`, which resolves against `KANDEV_WEB_PORT` (`apps/cli/src/args.ts:126`). Someone keeping `KANDEV_WEB_PORT` set in their shell, wondering why `make dev` ignores it, had nowhere to look.

Raised by the review bot on #2389 as a non-blocking suggestion, after that PR had already been approved.

## Validation

- `make help` in both PowerShell and Git Bash prints the corrected line.
- `make -n dev` is unchanged — this touches an `@echo` in the `help` target only.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2458?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ help:
 	@echo "  bootstrap        Install mise tools, workspace deps, and git hooks"
 	@echo "  bootstrap-e2e    Bootstrap plus Playwright browser/system deps"
 	@echo "  dev              Run backend + web via local CLI (auto ports)"
-	@echo "  dev PORT=38430 WEB_PORT=37430   Run dev on fixed ports (beat KANDEV_BACKEND_PORT/KANDEV_PORT)"
+	@echo "  dev PORT=38430 WEB_PORT=37430   PORT beats KANDEV_BACKEND_PORT/KANDEV_PORT, WEB_PORT beats KANDEV_WEB_PORT"
 	@echo "  dev DEV_ARGS='--verbose'        Pass extra flags through to the dev CLI"
 	@echo "  dev-prod-db      Run dev mode against the production db at ~/.kandev"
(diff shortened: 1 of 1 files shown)


JnManso:
Heads-up on an unrelated CI failure this PR surfaced: `label-allowlisted-fork` fails on every run.

```
RequestError [HttpError]: Resource not accessible by integration
  - https://docs.github.com/rest/issues/labels#add-labels-to-an-issue
##[error]Unhandled error: HttpError: Resource not accessible by integration
```

The job is `.github/workflows/claude-code-review.yml:79-97`, added in #2380. It adds `safe-to-review`/`safe-to-test` to fork PRs whose author is in `CLAUDE_REVIEW_ALLOWLIST`. The run log shows the token was granted `Issues: write`, so the declared `permissions:` block is being honoured and the API still refuses the `issues.addLabels` call.

Nothing to do with this diff — it changes one `@echo` in the Makefile — but it should reproduce on any fork PR from an allowlisted author, so it is probably firing on other contributors too rather than just here.

Happy to open a separate issue if that is easier to track.

PR #2494, opened 2026-08-10
fix(backend): guard SSH command output buffers against concurrent access

jcfs:
`runSSHCommandStdin` returns on context cancellation while `session.Run` is still executing, so `x/crypto/ssh`'s stdout/stderr copier goroutines were still writing into `outBuf`/`errBuf` when the function read them — a data race, and one that outlives the call since those copiers keep writing until the remote command actually ends. Found while writing the SSH executor tests in #2493; reported there rather than fixed inline because that PR is scoped test-only.

## Important Changes

- **`syncBuffer` replaces the plain `bytes.Buffer`** in `runSSHCommandStdin`. It deliberately does *not* implement `io.ReaderFrom`: `bytes.Buffer` does, and `io.Copy` prefers that path, which would let the SSH copier reach the underlying buffer without taking the lock. Exposing only `Write` keeps every mutation guarded — that subtlety is why the obvious "embed a Buffer" version would have silently kept the race.

## When this fired

Any cancellation or deadline with a remote command in flight: `sshPrepareTimeout` (10 min) expiring mid-prepare, a cancelled launch, backend shutdown. The visible symptom is a torn read of the diagnostic output we then surface to the user; the deeper one is unsynchronised writes into memory the caller has stopped tracking.

## Validation

- `CGO_ENABLED=1 go test -tags fts5 -race ./internal/agent/runtime/lifecycle/...` — passes, no goroutine leaks.
- `golangci-lint run ./internal/agent/runtime/lifecycle/...` — 0 issues.
- `CGO_ENABLED=1 go build -tags fts5 ./...` — clean.

Verified by mutation: reverting `syncBuffer` back to `bytes.Buffer` makes `TestRunSSHCommandAbandonsARunningCommandOnCancellation` reproduce the original race exactly —

```
WARNING: DATA RACE
  bytes.(*Buffer).String()
  ...lifecycle.runSSHCommandStdin()
  bytes.(*Buffer).grow()
  bytes.(*Buffer).ReadFrom()
  golang.org/x/crypto/ssh.(*Session).stdout.func1()
```

— and restoring it returns to green. Ran `-count=20 -race` on the new tests to confirm the cancellation timing is deterministic rather than lucky.

The regression test holds the remote command open across the cancellation, so it pins two things at once: that the caller is not left waiting on a command that outlived its context, and (under `-race`) that the buffers are safe to read while the copiers run. `TestSyncBufferSnapshotsConcurrentWrites` covers the type directly.

## Possible Improvements

Low risk. `syncBuffer` gives up `bytes.Buffer`'s `ReadFrom` fast path, so `io.Copy` now goes through a 32 KB intermediate buffer — irrelevant for the short command output this function handles (uname, mkdir, sha256, git).

> [!NOTE]
> **Stacked on #2493** (base `feature/cover-ssh-and-sprite-wjt`), because the regression test needs the in-process `newFakeSSHServer` harness that PR introduces. Retarget to `main` once #2493 merges; the diff of this PR alone is the two files shown.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2494?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/agent/runtime/lifecycle/executor_ssh_operations.go
+++ b/apps/backend/internal/agent/runtime/lifecycle/executor_ssh_operations.go
@@ -123,7 +123,9 @@ func runSSHCommandStdin(ctx context.Context, client *ssh.Client, cmd string, std
 	}
 	defer func() { _ = session.Close() }()
 
-	var outBuf, errBuf bytes.Buffer
+	// Synchronized because the cancellation path below reads these while
+	// session.Run is still going — see syncBuffer.
+	var outBuf, errBuf syncBuffer
 	session.Stdout = &outBuf
 	session.Stderr = &errBuf
 	if stdin != nil {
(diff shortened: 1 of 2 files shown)


jcfs:
> [!WARNING]
> **CI has not validated this PR.** `Backend Tests` (and the rest) are gated on `pull_request: branches: [main]`, so targeting `feature/cover-ssh-and-sprite-wjt` triggers none of them. The only checks that ran here are the two review bots — the "no failing checks" state is the suite never executing, not the suite passing.

Do not merge on the strength of the check list. The retarget to `main` after #2493 lands is what first subjects this fix to CI.

Local verification in the meantime, from `apps/backend`:

- `CGO_ENABLED=1 go test -tags fts5 -race ./internal/agent/runtime/lifecycle/...` — green, including `-count=20` on the new tests to confirm the cancellation timing is deterministic rather than lucky.
- `golangci-lint run ./internal/agent/runtime/lifecycle/... --timeout=10m` — 0 issues.
- `CGO_ENABLED=1 go build -tags fts5 ./...` — clean.
- Mutation: reverting `syncBuffer` to `bytes.Buffer` reproduces the original race trace under `-race`; restoring returns to green.

PR #2501, opened 2026-08-10
fix(agentctl): allow git pull --rebase past the flag allowlist

jcfs:
`POST /api/v1/git/pull` with `{"rebase": true}` could never succeed — it answered 200 with `{"success": false, "error": "potentially unsafe flag: --rebase"}` for every caller, because `securityutil.IsKnownSafeGitFlag` did not list `--rebase` and `runGitCommand` rejects any unlisted flag. `--abort` was missing from the same allowlist, so Pull's rebase-conflict recovery left the worktree stuck mid-rebase and `POST /api/v1/git/abort` was broken outright for both `merge` and `rebase`.

## Important Changes

- `IsKnownSafeGitFlag` now has a small exact-match list alongside the existing prefix-matched one. `--rebase` and `--abort` go in the exact list: prefix-matching them would also admit `--rebase-merges`, `--rebase=interactive` (spawns an editor), and `--abort-*`, none of which this codebase issues. `--` moved there too — it was already exempted from prefix matching by a special case, which the split removes.
- No bypass of `runGitCommand`; commands still run through `subproc.NewGitCommand` + `subproc.RunGitClass`. Nothing else in the allowlist changes.

## Validation

```
cd apps/backend && CGO_ENABLED=1 go test -tags fts5 -race -count=1 \
  ./internal/common/securityutil/... ./internal/agentctl/server/api/... ./internal/agentctl/server/process/...
golangci-lint run ./... --new-from-rev=main --timeout=10m   # 0 issues
```

Tests were written first and verified by mutation:

| Mutation | Test that failed |
|---|---|
| production change reverted | `TestHandleGitPull_RebaseReplaysLocalCommitOntoOrigin`: `pull --rebase failed: error="potentially unsafe flag: --rebase"` |
| `--abort` removed from the exact list | `TestHandleGitPull_RebaseConflictAutoAborts`: `.git/rebase-merge still exists — the conflicting rebase was not aborted` |
| `--rebase` moved to the prefix list | `TestIsKnownSafeGitFlagRejectsRebaseAndAbortVariants`: `IsKnownSafeGitFlag("--rebase-merges") = true, want false` |

New coverage:
- `git_pull_rebase_test.go` — two handler-level tests against a real repo with a bare `origin`. One proves `{"rebase": true}` replays the local commit on top of the remote tip (asserts `HEAD^ == origin tip` and no merge commits, so a plain merge cannot pass). The other drives a conflicting rebase and asserts the auto-abort ran: no `.git/rebase-merge`, HEAD back at the pre-pull tip, clean tree.
- `securityutil/git_test.go` — allow cases for `--rebase`/`--abort`, plus rejection cases pinning the variants that must stay refused.

## Possible Improvements

Low risk. The allowlist grew by exactly two exact-match entries and `runGitCommand` is the only caller. The refactor makes every other entry's prefix behaviour unchanged; `--` keeps its exact-only semantics.

Note: this overlaps #2499 (which adds `git_handlers_test.go` with a `newGitAPIFixture` helper) only in that both build a bare-origin fixture in the same package. No shared symbols, so no merge conflict expected; the two fixtures can be folded together in a follow-up once #2499 lands.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2501?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/agentctl/server/api/git_pull_rebase_test.go
+++ b/apps/backend/internal/agentctl/server/api/git_pull_rebase_test.go
@@ -1,0 +1,161 @@
+package api
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/kandev/kandev/internal/agentctl/server/config"
+	"github.com/kandev/kandev/internal/agentctl/server/process"
+	"github.com/kandev/kandev/internal/common/logger"
+)
+
+// postGitPull drives POST /api/v1/git/pull through the real router and decodes
+// the handler's result. Anything other than 200 is fatal: the handler answers
+// operational failures with 200 + Success:false, so a non-200 means the request
+// never reached GitOperator.Pull.
+func postGitPull(t *testing.T, srv *Server, body string) process.GitOperationResult {
+	t.Helper()
+	rec := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodPost, "/api/v1/git/pull", strings.NewReader(body))
+	req.Header.Set("Content-Type", "application/json")
+	srv.Router().ServeHTTP(rec, req)
+	if rec.Code != http.StatusOK {
+		t.Fatalf("POST /api/v1/git/pull status = %d, want 200: %s", rec.Code, rec.Body.String())
+	}
+	var result process.GitOperationResult
+	if err := json.Unmarshal(rec.Body.Bytes(), &result); err != nil {
+		t.Fatalf("decode pull response: %v (body %s)", err, rec.Body.String())
+	}
+	return result
+}
+
+// newPullAPIServer wires the api Server over an existing working copy.
+func newPullAPIServer(t *testing.T, repoDir string) *Server {
+	t.Helper()
+	log, _ := logger.NewLogger(logger.LoggingConfig{Level: "error"})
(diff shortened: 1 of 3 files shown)


jcfs:
Addressed the suggestion in fca003dcd rather than deferring it — the PR body claims `POST /api/v1/git/abort` is fixed, so that claim should be pinned by a test rather than inferred from the allowlist unit test.

Added `TestHandleGitAbort_ClearsInProgressRebase` and `TestHandleGitAbort_ClearsInProgressMerge`. Both leave the repository genuinely stuck (conflicting `git rebase main` / `git merge main`, with a fixture pre-check asserting `.git/rebase-merge` / `.git/MERGE_HEAD` actually exists before the request), then drive the endpoint and assert the in-progress state is gone, HEAD is back at the pre-operation tip, and the tree is clean.

Mutation-checked like the rest: removing `--abort` from the exact list fails both with `abort rebase failed: error="potentially unsafe flag: --abort"` / `abort merge failed: ...`.

Note for anyone reading the checks: the green CodeRabbit status on the first push was a quota skip ("Review limit reached"), not a clean review — this push should trigger a real one.

Issue #717, opened 2026-04-27
Stale approvals showing up after already approving & final agent message


caleb-reyes-op:
All of the following approval results showed up after having already approved everything (each tool call returned a result already), and the only after the agent had finished replying. I didnt show the total, final response for privacy. 

Running Safari, opencode acp, and GPT 5.4.

<img width="1558" height="842" alt="Image" src="https://github.com/user-attachments/assets/2b7ece9a-4980-42e9-bfe4-2e3bfbf32ed8" />

zeval:
Hey Caleb! Thanks for opening this issue, along with the other few. It's valuable to us to have someone who uses Kandev more intensively with OpenCode in particular. I'll be looking into your submitted issues and get back to you, I'l also let you know that we have a [Discord server](https://discord.gg/gWdCPGcFCD) in case you want to chat more in-depth about any particular issue 👍

PR #2525, opened 2026-08-11
feat(mcp): expose live agent permission requests

vojtabiberle:
External MCP clients cannot currently discover or answer live agent permission prompts, leaving
agents blocked whenever no browser operator is present. This adds an authorized, provider-neutral,
generation-safe list/resolve contract without changing auto-approval defaults.

## Important Changes

- Add `list_pending_agent_permissions_kandev` on the external MCP surface. It returns live,
  task-authorized snapshots with exact task/session/request/pending IDs, safe action details,
  immutable option IDs/kinds, and timestamps.
- Add `resolve_agent_permission_kandev`. It accepts only the exact identity tuple and one option
  originally offered by the provider; no command, cancellation, metadata, or free-form option can
  be supplied.
- Introduce a reusable orchestrator service boundary over agentctl live state. Task/session
  authorization runs before runtime access, and all unknown foreign resources collapse to the same
  not-found response.
- Claim a durable transcript audit record before provider delivery using compare-and-swap state.
  Concurrent, replayed, stale, replaced, and already-resolved requests cannot act on a newer prompt.
- Allowlist and bound presentation fields, redact credential-shaped text, and omit environment,
  headers, MCP arguments, option metadata, file contents, and diffs from the public contract.
- Preserve the existing web approval flow by carrying the same request generation identity through
  `permission.respond`; legacy cached messages without that identity cannot bypass strict resolve.

## MCP API

List one task or session:

```json
{"task_id":"task-123","session_id":"session-456"}
```

Resolve exactly one returned option:

```json
{
  "task_id":"task-123",
  "session_id":"session-456",
  "request_id":"request-789",
  "pending_id":"pending-abc",
  "option_id":"allow-once"
}
```

Audit metadata records the actor user, actor kind, source, exact option identity, timestamps, and
result. PAT secrets, token IDs, command text, and provider environment are never written to the
resolution audit.

## Validation

- Changed backend packages: 14 packages passed with `go test -tags fts5`, including agentctl,
  lifecycle, orchestrator, SQLite audit CAS, MCP handlers/server, and authenticated integration.
- Focused `internal/task/service` permission-resolution audit test passed.
- `golangci-lint run ./...` passed with 0 issues.
- Focused web hook/catalog tests passed (16 tests); `pnpm --filter @kandev/web lint`, web
  `typecheck`, `i18n:check`, `i18n:ratchet`, and workspace `format:check` passed.
- Public docs validation passed (1 test; 41 published pages).
- `make build-backend` and `make build-web-e2e build-e2e-plugin-package` passed.
- `pnpm e2e:raw --project=chromium tests/chat/permission-approval.spec.ts` passed (3 tests).
- Desktop/phone External MCP capture passed (1 Playwright test) with synthetic E2E data.
- Broader task-service/task-handler audits compile the changed interfaces; unrelated local
  repository tests fail in this runner because the container root is owned by `nobody`.

## Related Context

- Issues: #717, #1657
- Prior building blocks: #2424, #2445, #2442, #1037
- A refreshed search of open upstream PRs and public exact tool/action names found no equivalent
  external permission list/resolve implementation.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2525?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


carlosflorencio:
@codex review

carlosflorencio:
Thanks for the contribution! I think we can later expand this to also include agents turns that are using the ask question tool, but needs more changes.

I found two additional blockers that do not seem to be covered by the existing review threads:

1. The live permission state still has several competing terminal paths. `ResolvePermission` consumes the request under `permissionMu`, but `RespondToPermission`, `CancelPendingPermissions`, and the `ctx.Done()` path send directly to `ResponseCh` without atomically retiring the same generation first. A resolver can therefore report success while the provider receives cancellation or a legacy response.

   Could all terminal paths use one generation-aware consume operation, roughly:

   ```go
   func (m *Manager) consumePermission(
       requestID, pendingID string,
       response *adapter.PermissionResponse,
       terminalCode string,
   ) error {
       m.permissionMu.Lock()
       defer m.permissionMu.Unlock()

       pending := m.pendingPermissions[pendingID]
       if pending == nil || pending.RequestID != requestID ||
           pending.State != streams.PermissionStatusPending {
           return &PermissionOperationError{Code: streams.PermissionErrorStale}
       }

       pending.State = streams.PermissionStatusResolving
       delete(m.pendingPermissions, pendingID)
       m.addPermissionTombstoneLocked(requestID, terminalCode)
       pending.ResponseCh <- response
       return nil
   }
   ```

   The exact implementation can differ, but resolution, cancellation, replacement, bulk cancellation, and context cancellation should all claim the same state transition. A deterministic resolve-vs-cancel race test would be useful here.

2. `stripURLCredentials` fails open when `url.Parse` rejects malformed input:

   ```go
   parsed, err := url.Parse(parseValue)
   if err != nil {
       return value
   }
   ```

   For example, a malformed destination containing invalid percent escapes can retain `user:password@host`, and the generic redaction expressions do not necessarily catch that shape. Since this projection crosses a public boundary, parse failure should omit the destination instead of returning the original text:

   ```go
   stripped, ok := stripURLCredentials(destination)
   if !ok {
       projected.Redacted = true
       break
   }
   ```

   Please add malformed absolute and schemeless URL cases with credential canaries.

Also resolve all threads when you address all the feedback 💪

nova28:
This PR's work has been folded into #2875 at @carlosflorencio's request, so the two overlapping external-MCP surfaces (question-answering + permission-resolution) land as one consolidated PR instead of two.

What happened, concretely:

- Merged this branch's 4 commits (`95739bf54`, `1575f5f84`, `c5566fb6a`, `7935cca56`) into #2875 via `git merge`, so this history is preserved as ancestors rather than flattened.
- Fixed both P1s from @carlosflorencio's review on top of the merge:
  - **Terminal-path race**: unified `RespondToPermission`, `CancelPendingPermissions`, and the request's own `ctx.Done()` path behind a single generation-aware consume operation under `permissionMu`, so exactly one wins.
  - **Fail-open redaction**: `stripURLCredentials` now fails closed on a `url.Parse` error (redacts instead of passing the raw destination through).
- Fixed two more P1s found in review of the merged code (not originally flagged on this PR): `normalizePermissionActionType` was silently dropping detail fields for every real ACP tool kind, and a `permission_cancelled` event matched by `pending_id` alone could let a delayed cancellation for a reused `pending_id` expire the wrong message.
- Re-verified the remaining ~8 open review threads (coderabbitai / chatgpt-codex-connector) against the current merged code — all were already addressed by this PR's own later "harden" commits (`c5566fb6a`, `7935cca56`), so nothing further was needed there.
- Full validation (backend `-race`, frontend vitest/lint/typecheck/i18n, and e2e for both `clarification.spec.ts` and `permission-approval.spec.ts`) is green on the combined branch.

@vojtabiberle — thank you for the original design and implementation; it carries over essentially unchanged, just consolidated with the other in-flight MCP surface and with the review findings above addressed. A maintainer can close this one as superseded by #2875 whenever convenient.

PR #2526, opened 2026-08-11
feat(backend): count accepted step-completion signals

nova28:
## The problem

Auto-advance on a workflow step only fires when the agent calls `step_complete_kandev`. When a card
sits in a step long after its work is finished, there is today no way to tell which of these
happened:

- the agent never called the tool, or
- the agent called it and the transition didn't run.

Nothing counts the calls, so the two cases are indistinguishable from outside.

We hit exactly this on our own board. One card called `step_complete_kandev` five times over six
minutes before it advanced. Another called it once, then sat in the same step for 31 more minutes
with a clean tree, four commits, and its work verifiably done. Same-looking message log, completely
different failure.

This PR ships the counter for the first half of that question — the one ADR 0015 already named as
the headline "is the explicit-signal flow working" metric, and deferred.

## What it does

- New package `internal/workflow/signalmetrics` publishes
  `workflow_step_completion_signal_received_total` as an `expvar.Map`, labelled by `source` and
  `agent_type`, using the same `k=v;k=v` label convention as the existing `routing_*` metrics in
  `internal/office/scheduler/metrics_vars.go`.
- `handleStepComplete` (`internal/mcp/handlers/handlers.go`) increments it once per **accepted**
  signal, at the durable metadata-bag write rather than after the event publish — so a publish
  failure (which routes the agent's retry through the existing `already_signaled` dedup path) can't
  cause a missed count. Deduped repeat calls are not counted.
- Labelled by `agent_name` (the bounded registry-facing type, e.g. `"claude"`), not `agent_id` (the
  per-install UUID), matching the convention documented at
  `internal/task/service/service_turns.go:552`.
- `AGENTS.md` documents the new `workflow_*` namespace under Observability, next to `routing_*`.

Counters only: no schema change, no API change, no behaviour change. Visible at `/debug/vars` in
dev mode.

## Why the second counter is deliberately missing

ADR 0015 names a pair. `step_completion_signal_fallback_used_total` is **not** published here,
because its only specified increment site — a manual "Mark complete & advance" fallback button —
doesn't exist anywhere yet (`models.StepCompletionSourceManualFallback` has zero call sites; nothing
in `apps/web/` matches). Publishing it now would pin it at zero and make the fallback/received ratio
permanently misreport "0% fallback".

The `source` label already carries that dimension. Once the fallback button ships and calls
`signalmetrics.RecordSignalReceived` with `source="manual_fallback"`, the same ratio is computable
with no second counter and no rework of this map. I left ADR 0015 itself unmodified rather than
editing a decision record unilaterally — happy to update it in this PR if you'd prefer the divergence
recorded there.

## Where this is going

These two PRs (#2520, #2526) are the backend half of one thing: being able to answer *what happened
to this card, and why did it sit there*. Full answer in my comment below.

## Validation

```
$ cd apps/backend && go test -count=1 -tags fts5 ./internal/mcp/handlers/... ./internal/workflow/signalmetrics/...
ok  	github.com/kandev/kandev/internal/mcp/handlers	8.612s
ok  	github.com/kandev/kandev/internal/workflow/signalmetrics	1.121s

$ make lint          # 0 issues
$ make typecheck     # exit 0
$ make lint-format   # all files match Prettier style
$ (cd apps/web && pnpm run i18n:ratchet)   # no UI source touched
```

`make test` has six failures (`agentctl/server/config`, `agentctl/server/process`,
`system/storage/workspaces`) that reproduce byte-for-byte against the merge base `66eb87ac3` in a
scratch worktree, so they are pre-existing and unrelated.

`make test-e2e` not run locally: the diff has no user-visible behaviour (one in-process `expvar`
increment), so no E2E spec can observe it. CI's path gate on `apps/backend/**` still runs it.

## Follow-up spotted, not fixed here

`handleStepComplete`'s dedup check (`LoadPendingStepSignal` then `SetSessionMetadataKey`, both from
#1276) is a read-then-write race that predates this PR: two simultaneous `step_complete_kandev`
calls for the same step can both pass it, causing a double step transition. The atomic primitive to
close it already exists (`SetSessionMetadataKeyIfAbsent`,
`internal/task/repository/sqlite/session.go:1187`). Out of scope for a counters-only PR; worth a
follow-up, and possibly related to the five-calls-before-advancing case above.

## Checklist


nova28:
Fixup pass: all substantive checks are green — Backend Tests (unit + Postgres + windows), Architecture boundaries, Build, all 14 E2E shards, all 6 E2E container shards, Harness lint, pr-title lint. CodeRabbit's automated review passed 5/5 pre-merge checks with no findings.

One CI job is red — `label-allowlisted-fork` (403 "Resource not accessible by integration" adding the `safe-to-review`/`safe-to-test` labels). This is a token-scope limitation on `pull_request_target` fork events unrelated to this PR's diff (no workflow files are touched here) and not something a contributor can fix from this branch — flagging for repo-admin awareness in case the `issues: write` permission isn't actually being honored for fork-originated `pull_request_target` runs.

No open review threads to address (confirmed via `gh api graphql` on `reviewThreads`: `totalCount: 0`). `cubic` is neutral (org hit its monthly review-line quota, unrelated to this PR). `opencode-review-fork` is gated off by repo config for this contributor/event. `claude-review-fork` completed with no findings posted.

Base is unmoved (`origin/main` == this PR's merge base), so no rebase was needed.

carlosflorencio:
@codex review

carlosflorencio:
Now i see these are part of an audit trail feature:

- https://github.com/kdlbs/kandev/pull/2520
- https://github.com/kdlbs/kandev/pull/2520

Would love to get the bigger picture @nova28, what's the idea after having this data on the backend?

nova28:
@carlosflorencio fair criticism on both counts, and thanks for asking rather than just closing them.
Here's the bigger picture, and I've rewritten both descriptions to lead with it instead of with ADR
numbers.

**Context:** we run Kandev as our primary development workflow, not as a demo. One board, ~75 cards,
~1,750 agent turns, real spend. That's given us a set of questions we can't answer from the store
today, and every one of them bottoms out in the same missing data.

**The two questions**

1. **Where does the time and money go inside a card?** Cost events bill over windows that cross
   workflow-step boundaries. Nothing records when a card changed step, so per-step attribution is
   inference: on our instance 47% of spend sits cleanly inside one step, 30% is only "mostly one of
   two steps", and 22% can't be attributed at all. Step-transition rows are the boundary markers
   that fix that. They're also exactly what `GET /api/v1/sessions/:id/workflow/history` was designed
   to return.

2. **Did a step finish, or finish and fail to advance?** This is the one that actually hurts. With
   `auto_advance_requires_signal` on, a card advances when the agent calls `step_complete_kandev`.
   We have a card that called it, then sat in Build for another 31 minutes — `RUNNING` session,
   clean tree, four commits, its step's own outcome gate satisfied. We have another that called it
   five times over six minutes before it moved. The MCP `tool_call` row has no paired result row, so
   from the log **"the call failed" and "the call succeeded and the board didn't move the card" are
   indistinguishable.**

   That's not a telemetry nicety. Anything supervising a board — a person, a dashboard, a scheduler
   — reads a card as "building for two hours" when it finished ninety minutes ago, and WIP counts
   are wrong in the same direction.

**What the two PRs are**

- **#2526** makes the signals countable: how many `step_complete_kandev` calls were accepted, by
  source and agent type. Answers "did the agent signal at all", per agent type — which matters
  because ADR 0015 explicitly bets on instruction-following that varies across Claude, Codex,
  Cursor and OpenCode, and right now that bet is unmeasured.
- **#2520** makes the moves auditable: one row per transition, with the real trigger and the real
  actor. Answers "did the board act on it, and why did this card move".

Together they close the gap between those two, which is where our stuck cards live.

**What we'd build on it, in order**

1. A per-step timeline for a session ("Build 42 min → Review 8 min → PR Fixup 3.1 h") — which is
   the endpoint that already ships, finally having data behind it.
2. A "why did this card move" answer for any transition, agent or human. Right now the only record
   is the chat transcript.
3. Per-step cost attribution in our own reporting, using transitions as the window boundaries.
4. The ADR 0015 health metric — fallback-used over signal-received, per agent type — once the manual
   "Mark complete & advance" button exists. That's why #2526 ships only one of the two counters:
   the other one has no call site yet and would sit at a permanent, misleading zero.

**Two things worth saying plainly**

- Neither PR adds UI, and neither is speculative instrumentation. #2520 in particular is closer to a
  bug fix: the route, the WS action and a typed frontend client all ship today and all return
  `{"history":[]}` on every install, because `CreateStepTransition` has no production caller on
  `main`. That's a public API that has always been wrong, silently.
- I'd rather land these small and separate than as one "audit trail feature", but if you'd prefer
  them stacked or want a different scope split, say so and I'll reshape them.

Also taking the title/description feedback for future PRs: symptom and user impact first, ADR
reference at the bottom as provenance only.

carlosflorencio:
@nova28 gotcha, now this feature makes more sense and is useful 💪. We can later show this audit info on the UI.

Will push some improvements to both PRs.

PR #2536, opened 2026-08-11
fix(launcher): detect wildcard listeners in port availability probe

carlosflorencio:
## Problem

`make dev` failed to bind the default backend port `38429` and did **not** fall back to a
random port, even though the launcher is supposed to. It happened whenever another Kandev
instance (e.g. `make start-debug`) was already running.

The launcher's port-availability check, `canBind` in
`apps/backend/internal/launcher/ports.go`, only bind-probed `127.0.0.1`:

```go
func canBind(port int) bool {
	ln, err := net.Listen("tcp", fmt.Sprintf("127.0.0.1:%d", port))
	...
}
```

A running backend binds the **wildcard** address (`0.0.0.0` and `[::]`). On macOS/BSD a
bind to the specific `127.0.0.1` address succeeds even while a wildcard listener already
holds the port, and Go's `net.Listen` sets `SO_REUSEADDR`, which reinforces the false
success. So `canBind(38429)` returned `true` against the live backend,
`pickAvailablePortExcept` kept the occupied preferred port, and the child backend then
died on the real wildcard bind.

## Root cause

Regression from PR #2411, which moved `make dev` to the native Go launcher. The previous
TypeScript launcher (`apps/cli/src/ports.ts`) ran a **connect** probe on both `127.0.0.1`
and `::1` in addition to the bind probe, and only reported a port free when nothing
answered the connect *and* the bind succeeded. Its own comment called out this exact trap:
"detects ports where a listener is bound with `SO_REUSEADDR` (Node's default on macOS,
bind-only check would falsely succeed)." The Go rewrite dropped the connect probe.

## Fix

Restore the dual connect-and-bind, dual-stack probe in `canBind`:

- Connect-check `127.0.0.1` and `::1` with a bounded 500ms timeout. If either answers,
  the port is busy.
- Otherwise attempt the loopback bind. Success means the port is free.

The bind probe is kept because it catches reservations a connect misses (Windows phantom
reservations, TIME_WAIT). The bounded timeout keeps a silently dropped SYN (WSL2 mirrored
networking to an unbound loopback port) from hanging port selection. The exported
signature is unchanged, so all callers (explicit backend/web preflight and automatic
preferred-then-random selection) are fixed by the single change.

## Testing

- Added `TestCanBindDetectsWildcardListener` (stands up a wildcard listener, asserts
  `canBind` returns `false` and `pickAvailablePortExcept` does not return the occupied
  port) and `TestCanBindAcceptsFreePort`. The wildcard test fails before the change and
  passes after.
- `go test ./internal/launcher/ -race -count=1` passes.
- `gofmt` clean; `golangci-lint run ./internal/launcher/... --new-from-rev=origin/main`
  reports 0 issues.

## Spec / plan

- Spec: `docs/specs/port-collision-safety/spec.md` (amended: added the
  "Port-availability probe detects wildcard listeners" behavior and a
  "Wildcard-listener port availability" scenario group).
- Plan: `docs/plans/launcher-port-wildcard-detection/`.

---
Co-authored by Augment Code

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2536?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/go.mod
+++ b/apps/backend/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/godbus/dbus/v5 v5.1.0
 	github.com/google/uuid v1.6.0
 	github.com/gorilla/websocket v1.5.3
+	github.com/hashicorp/go-hclog v1.6.3
 	github.com/hashicorp/go-plugin v1.8.0
 	github.com/jackc/pgx/v5 v5.9.2
 	github.com/jmoiron/sqlx v1.4.0
@@ -21,6 +22,7 @@ require (
 	github.com/moby/moby/client v0.5.0
 	github.com/nats-io/nats.go v1.31.0
 	github.com/pkg/sftp v1.13.10
+	github.com/robfig/cron/v3 v3.0.1
 	github.com/santhosh-tekuri/jsonschema/v6 v6.0.2
 	github.com/spf13/viper v1.16.0
 	github.com/stretchr/testify v1.11.1
@@ -71,7 +73,6 @@ require (
 	github.com/goccy/go-json v0.10.2 // indirect
 	github.com/golang/protobuf v1.5.4 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.28.0 // indirect
-	github.com/hashicorp/go-hclog v1.6.3 // indirect
 	github.com/hashicorp/hcl v1.0.0 // indirect
 	github.com/hashicorp/yamux v0.1.2 // indirect
 	github.com/invopop/jsonschema v0.13.0 // indirect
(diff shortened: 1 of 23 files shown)


carlosflorencio:
@codex review

carlosflorencio:
@codex review

PR #2568, opened 2026-08-12
fix(test): pin NODE_ENV so React unit tests can render

yattdev:
Local Vitest runs inside the runtime image inherited `NODE_ENV=production`, which made React load its production build and voided every render-based frontend unit test. This pins Vitest workers back to `test`, narrows the fix to the only remaining Vitest package after [#2411](https://github.com/kdlbs/kandev/pull/2411), and makes CI reproduce and loudly guard the broken environment that originally slipped through.

## Important Changes

- Pin `process.env.NODE_ENV = "test"` in `apps/web/vitest.config.ts`, add a preflight in `apps/web/vitest.setup.ts`, and add `apps/web/vitest-environment.test.tsx` so missing `React.act` fails with a named diagnosis instead of a generic `TypeError`.
- Keep `apps/.npmrc` at `production=false` so workspace installs inside the runtime image still resolve devDependencies under inherited `NODE_ENV=production`.
- Export `NODE_ENV=production` in `.github/workflows/frontend-tests.yml` and add `.github/scripts/frontend-tests-workflow-contract_test.py` plus `lint-action-pinning.yml` trigger coverage so CI now catches removal of the Vitest pin, the workflow export, or the `.npmrc` guard.
- Document the behavior in `apps/web/AGENTS.md` and the repo `AGENTS.md`, and update the PR scope to reflect that `apps/cli` stopped using Vitest in [#2411](https://github.com/kdlbs/kandev/pull/2411).
- Root-cause attribution: the environment interaction started when [#303](https://github.com/kdlbs/kandev/pull/303) added `NODE_ENV=production` to the runtime image. That was correct for production; this PR fixes the missing test-environment pin around it.

## Validation

- `cd apps/web && NODE_ENV=production pnpm vitest run components/github/pr-status-chip.test.tsx`
- `cd apps/web && NODE_ENV=production pnpm test`
- `cd apps/cli && NODE_ENV=production pnpm test`
- `cd apps/web && pnpm lint`
- `cd apps/web && pnpm run typecheck`
- `python3 .github/scripts/frontend-tests-workflow-contract_test.py`
- `python3 .github/scripts/lint-action-pinning.py`
- Mutation-checked the workflow contract against deleting or weakening the guarded workflow lines, including `paths` → `paths-ignore`, commented-out step lines, and removed path triggers.

## Possible Improvements

Low risk, but outright deletion of the workflow step that runs the contract test is still ultimately enforced by branch protection rather than something a file-level assertion can prove from inside the deleted step.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2568?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yattdev:
@carlosflorencio — tagging you because this traces back to a line in #303, and you have the most context on the runtime image. To be clear up front: nothing about that PR was wrong, and there is no action needed on it.

#303 added `ENV ... NODE_ENV=production` to the runtime `Dockerfile`, which is the correct setting for a production image. What has changed since is *where* tests get run: dev and agent sessions now run `pnpm vitest` inside that same image, so they inherit `NODE_ENV=production`.

That turns out to be fatal for React tests, through a chain that is invisible unless you go looking for it:

- Vitest only *defaults* the mode (`process.env.NODE_ENV ??= "test"`) and passes the inherited value straight to its worker pool.
- `react/index.js` picks its build from that variable, and `cjs/react.production.js` does not export `act`.
- `@testing-library/react`'s act-compat shim silently falls back to the deprecated `react-dom/test-utils`, whose production build calls `React.act` — and throws `TypeError: React.act is not a function` on the first `render()`.

Net effect: every component and hook test failed before reaching an assertion, while CI stayed green the whole time, because the CI image (`.github/docker/ci-base/Dockerfile`) sets no `NODE_ENV`. It reproduces in two lines:

```
NODE_ENV=production node -e "require('react-dom/test-utils').act(()=>{})"   # TypeError: React.act is not a function
NODE_ENV=test       node -e "typeof require('react').act"                   # function
```

The same variable was also making `pnpm install` report `devDependencies: skipped because NODE_ENV is set to production`, so a fresh worktree in the container got no vitest, eslint, or tsc at all.

**How this is fixed:** the Dockerfile is untouched — the runtime image should keep `NODE_ENV=production`. Instead the pin moves to the layer that owns test configuration: `apps/web/vitest.config.ts` and `apps/cli/vitest.config.ts` set `process.env.NODE_ENV = "test"`, and `apps/.npmrc` sets `production=false` so the workspace install always includes dev dependencies. (The only Dockerfile that installs the workspace is `e2e.Dockerfile`, which wants them; the runtime image ships the Go binary and Vite output and never runs `pnpm install`.)

There is also a guard so this cannot go unnoticed again: a preflight in `vitest.setup.ts` that throws with the actual diagnosis instead of a `TypeError` from inside `node_modules`, plus `vitest-environment` tests in both packages that fail by name if the pin is ever removed.

Would appreciate your eye on the `.npmrc` and image-boundary reasoning in particular — that is the part closest to your original change.

yattdev:
Update: rebased this onto `main` to resolve merge conflicts (this branch was 74 commits behind). That pulled in #2411 ("move make dev to the native Go launcher", @carlosflorencio), which had already deleted `apps/cli/src/**` and `apps/cli/vitest.config.ts` — `apps/cli` no longer uses Vitest at all, so the `apps/cli` half of this fix (and its guard test) is now obsolete and has been dropped. No action needed from anyone; noting it here for visibility since it changes this PR's diff. The `apps/web` fix — the actual bug — is untouched and re-verified green after the merge (see the updated PR description).

carlosflorencio:
Thanks for digging into this. The Vitest pin is a sensible local safeguard, but I think two issues need addressing before merge:

1. `apps/.npmrc` only protects this repository. The production image exports `NODE_ENV=production`, and Kandev passes that environment into agent processes, setup scripts, and terminals. Other Node repositories opened inside Kandev will still skip dev dependencies or select production behavior.

Since the current image runs a Go server with prebuilt frontend assets, I would remove `NODE_ENV` from the global Docker environment, or strip it before launching user processes:

```dockerfile
ENV KANDEV_NO_BROWSER=1 \
    ... \
    HOSTNAME=0.0.0.0
# Do not set NODE_ENV globally.
```

The Vitest pin and possibly `production=false` can remain as repository-level defenses, but they should not be the production-boundary fix.

2. The new regression test does not exercise the problematic condition in CI. CI starts without `NODE_ENV`, so Vitest defaults it to `test`; removing the pin would still leave the test green. CodeRabbit raised this for the old CLI test, but the same gap remains for the web test after the CLI files were removed.

A focused way to cover both changes would be:

```yaml
- name: Install dependencies
  env:
    NODE_ENV: production
  run: pnpm install --frozen-lockfile

- name: Run tests
  env:
    NODE_ENV: production
  run: pnpm --filter @kandev/web test
```

Please also add `apps/.npmrc` to the workflow path filters.

Finally, please drop the root `AGENTS.md` addition. This is a scoped package-manager detail, and the statement that only the Go binary and Vite output are published overlooks the npm CLI shim in `apps/cli`. The scoped web testing note is enough.

carlosflorencio:
Move to ready for review when the pr is ready for my review again 💪

PR #2592, opened 2026-08-12
test(gitlab): scrub ambient GitLab env vars for host-trust tests

yattdev:
Five `internal/gitlab` and `internal/agentctl/server/process` tests fail on any machine with `KANDEV_GITLAB_HOST`/`GITLAB_HOST`/`GITLAB_TOKEN` set in the ambient environment, because the tests read that ambient state instead of pinning it and no CI workflow has ever set those vars to catch it. Both packages' `TestMain` now scrub the three vars so the suite is hermetic regardless of the parent shell, with new coverage and a CI job closing the gap that let this ship unnoticed.

## Important Changes

- Scrub `KANDEV_GITLAB_HOST`/`GITLAB_HOST`/`GITLAB_TOKEN` in `internal/gitlab` and `internal/agentctl/server/process`'s `TestMain`, plus a guard test per package that fails if the scrub is removed.
- Close the coverage gap #1832 left: `trustedEnvironmentTokenHost` env precedence, the self-hosted trusted-origin path end-to-end, and the agentctl `os.Environ()` fallback reaching GitLab remote-host detection — all behavior #1832 shipped but never covered.
- Add a `test_ambient_env` CI job that runs both packages under poisoned `.invalid`-TLD env values, wired into the `test` gate's `needs`/result check, so a regression here is caught by CI instead of only reproducing on a machine configured against a self-hosted GitLab.
- Document the pattern in `apps/backend/AGENTS.md` and fix goleak package-list drift (`internal/gitlab/` was missing).

This closes a **hermeticity gap**, not a regression: #1832 ("feat: complete GitLab integration parity", merged 2026-07-21) introduced all five tests and both error strings, but since CI never sets these vars, the suite was green before this change and stays green after it — no production behavior changes here (`internal/gitlab` and `internal/agentctl/server/process` production files are untouched; `git diff` of the five original test bodies is empty).

## Validation

- `go test ./internal/gitlab/... ./internal/agentctl/server/process/...` — passes both with the ambient self-hosted-GitLab env set and with `env -u GITLAB_HOST -u KANDEV_GITLAB_HOST -u GITLAB_TOKEN`
- `go test -race` on both packages — passes both ways; also verified under `-shuffle` (randomized test order) and `-count=2` (repeat runs) with the env poisoned
- `TMPDIR=/data/tmp-qa go test ./...` from `apps/backend` — full suite (225 packages) exits 0 with the ambient env still set
- `make -C apps/backend lint` — 0 issues
- Reverted each `TestMain` scrub locally and confirmed the corresponding guard test **and** all five original tests fail; restored and re-verified green
- Reproduced the new CI job's exact command/env locally with a scrub reverted — the job goes red; restored — green

## Possible Improvements

Low risk — test-only change (no production file under either package is touched); `internal/agent/credentials/env_provider.go` also reads ambient `GITLAB_TOKEN` but isn't covered by the new CI job — it passes today and is out of scope for this change.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2592?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yattdev:
@carlosflorencio — heads up on this one, not a complaint about #1832.

`go test ./...` is red on a clean `main` checkout whenever `KANDEV_GITLAB_HOST`/`GITLAB_HOST`/`GITLAB_TOKEN` are set in the ambient shell (e.g. a machine configured against a self-hosted GitLab). Five tests fail:

- `internal/gitlab`: `TestResolveGitLabExecutionCredentialsByAuthMethod`, `TestEnvironmentTokenAllowsImmutableStartupHost`, `TestPollerStartRecordsHealthForEveryConfiguredWorkspace`
- `internal/agentctl/server/process`: `TestGitOperatorCreatePR_GitLabUsesGLab`, `TestGitOperatorCreatePRGlabUsesProjectDefaultBeforeExistingLookup`

`git log -S` traces all five, plus both error strings they assert on, to #1832. It's not a regression — nothing since has touched `trustedEnvironmentTokenHost` (`service_config.go`) or the `os.Environ()` fallback in `GitOperator.environmentValue` (`git.go`), and CI has always been green because no workflow sets those vars. The tests just read ambient process env instead of pinning it, so they only fail on a machine that happens to have a self-hosted GitLab configured.

This PR scrubs the three vars in both packages' `TestMain` (tests that need a specific value now set it via `t.Setenv`), adds a guard test per package so the scrub can't silently regress, and adds a `test_ambient_env` CI job that runs both packages under poisoned values so this class of bug is CI-visible going forward.

It also closes a coverage gap `#1832` left open — nothing exercised `trustedEnvironmentTokenHost`'s env precedence or the self-hosted trusted-origin path before this. I'd appreciate you double-checking `TestEnvironmentTokenTrustsSelfHostedOriginOverStartupHost` (new, in `service_config_test.go`) matches what you intended: with `KANDEV_GITLAB_HOST` set to an origin different from the startup host, an environment-auth config for that origin is accepted and one for the startup host is rejected. That's the self-hosted deployment behavior your PR built but never had a test for — want to make sure the new test asserts what you meant rather than what I inferred from the code.

carlosflorencio:
Pushed follow-up commit `5ce2735ff` with the review improvements:

- Hermeticity guard failures now report only the variable name, never an ambient GitLab token value.
- The agentctl process test suite now scrubs and guards `KANDEV_GITLAB_HOST`, legacy `GITLAB_HOST`, and `GITLAB_TOKEN`.
- Removed the duplicate GitLab guard because `env_hermetic_test.go` already provides the safe package-level coverage.
- Corrected `apps/backend/AGENTS.md` to distinguish `TestMain` in `goleak_test.go` from the scrub helper in `env_hermetic_test.go`.

The normal pre-commit and commit-message hooks passed. I did not run the full test suite or linters locally, per the review request; CI can validate the pushed head.

PR #2633, opened 2026-08-13
fix(queue): auto-merge sibling-session messages and full-queue admissions

Fclem:
Auto-merge skipped consecutive queued messages when they came from different sessions of the same sender task (per-session provenance keys blocked the metadata-equivalence fold), and rejected foldable messages outright once the queue hit its per-session cap. Both cases now fold the same way manual merge does, keeping the queue compact at and below capacity. Addresses the "Fix auto message merging" improve task.

## Important Changes

- `messagequeue`: `comparableAutoMetadata` now ignores `sender_session_id`/`sender_session_name` when deciding fold compatibility. Agent entries already must share `sender_task_id` (via `autoMergeAllowed`), so two messages from different sessions of that task — e.g. interleaved sibling reviewer sessions — fold like manual merge treats them; the surviving entry keeps the tail's session attribution.
- `messagequeue`: admission at a full queue (`ErrQueueFull`) now attempts to fold the incoming message directly into the tail via a new repository method `AutoMergeCandidateIntoAbove` (SQLite + memory). The fold is the admission, so a compatible message is accepted instead of rejected with "Queue is full". The attachment-claim (`afterInsert`) path is excluded because it claims staged attachments against a persisted source row the fold never creates.

## Validation

- `go test -tags fts5 ./internal/orchestrator/... ./internal/backendapp/ ./internal/mcp/...` — pass (remaining full-suite failures in `internal/launcher`, `internal/system/updates`, `internal/agentctl/server/*` are pre-existing environment issues, reproduced on a clean tree).
- `golangci-lint run ./...` — 0 issues; `gofmt` clean.
- New unit tests (red→green): repository fold for same-task/different-session agent entries across memory+sqlite+postgres backends, full-queue candidate fold, incompatible/disabled/after-insert rejections; handler-level WS tests for fold-at-capacity and `queue_full` rejection.
- Playwright e2e: "automatic merge accepts a compatible message into a full queue" (`apps/web/e2e/tests/chat/message-queue.spec.ts`) — passed via `run-e2e.sh --host`.
- Verified against the reported production case from backend logs: session `45da903d` on task `94a59d0f` — same-session reviewer chains folded, different-session transitions stayed separate (the bug), and the queue sat full with `queue full` rejections; both behaviors now fixed.

## Possible Improvements

Low risk: a merged entry keeps the tail entry's session attribution (identical to existing manual-merge semantics); attachment-bearing messages at a full queue still reject with `ErrQueueFull`.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2633?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/orchestrator/handlers/queue_handlers_auto_merge_test.go
+++ b/apps/backend/internal/orchestrator/handlers/queue_handlers_auto_merge_test.go
@@ -91,6 +91,51 @@ func TestWsQueueMessage_AutoMergeBlocksLaterAdmissionUntilClaimFinalizes(t *test
 	require.Equal(t, "first\n\nsecond\n\nthird", status.Entries[0].Content)
 }
 
+func TestWsQueueMessage_AutoMergeFoldsAdmissionAtFullQueue(t *testing.T) {
+	handlers, service := setupQueueHandlers(t)
+	// Fill the queue with auto-merge disabled so the session actually reaches
+	// capacity, then enable it: the next compatible message must fold into the
+	// tail instead of surfacing "queue full".
+	service.SetMaxPerSession(1)
+	service.SetAutoMergeEnabled(false)
+	_, err := service.QueueMessage(context.Background(), "session", "task", "first", "", messagequeue.QueuedByUser, false, nil)
+	require.NoError(t, err)
+	service.SetAutoMergeEnabled(true)
+
+	response, err := handlers.wsQueueMessage(context.Background(), createTestMessage(t, ws.ActionMessageQueueAdd, map[string]interface{}{
+		"session_id": "session",
+		"task_id":    "task",
+		"content":    "second",
+	}))
+	require.NoError(t, err)
+	require.Equal(t, ws.MessageTypeResponse, response.Type)
+	var queued messagequeue.QueuedMessage
+	require.NoError(t, json.Unmarshal(response.Payload, &queued))
+	require.Equal(t, "first\n\nsecond", queued.Content)
+	status := service.GetStatus(context.Background(), "session")
+	require.Equal(t, 1, status.Count)
+	require.Equal(t, "first\n\nsecond", status.Entries[0].Content)
+}
+
+func TestWsQueueMessage_AutoMergeAtFullQueueRejectsIncompatible(t *testing.T) {
+	handlers, service := setupQueueHandlers(t)
+	service.SetMaxPerSession(1)
+	service.SetAutoMergeEnabled(false)
+	_, err := service.QueueMessage(context.Background(), "session", "task", "first", "", messagequeue.QueuedByUser, false, nil)
+	require.NoError(t, err)
+	service.SetAutoMergeEnabled(true)
+
+	response, err := handlers.wsQueueMessage(context.Background(), createTestMessage(t, ws.ActionMessageQueueAdd, map[string]interface{}{
+		"session_id": "session",
+		"task_id":    "task",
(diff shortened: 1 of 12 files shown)


Fclem:
Open design question for reviewers: this PR aligns auto-merge with the manual-merge eligibility rules (agent entries merge whenever they share the same `sender_task_id`, regardless of which session of the task sent them). We are not certain that is the behaviour we want in the long run. The two candidates:

1. **Manual-merge semantics (what this PR does):** same sender task is enough to fold, so interleaved messages from sibling sessions (e.g. reviewer sessions) of one task compact into a single queued entry. This matches what a user can already do by hand from the queue panel, and it is what unblocked the reported queue-full case.

2. **Auto-merge's stricter semantics (status quo before this PR):** entries must also have equivalent per-session metadata, so only consecutive messages from the *same* sender session fold. If that is the intended invariant (per-session attribution should never be mixed), the stricter rule should be backported onto the manual merge path instead, in a separate PR, so manual and automatic merging stay consistent either way.

This PR deliberately does not take a position beyond consistency: manual merge currently allows the fold, so auto-merge should not reject what the user can do by hand. Happy to flip the direction (restrict manual merge in a follow-up) if maintainers prefer option 2.

carlosflorencio:
Thanks for raising the design question.

I prefer the stricter rule for automatic merging: messages from different sender sessions should stay separate. Those sessions are different agents and execution contexts, even when they belong to the same task. If we merge R1 and R2 but retain only R1's metadata, R2's content is incorrectly attributed to R1.

I don't think manual and automatic merge need identical rules. Manual merge is an explicit user decision, while automatic merge should preserve provenance. Please keep the session fields in the metadata comparison:

```go
if key == MetadataEntityReferences || key == MetadataContextFiles {
	continue
}
// sender_session_id and sender_session_name remain comparable
```

One additional blocker not covered by the existing inline comments is the Postgres concurrency behavior in `AutoMergeCandidateIntoAbove`. The session mutex is process-local. Two backend processes can read the same tail, both return success, and the second update can overwrite the first accepted message because the `UPDATE` checks only `id` and `session_id`.

The capacity decision and tail merge should ideally be one repository-owned transaction. At minimum, use a read-snapshot/CAS condition and treat zero affected rows as a concurrent change:

```sql
UPDATE queued_messages
SET content = ?, attachments_json = ?, metadata_json = ?
WHERE id = ? AND session_id = ?
  AND content = ?
  AND attachments_json = ?
  AND metadata_json = ?
```

Please also add a deterministic Postgres test using two repository instances to prove that concurrent admissions cannot silently lose content.

Fclem:
Superseded by #2656 (draft). Per the review on this thread, the direction flipped: `sender_session_id` / `sender_session_name` remain comparable, so auto-merge keeps prompts from different sessions of the same task separate (the stricter rule is kept, not backported onto manual merge). The full-queue fold survives in #2656 with a compare-and-swap guard on the tail update plus a deterministic two-instance Postgres concurrency test, so two backend processes sharing a queue cannot both fold successfully and silently lose one accepted message. Closing this PR in favor of that one.

Fclem:
Closing in favor of #2656.

PR #2645, opened 2026-08-13
fix: show shared clarification context (#2641)

### LINKED ISSUE #2641: ask_user_question: top-level `context` parameter is never rendered to the operator
## What happened

The MCP tool `ask_user_question_kandev` accepts a top-level `context` parameter, documented in its schema as *"Optional shared background information to help the user understand why you're asking these questions."* The call succeeds and returns normally — but the string is never rendered anywhere in the Kandev web UI. The operator never sees it.

## Expected behavior

Text passed in `context` should be displayed once, above or around the question cards, as the schema promises. Alternatively, if rendering it is not supported, the parameter should be removed from the tool schema (or explicitly documented as ignored) so agents don't rely on it.

## Diagnosis

Two black-box probes, each planting a unique sentinel token in every field the tool exposes, then asking the operator to type back every token visible on screen.

**Probe 1** was inconclusive by my own error — the prompt text itself listed all the sentinel names, so the operator saw them regardless of which field rendered.

**Probe 2** was clean: each field contained only its own sentinel, and no field named another's. The operator typed back: `KRYPTON, TOPAZ-5, ALL-CAPS, IRIS-3, WORD, UI`.

| Field | Sentinel | Visible to operator |
|---|---|---|
| `context` (first line) | `COMET-7` | ✗ **no** |
| `context` (last line) | `PELICAN-9` | ✗ **no** |
| question `title` | `KRYPTON` | ✓ yes |
| question `prompt` | `IRIS-3` | ✓ yes |
| option `description` | `TOPAZ-5` | ✓ yes |

Both sentinels from `context` were absent; every other field rendered. Because sentinels bracketed the start *and* end of the context string, this rules out truncation — the field is dropped entirely, not clipped.

## Steps to reproduce

1. From an agent session, call `ask_user_question_kandev` with a distinctive `context` string, and with question prompts that do **not** repeat any of that text.
2. Answer the question from the Kandev web UI.
3. The context string appears nowhere on screen.

## Impact

This is a silent contract violation. The schema actively instructs agents to put shared background in `context`, so agents front-load their rationale there and write question prompts that assume it was delivered. The operator sees questions stripped of the reasoning behind them, and the agent gets no error, no warning, and no way to detect the loss.

Observed in the wild before this diagnosis: an agent asked the operator a question missing context it believed it had shared, and after several failed attempts concluded *"The operator's UI evidently doesn't show the question context."* The agent had to infer the bug from the operator's confusion.

## Environment

- Surface: Kandev web UI (operator confirmed via the probe's second question)
- Reproduced 2026-08-13, across two separate tool calls in a single agent session
- Tool: `ask_user_question_kandev` (kandev MCP server)

## Suggested fix

Render `context` once above the question card stack. If that's not desired, drop it from the tool schema so agents inline background into each `prompt` instead — the workaround agents must use today, once they know.

carlosflorencio:
Operators could not see the shared background supplied with pending clarification questions, which made context-dependent questions incomplete. The existing context now renders once above the clarification carousel and remains visible on desktop and phone widths.

## Validation

- `cd apps/web && pnpm e2e:run tests/chat/clarification.spec.ts -- --grep "shared context"` (1 passed)
- `cd apps/web && pnpm e2e:run --project mobile-chrome tests/chat/mobile-clarification.spec.ts -- --grep "shared context"` (1 passed)
- `cd apps/web && pnpm run typecheck`
- Changed-file ESLint with `--max-warnings 0`
- `cd apps/web && pnpm run i18n:ratchet`
- `cd apps/web && pnpm run i18n:check` (pass; existing advisory catalog warnings)
- `git diff --check`
- Disposable production-build screenshot capture passed; desktop and mobile assets were visually inspected and compressed for PR media.
- Public docs are unaffected because this changes the rendering of existing clarification metadata only.

Closes #2641

## Screenshots

Desktop clarification context
Mobile clarification context

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2645?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/web/components/task/chat/clarification-input-overlay.tsx
+++ b/apps/web/components/task/chat/clarification-input-overlay.tsx
@@ -597,7 +597,7 @@ export function ClarificationInputOverlay({
       {sharedContext && (
         <div
           data-testid="clarification-context"
-          className="mx-4 mb-2 break-words rounded-md border border-border/70 bg-muted/30 px-3 py-2 text-sm text-muted-foreground whitespace-pre-wrap"
+          className="mx-4 mt-3 mb-2 break-words whitespace-pre-wrap"
         >
           {sharedContext}
         </div>
--- a/apps/web/components/task/chat/clarification-input-overlay.tsx
+++ b/apps/web/components/task/chat/clarification-input-overlay.tsx
@@ -58,6 +58,11 @@ function sortMessagesByQuestionIndex(messages: Message[]): Message[] {
   });
 }
 
+function readSharedContext(message: Message | undefined): string | null {
+  const context = (message?.metadata as ClarificationRequestMetadata | undefined)?.context;
+  return context?.trim() ? context : null;
+}
+
 function isQuestionAnsweredAt(
   messages: readonly Message[],
   answers: Record<string, ClarificationAnswer>,
@@ -537,6 +542,7 @@ export function ClarificationInputOverlay({
   // messages or shrunk bundles never put us out of range.
   const total = sortedMessages.length;
   const activeIndex = total === 0 ? 0 : Math.min(rawActiveIndex, total - 1);
+  const sharedContext = readSharedContext(sortedMessages[0]);
 
   useResolveCallback(group.submitState, onResolved);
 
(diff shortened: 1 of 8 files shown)


carlosflorencio:
Addressed the two valid minor review comments in `1623d7947`: clarification locators now scope through the visible active chat panel, and the task verification commands run in independent subshells. The focused desktop and mobile shared-context E2E checks passed again against fresh production builds; commit hooks also passed.

PR #2674, opened 2026-08-14
fix(orchestrator): keep Send Now replacement from dying on cancel

GodricTM:
## Summary
Send Now is supposed to cancel the live turn and immediately prompt with the queued message. Two bugs prevented that:

1. **First Send Now died empty.** A late stream-complete from the cancelled turn called `completeTurnForTaskSession`, which unconditionally cleared the accepted queued-dispatch marker. The replacement `promptTask` then failed with `queued dispatch superseded by a newer one for this session`. The queue entry was restored and the UI showed "The agent finished without producing any output."
2. **Later Send Now conflicted.** After the first replacement owned execution, the accepted marker stayed set for the whole turn (`executor.Prompt` blocks until the turn ends). A second Send Now hit `ErrSendNowConflict` and toasted "Failed to send the queued message now."

## Changes
- Stream-only predecessor completion no longer wipes an accepted Send Now/FIFO successor, and will not close a bound successor turn.
- The ready-path `completeTurnForSession` still clears the marker when the successor turn itself settles.
- After a Send Now replacement claims execution, mark it `live` so a later Send Now can cancel-and-replace that turn.
- FIFO handoff that is still in the accepted (not live) phase continues to conflict, so two dispatches cannot race in that window.

## Test plan
- [x] `go test ./internal/orchestrator -run TestStreamComplete|TestLiveSendNow|TestSendQueuedNow|TestExecuteSendNow|TestSelectSendNow|TestSendNowWorkers|TestExplicitCancellation`
- [x] Live repro: queue a message while an agent is working, click Send Now — replacement prompt is sent instead of empty-output stop
- [ ] After rebuild: queue another message during that replacement turn and click Send Now again — should cancel and deliver the new prompt, not toast send_now_conflict

## Notes
Manually verified on Windows against a local `runtime-0.87.1` build. No UI files changed, so no Playwright e2e updates.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2674?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


carlosflorencio:
Thanks for the contribution. I pushed follow-up commit `1bb4ab9bf9` with these improvements:

- keeps Send Now ownership under the per-session guard through the RUNNING/turn claim before marking it live;
- makes terminal completion/error handling generation-aware, so successor events settle their own turn and delayed predecessor events cannot clear it;
- guards dispatch phase transitions and reports completion-reconciliation exhaustion;
- adds barrier-based race coverage, bound successor-turn assertions, and cleanup-safe prompt gates.

PR #2697, opened 2026-08-15
fix(orchestrator): route workflow steps to profile sessions (#2692)

### LINKED ISSUE #2692: Workflow transitions reuse the previous session instead of applying the step agent profile
### Summary

A workflow can define different agent profiles per step, but when a task moves from one step to another, Kandev reuses the existing session and sends the next step prompt to it. The session keeps the original agent profile/model.

Example workflow:

```text
Backlog / Definition session: Codex GPT-5.6-Luna
Explore step profile: Codex GPT-5.6-Sol
```

Observed result: after moving the task to Explore, the task remains on the Luna session instead of starting a Sol session.

### Reproduction

1. Create a workflow with step A assigned to agent profile A/model A.
2. Create a later step B assigned to agent profile B/model B.
3. Enable `auto_start_agent` on step B and enter B through a workflow transition.
4. Complete/advance from step A to step B.
5. Observe that the existing session continues with profile A/model A while receiving step B's prompt.

Observed task data included:

```text
workflow_step_id: Explore
step agent_profile_id: Codex/Sol profile
primary session agent_profile_id: Codex/Luna profile
runtime model: gpt-5.6-luna
session origin: task_initial
```

### Code path

`apps/backend/internal/orchestrator/workflow_callbacks.go` calls `LaunchSession` with the existing `SessionID` and the target `WorkflowStepID`.

`apps/backend/internal/orchestrator/session_launch.go` routes this to `StartSessionForWorkflowStep`.

`apps/backend/internal/orchestrator/task_operations.go` (`StartSessionForWorkflowStep`) loads the existing session, calls `ensureSessionRunning`, and sends the step prompt. It does not create/relaunch a session using the target step's resolved `agent_profile_id`.

The existing profile-resolution helper (`resolveEffectiveAgentProfile`) is used for launch/start paths, but the workflow-step path does not appear to use it to replace an already-running session.

### Expected behavior

When a workflow step's resolved profile differs from the current session profile, Kandev should either:

1. stop/retire the current session and launch a new session using the target step's profile, or
2. explicitly document/enforce that all steps in a task must use the same agent profile, or
3. expose a workflow option such as `new_session_on_profile_change` / `restart_session_on_step_profile_change`.

The safest default for stage-gated workflows is to create a fresh session when the profile/model changes, preserving the same task/worktree but isolating the stage context.

### Related behavior

The same initial-session reuse can also cause a session created in a passive Backlog step to lack the target step's `step_complete_kandev` tool/configuration. The next step may therefore finish in text without emitting the required explicit completion signal.

### Suggested tests

Add an orchestrator integration test with two workflow steps and two agent profiles. Assert that entering the second step launches a new session with the second profile, or that the old session is explicitly retired before the new one starts. Also assert that the second session receives the second step's MCP/sysprompt tool inventory and completion-signal configuration.

fsmw:
Fixes kdlbs/kandev#2692

## Important Changes

- Routes direct engine-driven workflow `on_enter` actions through a shared
  destination-session preflight.
- Creates or reuses the session for the destination agent profile before
  executing reset, configuration, mode, or auto-start actions.
- Preserves task environment inheritance, primary-session promotion, queue
  transfer, and previous-session retirement.
- Rejects workflow-step prompts when the supplied session profile does not match
  the resolved destination profile.
- Adds regression coverage for direct dispatch, profile switching, session
  continuity, and prompt-boundary mismatches.

## Validation

- `go test ./internal/orchestrator -run
  'Test.*(SwitchWorkflow|WorkflowStepProfile|SwitchSessionForStep|StartSessionForWorkflowStep)' -count=1`
- `go test -tags fts5 ./internal/orchestrator/...`
- `golangci-lint run ./internal/orchestrator/... --timeout=5m`
- Public documentation validation passed; no public docs change was required.
- Manual UAT completed.

## Possible Improvements

Low risk. The change is isolated to orchestrator workflow routing and keeps
passthrough-session behavior unchanged.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2697?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/orchestrator/event_handlers_test.go
+++ b/apps/backend/internal/orchestrator/event_handlers_test.go
@@ -42,6 +42,7 @@ import (
 type mockStepGetter struct {
 	steps                  map[string]*wfmodels.WorkflowStep // stepID -> step
 	workflowAgentProfileID string                            // returned by GetWorkflowMeta
+	workflowAgentProfiles  []string                          // optional profiles returned per call
 	workflowPrompts        map[string]string                 // workflowID -> prompt
 	workflowMetaCalls      int                               // GetWorkflowMeta invocations
 	workflowMetaErr        error                             // optional error from GetWorkflowMeta
@@ -90,6 +91,7 @@ func (m *mockStepGetter) GetPreviousStepByPosition(_ context.Context, workflowID
 func (m *mockStepGetter) GetWorkflowMeta(_ context.Context, workflowID string) (WorkflowMeta, error) {
 	m.workflowMetaMu.Lock()
 	m.workflowMetaCalls++
+	callNumber := m.workflowMetaCalls
 	delay := m.workflowMetaDelay
 	m.workflowMetaMu.Unlock()
 	if delay > 0 {
@@ -102,8 +104,12 @@ func (m *mockStepGetter) GetWorkflowMeta(_ context.Context, workflowID string) (
 	if m.workflowPrompts != nil {
 		prompt = m.workflowPrompts[workflowID]
 	}
+	profileID := m.workflowAgentProfileID
+	if callNumber <= len(m.workflowAgentProfiles) {
+		profileID = m.workflowAgentProfiles[callNumber-1]
+	}
 	return WorkflowMeta{
-		AgentProfileID: m.workflowAgentProfileID,
+		AgentProfileID: profileID,
 		Prompt:         prompt,
 	}, nil
 }
--- a/apps/backend/internal/orchestrator/event_handlers_workflow.go
+++ b/apps/backend/internal/orchestrator/event_handlers_workflow.go
@@ -1721,17 +1721,21 @@ func (s *Service) completeAndStopSession(ctx context.Context, taskID string, ses
 	}
 }
 
-// maybySwitchSessionForProfile checks whether the step requires a different agent profile
-// and switches the session if so. Passthrough sessions are returned unchanged.
-// When the step has no explicit step/workflow profile, the current session is
-// preserved so workflow advancement does not reset the user's current agent mode.
-// Returns the effective session (new or original) and whether processing should continue.
-// A false return means the switch failed; the caller should return immediately.
-func (s *Service) maybySwitchSessionForProfile(
+// prepareWorkflowStepSession resolves the session that must execute a workflow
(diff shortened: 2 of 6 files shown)


carlosflorencio:
Thanks for working on this. One request-context issue remains around the workflow metadata cache.

`prepareDirectWorkflowStepEntry` creates a derived `ctx`, but that value exists only inside the helper. `HandleTrigger` still receives the original context, so an `auto_start_agent` callback creates another cache in `StartSessionForWorkflowStep`.

For a workflow-default profile, a concurrent workflow edit can make preflight route to one profile and make the prompt guard resolve another profile after the source session is retired.

Could the dispatcher seed the cache once and pass the same context through both calls?

```go
ctx = withWorkflowMetaCache(ctx)
sessionID, preloadedState, err = svc.prepareDirectWorkflowStepEntry(ctx, taskID, sessionID)
// ...
_, err = eng.HandleTrigger(ctx, ...)
```

This also makes the existing one-metadata-read-per-entry contract true for direct workflow switches.

I've approved the CI to run 💪

fsmw:
Addressed Carlos' workflow metadata cache feedback in b98b5ad87. The direct dispatcher now creates one request-scoped cache and passes that derived context to both direct-entry preflight and HandleTrigger; downstream defensive seeding remains idempotent. Added regression coverage with a versioned metadata provider proving one read and stable profile resolution. Validation: go test ./internal/orchestrator ./internal/workflow/engine -count=1, gofmt/diff check, and make -C apps/backend lint (0 issues).

fsmw:
Addressed Greptile's replay/idempotency finding in 871e237. The dispatcher now checks the workflow operation ledger before direct on_enter preflight, preventing an already-applied operation from creating/promoting a destination session or retiring the initiating session; the engine's existing idempotency check remains in place before action execution. Added regression coverage for the applied-operation replay. Validation: go test ./internal/orchestrator ./internal/workflow/engine -count=1, gofmt/diff check, and make -C apps/backend lint (0 issues).

PR #2699, opened 2026-08-15
fix: persist Codex MCP approval (#2691)

### LINKED ISSUE #2691: Codex MCP Always allow is not persisted for dynamically injected Kandev MCP
### Summary

When Kandev launches Codex with the task-scoped Kandev MCP server, Codex asks for permission for MCP tools again in every new session. Selecting `Always allow` works for the current Codex session, but the decision is not reused by the next Kandev task/session.

This is especially visible with Codex MCP calls such as `set_task_title_kandev`, `create_task_kandev`, and `move_task_kandev`.

### Reproduction

1. Configure a Codex agent profile in Kandev with `auto_approve: true`.
2. Start a task that receives the dynamically injected `kandev` MCP server.
3. Run a Kandev MCP tool from Codex.
4. Select `Always allow`.
5. Start a different Definition/task session.
6. Run a Kandev MCP tool again.
7. Codex asks for permission again.

The normal Codex command approval setting is already disabled (`--ask-for-approval never`), but this does not suppress the MCP-specific approval prompt.

### Technical findings

The task-scoped server is injected in:

`apps/backend/internal/agentctl/server/api/agent.go` (`injectKandevMcpServers`)

For Codex, the server is converted to CLI config overrides in:

`apps/backend/internal/agent/mcpconfig/passthrough.go` (`CodexStrategy` / `codexServerArgs`)

The generated arguments include the dynamic URL, for example:

```text
-c mcp_servers.kandev.url="http://localhost:<dynamic-port>/mcp"
```

but `codexServerArgs` does not include Codex's per-server MCP approval setting:

```text
mcp_servers.kandev.default_tools_approval_mode="auto"
```

Kandev also uses an isolated Codex home/session directory in the runtime, so a host-level `~/.codex/config.toml` setting does not reliably apply to the Codex process launched by Kandev.

The existing `AGENTCTL_AUTO_APPROVE_PERMISSIONS` / profile `auto_approve` path appears to cover Kandev's ACP-side permission handling, while this prompt is generated by Codex's MCP tool approval policy. Therefore `auto_approve: true` and `--ask-for-approval never` do not currently prevent this specific prompt.

### Suggested fix

For the trusted, task-scoped Kandev MCP server, inject the Codex setting together with the dynamic URL:

```text
-c mcp_servers.kandev.default_tools_approval_mode="auto"
```

Prefer making this explicit and scoped to the Kandev server rather than globally auto-approving every external MCP server. A more general design could add an MCP-server approval-mode field to the internal resolved-server model and have the Codex strategy emit it only when configured.

### Suggested tests

Add a `CodexStrategy` unit test asserting that the Kandev server emits:

```text
-c mcp_servers.kandev.default_tools_approval_mode="auto"
```

and that an unrelated external MCP server does not receive that setting by default. An integration test should launch a fresh Codex task session twice and verify that the Kandev MCP permission prompt is not shown on the second session.

### Environment / verification note

The repository was inspected at commit `39e9bc77f` (`main`). Static code tracing identifies the missing Codex MCP approval override. The Go unit test could not be run in the reporting environment because the Go toolchain was not installed (`go: command not found`).

fsmw:
## Summary

- Persist Codex MCP approval for Kandev's dynamically injected `kandev` server.
- Emit the server-level `default_tools_approval_mode="auto"` override only for Kandev.
- Keep external MCP approval behavior unchanged.

## Validation

- `go test ./internal/agent/mcpconfig ./internal/agent/agents ./internal/agent/runtime/lifecycle`
- `gofmt` and `git diff --check`
- Codex 0.147.0 accepted the override through repeated `-c` arguments.
- UAT completed with two fresh Codex sessions and dynamic MCP ports; no repeated Kandev authorization prompt occurred.
- UAT server shutdown was graceful after the sessions completed.

`golangci-lint` could not load the package in the local read-only runner and reported `no go files to analyze`; this is an environment/tooling limitation.

Fixes #2691

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2699?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/agent/mcpconfig/passthrough.go
+++ b/apps/backend/internal/agent/mcpconfig/passthrough.go
@@ -170,7 +170,7 @@ type CodexStrategy struct{}
 
 const (
 	codexKandevServerName   = "kandev"
-	codexKandevApprovalMode = "auto"
+	codexKandevApprovalMode = "approve"
 )
 
 func (CodexStrategy) BuildPassthroughMCP(servers []types.McpServer, _ PassthroughPaths) (PassthroughArtifacts, error) {
--- a/apps/backend/internal/agent/mcpconfig/passthrough.go
+++ b/apps/backend/internal/agent/mcpconfig/passthrough.go
@@ -168,6 +168,11 @@ func claudeRemoteType(t string) string {
 // inferred from the presence of `command` vs `url` (there is no `type` key).
 type CodexStrategy struct{}
 
+const (
+	codexKandevServerName   = "kandev"
+	codexKandevApprovalMode = "auto"
+)
+
 func (CodexStrategy) BuildPassthroughMCP(servers []types.McpServer, _ PassthroughPaths) (PassthroughArtifacts, error) {
 	args := make([]string, 0, len(servers)*4)
 	for _, srv := range servers {
@@ -214,6 +219,9 @@ func codexServerArgs(srv types.McpServer) ([]string, error) {
 		if len(srv.Headers) > 0 {
 			pairs = append(pairs, pair{"http_headers", srv.Headers})
 		}
+	}
+	if srv.Name == codexKandevServerName {
+		pairs = append(pairs, pair{"default_tools_approval_mode", codexKandevApprovalMode})
 	}
 	// NOTE: env vars and headers are JSON-encoded into `-c` overrides, which
 	// land in the process argument list (visible via `ps aux`) — a local user
--- a/apps/backend/internal/agent/mcpconfig/passthrough_test.go
+++ b/apps/backend/internal/agent/mcpconfig/passthrough_test.go
@@ -140,6 +140,7 @@ func TestCodexStrategy_Args(t *testing.T) {
 	// stdio server: command + args + env
 	for _, want := range []string{
 		`-c mcp_servers.kandev.url="http://localhost:1234/mcp"`,
+		`-c mcp_servers.kandev.default_tools_approval_mode="auto"`,
 		`-c mcp_servers.github.command="npx"`,
 		`-c mcp_servers.github.args=["-y","@mcp/github"]`,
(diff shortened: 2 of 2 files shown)


carlosflorencio:
Thanks for the contribution! I traced this against Codex 0.147 and found one approval-mode detail worth adjusting.

`default_tools_approval_mode="auto"` is already Codex’s default and remains annotation-sensitive, so unannotated Kandev tools can still prompt. Codex only unconditionally skips this gate with `approve` ([approval logic](https://github.com/openai/codex/blob/a95a6fe333c276623ef172f9f7825ac2790be184/codex-rs/core/src/mcp_tool_call.rs#L2214-L2225)).

Could we change the Kandev-specific override to:

```go
const codexKandevApprovalMode = "approve"
```

This should remain independent of the profile’s broad “Auto-approve all permissions” toggle. The goal is to trust only Kandev’s internally injected, task-scoped MCP server while leaving external MCP servers and normal agent permission requests unchanged.

fsmw:
Addressed Carlos' approval-mode feedback in eaa39e5. The Kandev-only Codex override now uses default_tools_approval_mode=approve instead of auto, so internally injected Kandev tools bypass the annotation-sensitive gate while external MCP servers receive no approval override. Updated regression expectations for both dynamic launch URLs. Validation: go test ./internal/agent/... -count=1, gofmt/diff check, and make -C apps/backend lint (0 issues).

carlosflorencio:
I'm re-running the failed tests, should be flaky. Already have a task working on improving it.

PR #2849, opened 2026-08-20
fix: stop update_agent_profile from renaming agents when only the model changes

nova28:
**Today:** calling the `update_agent_profile_kandev` MCP tool (or the underlying `PATCH /api/v1/agent-profiles/:id` / `mcp.update_agent_profile` WS route) with only `model` — no `name` — silently overwrites the agent's name with the raw model id (e.g. "Product Manager" becomes "Sonnet") and shoves the real name into `agent_display_name` instead.
**After this:** `model` and `name` are updated independently; patching one never derives or overwrites the other.
**Who hits this:** any MCP/WS caller of `update_agent_profile` that omits `name`, which is exactly how an agent client would issue a model-only patch. Four Office agents got silently renamed this way in one day before it was caught; a renamed agent also breaks Office config import, which matches profiles by name and creates a duplicate instead of updating the existing one.
**Scope:** standalone — a single deletion of an unconditional name-from-model derive plus its now-orphaned helper, in one file.
**Not here:** `UpdateProfile` has no guard against mutating workspace-scoped Office profiles at all (that's how an instance-level call could touch an Office agent's row in the first place) — pre-existing, unrelated to this diff, and fixing it would be a contract change; noted for a follow-up.

A model-only patch to an agent profile was renaming the agent instead of just changing its model, because the update handler unconditionally derived a new name from the model whenever the request omitted `name`. That derive is now deleted, so updating `model` alone leaves `name` untouched, and vice versa.

## Validation

- `go test ./internal/agent/settings/controller/ -run 'TestUpdateProfile' -v` — added `TestUpdateProfile_ModelOnlyLeavesNameUnchanged` (verified to fail against the pre-fix code, see below) and `TestUpdateProfile_NameOnlyLeavesModelUnchanged`; both pass, along with the two pre-existing `TestUpdateProfile_*` tests.
- Regression proven non-phantom: ran the new tests against the merge-base commit in a scratch worktree — `TestUpdateProfile_ModelOnlyLeavesNameUnchanged` fails there (`response name = "sonnet", want unchanged "Product Manager"`), confirming it actually catches the bug.
- `make -C apps/backend lint` — 0 issues.
- `make -C apps/backend test` (full suite) — only pre-existing, unrelated failures remain (7 tests across `agent/managedruntime`, `agent/settings/controller`'s npm-cache test, `agentctl/server/config`, `system/storage/workspaces`, all `open ...: not a directory` sandbox-filesystem or login-shell PATH issues). Reproduced identically against the merge-base commit in a separate scratch worktree to confirm they predate this change.
- `make fmt`, `make typecheck`, `make lint-format`, `pnpm run i18n:ratchet` (from `apps/web`) — all clean. No UI files touched, so E2E is not required and was not run.
- Independent second-opinion review (Codex, read-only) over the same diff: no defects found.
- Re-ran the full gauntlet after rebasing onto current `main`: identical results.

## Possible Improvements

Low risk: this is a pure deletion (no new code paths) that restores the pre-existing behavior for the one case (`name` omitted) that a prior commit changed; the web UI was never affected since it always sends `name` alongside `model` on every profile save.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2849?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/agent/settings/controller/profile_crud.go
+++ b/apps/backend/internal/agent/settings/controller/profile_crud.go
@@ -179,11 +179,6 @@ func (c *Controller) UpdateProfile(ctx context.Context, req UpdateProfileRequest
 	}
 	if req.Model != nil {
 		profile.Model = *req.Model
-		if req.Name == nil {
-			if newName := c.resolveProfileNameForModel(ctx, profile.AgentID, *req.Model); newName != "" {
-				profile.Name = newName
-			}
-		}
 	}
 	if req.FallbackModel != nil {
 		profile.FallbackModel = strings.TrimSpace(*req.FallbackModel)
@@ -926,31 +921,4 @@ func validateEnvVarValue(ev dto.ProfileEnvVarDTO, i int) error {
 		}
 	}
 	return nil
-}
-
-// resolveProfileNameForModel looks up the agent by ID, fetches its model list (using cache),
-// and returns the display name for the given model ID. Returns empty string on failure.
-func (c *Controller) resolveProfileNameForModel(ctx context.Context, agentID, modelID string) string {
-	agent, err := c.repo.GetAgent(ctx, agentID)
-	if err != nil {
-		return ""
-	}
-	if _, ok := c.agentRegistry.Get(agent.Name); !ok {
-		return ""
-	}
-
-	// Look up the model's display name from the host utility capability
-	// cache. If the cache isn't populated yet (probes not finished, agent
-	// not probed) we fall through to the raw model ID — better than
-	// blocking the save.
-	if c.hostUtility != nil {
-		if caps, ok := c.hostUtility.Get(agent.Name); ok {
-			for _, m := range caps.Models {
-				if m.ID == modelID {
-					return m.Name
-				}
-			}
-		}
-	}
-	return modelID
 }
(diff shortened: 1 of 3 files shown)


nova28:
Both automated-review suggestions are valid and fixed in 0b2d6d543:

- Claude review bot: added a `stored.AgentDisplayName` assertion to `TestUpdateProfile_ModelOnlyLeavesNameUnchanged` (unchanged at `"Test Agent"`, the value `CreateProfile` derives from the agent config).
- CodeRabbit: added a `stored.Name` assertion to `TestUpdateProfile_NameOnlyLeavesModelUnchanged` so the test actually proves persistence, not just the in-memory response.

Both new assertions pass on the current code and both were verified as real gaps before fixing.

carlosflorencio:
Thanks for the contribution. I pushed a test fix that makes the profile fake store snapshots and persist updates. The regression tests now verify both the API response and the stored profile state. This prevents a test from passing when the repository write is skipped.

PR #2871, opened 2026-08-21
fix(scripts): make pr-state work on macOS's stock bash and jq 1.6

nova28:
On a stock macOS toolchain, `scripts/pr-state` silently misreported PR state: unresolved review threads came back as `null` or the script hung and died with no output at all, both of which read as a clean PR to `pr-fixup`'s clean predicate.

## Important Changes

- `"${cursor_args[@]}"` on an empty array aborts review-thread pagination under `set -u` on bash 3.2 (macOS's stock `/bin/bash`, since Apple has frozen it at 3.2.57 over the GPLv3 license and ships no newer version) — switched to `${cursor_args[@]+"${cursor_args[@]}"}`, which is safe on an empty array.
- `has_explicit_blocker_statement`'s prefix strip used `gsub` on an entirely-optional pattern, which matches the empty string at every position and never terminates — switched to `sub`, which is also the semantically correct operator since the pattern is `^`-anchored and only one replacement was ever intended.

## Validation

Verified on this host's actual toolchain (`bash 3.2.57(1)-release`, confirmed via `bash --version`; this is the unmodified stock `/bin/bash` that `#!/usr/bin/env bash` resolves to on macOS):

- `bash scripts/pr-state.test.sh` — 43/43 tests pass, exit 0.
- Reverted the `cursor_args` fix alone: the suite fails immediately with `scripts/pr-state: line 755: cursor_args[@]: unbound variable`, failing the `threads count` test — reproducing the reported RED state.
- Reverted the `sub`/`gsub` fix alone: the suite hangs and times out (exit 124) partway through, producing no further output — reproducing the reported silent-hang behavior (observed on this host's jq 1.7.1, not only jq 1.6).
- `make test-scripts` has two pre-existing failures, unrelated to this change and present identically on `origin/main`:
  - `scripts/dev-prod-db-path.test.sh` fails a Windows-path expectation.
  - `scripts/opencode-code-review.test.sh` fails at "Harness requires trusted producer only for the dedicated OpenCode App"; this file is byte-identical to `origin/main` (`git diff origin/main -- scripts/opencode-code-review.test.sh` is empty).

The repo declares no bash or jq version requirement anywhere (`scripts/doctor` only checks for `pre-commit`), and CI runs ubuntu-latest with bash 5 and a newer jq, so neither bug reproduces there. Four other scripts carry the same latent bash 3.2 empty-array hazard — `dev-isolated`, `kandev-instances`, `kandev-kill`, `kandev-logs` — and are deliberately out of scope here.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2871?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/pr-state
+++ b/scripts/pr-state
@@ -393,7 +393,7 @@ review_evidence_json() {
         markdown_prose_lines
         | any(
             .
-            | gsub("^[[:space:]]*(?:(?:[-*+]\\s+)|(?:[0-9]+\\.\\s+))?(?:\\*\\*|__|_|\\*)?"; "")
+            | sub("^[[:space:]]*(?:(?:[-*+]\\s+)|(?:[0-9]+\\.\\s+))?(?:\\*\\*|__|_|\\*)?"; "")
             | test("^(?:blocked|changes[ _-]?requested|action required|must fix|required fix|please fix)\\b"; "i")
           );
       def review_eligibility:
@@ -752,7 +752,7 @@ fetch_review_threads_nodes_json() {
             }
           }
         }' \
-      -f owner="$owner" -f repo="$repo" -F pr="$pr_number" "${cursor_args[@]}")"; then
+      -f owner="$owner" -f repo="$repo" -F pr="$pr_number" ${cursor_args[@]+"${cursor_args[@]}"})"; then
       rm -f "$nodes_file"
       return 1
     fi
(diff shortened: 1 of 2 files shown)


carlosflorencio:
Thanks for the contribution. We pushed 1090c18 with a portable sleep, kill, and wait watchdog for the jq regression test, plus clearer wording for the conservative gsub guard. The focused bash scripts/pr-state.test.sh suite passes all 43 cases.

PR #2969, opened 2026-08-23
fix(workflow): queue runs for reviewers added before a task reaches Review

nova28:
**Today:** A reviewer added to a task while it's still on an earlier step (e.g. Work) never gets asked to review once the task reaches Review — the quorum guard counts them as a required seat, but the fan-out that actually queues their run can't see them, so zero runs queue and `all_approve` waits forever on a seat nobody was ever asked to fill.
**After this:** The fan-out reads the same participant population the quorum guard already counts — per-task rows at any step, unioned with template rows at the step being evaluated — so a reviewer added at any earlier step gets a run queued once the task reaches a step that needs them.
**Who hits this:** Anyone using workflow steps with participant-based reviewers/approvers (`all_approve` and similar quorum guards) — this is the actual root cause behind "the review loop looks broken" even after #2907 makes the fan-out fire.
**Scope:** standalone fix in `workflow/engine/`; not the same change as #2907 (a 23-file `orchestrator/` rewrite touching the same review loop from a different angle, not touched here).
**Not here:** the same shape of gap for cross-task `participant_role` fan-out (target task ≠ the task the trigger is on) is intentionally left step-scoped — widening it needs a way to resolve the *target* task's own workflow id, which is a contract addition. Tracked as a follow-up, not fixed here.

A reviewer added to a task while it sits on an earlier step gets counted as a required approver once the task reaches Review, but the run that would let them actually approve never queues, so the review step hangs forever on a seat nobody was ever asked to fill. The fan-out now reads the exact participant population the quorum guard already counts, instead of a narrower step-scoped query.

## Important Changes

- Extracted the quorum guard's per-task/template participant read into a shared `gatherParticipantSlate` helper so counting and waking use one query instead of two different ones.
- Added `roleSeatsForFanOut`, reusing the guard's canonicalize/dedupe, so a reviewer seated at multiple steps queues exactly one run.
- Kept the cross-task `participant_role` fan-out step-scoped on purpose (see "Not here" above) — widening it would let a per-task row stamped on a step the target task no longer uses leak into a different workflow's quorum evaluation.

## Validation

- `go test -run 'TestQueueRunForEachParticipantCallback_WakesParticipantAddedOnEarlierStep|TestQueueRunForEachParticipantCallback_ReviewerAtBothStepsQueuesOnce|TestQueueRunCallback_ParticipantRoleWakesParticipantAddedOnEarlierStep|TestQueueRunCallback_ParticipantRoleCrossTaskDoesNotLeakStaleStepRow|TestQueueRunCallback_ParticipantRoleUsesResolvedTargetStep' -v ./internal/workflow/engine/...` — all pass. The headline test reproduces "0 runs queued" against the merge-base commit (proven in a scratch worktree) and is green on this branch.
- `go test -count=1 ./internal/workflow/engine/...` — ok
- `make fmt`, `make lint` (0 issues), `make lint-format`, `cd apps/web && pnpm run i18n:ratchet` — all clean
- `make test` (full backend suite) — green except two failures independently proven pre-existing against the merge-base in a scratch worktree, both unrelated to this diff: `internal/system/storage/workspaces` (macOS `/var/folders` symlink handling) and an intermittent `internal/launcher` flake (a different test fails each run; reproduces identically at the merge base)
- No `apps/web/` files touched by this diff, so no E2E spec is required under CONTRIBUTING's allowlist
- Independent adversarial review: the kandev `code-review` skill plus a cross-vendor `codex` CLI (GPT-5.2) pass, both legs, zero blockers after one fix round for a cross-task scoping regression the first round caught

## Possible Improvements

Low risk — a pure read-path query-scope fix with no schema or contract change. Follow-ups not done here: add fake coverage for the `WorkflowScopedParticipantStore` branch (a pre-existing gap, not introduced by this change), drop the stale "for quorum" wording from two error messages now shared with the fan-out, and give cross-task `participant_role` its own workflow-id resolver so it can be widened safely.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2969?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/workflow/engine/phase2_callbacks.go
+++ b/apps/backend/internal/workflow/engine/phase2_callbacks.go
@@ -102,7 +102,11 @@ func (c QueueRunCallback) resolveTarget(
 		if err != nil {
 			return nil, "", err
 		}
-		agentIDs, err := c.resolveParticipantRole(ctx, stepID, taskID, role)
+		workflowID := ""
+		if taskID == in.State.TaskID {
+			workflowID = in.State.WorkflowID
+		}
+		agentIDs, err := c.resolveParticipantRole(ctx, stepID, taskID, workflowID, role)
 		return agentIDs, stepID, err
 	case strings.HasPrefix(target, TargetAgentProfile):
 		id := strings.TrimPrefix(target, TargetAgentProfile)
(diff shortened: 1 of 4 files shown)


nova28:
Thanks for the flag — I checked this against the actual repository behavior and it doesn't reproduce.

`resolveParticipantRoleStepScoped` calls `c.Participants.ListStepParticipants(ctx, stepID, taskID)`, which is `ParticipantAdapter.ListStepParticipants` → `Repository.ListStepParticipantsForTask(ctx, stepID, taskID)` (`internal/workflow/repository/phase2_sqlite.go:219`).

That repository method already merges template rows (`task_id = ''`) with per-task overrides (`task_id = taskID`) via `mergeParticipantRows` (`phase2_sqlite.go:341-363`): when a per-task row and a template row share the same `(role, agent_profile_id)` key at the queried step, the template row is dropped and only the per-task row survives. The method's own doc comment states this explicitly: "Per-task rows take precedence when both define the same (role, agent_profile_id) pair."

So the exact scenario you described — an agent seated both explicitly via `AddTaskParticipant` at the current step *and* via the workflow template for that step — is already collapsed to a single row before `resolveParticipantRoleStepScoped`'s role filter ever runs. There's no duplicate to dedupe at the call site; the merge already happened one layer down.

I don't see a code change to make here. Resolving as not reproducing.

nova28:
@coderabbitai I looked at the same claim (also raised by the claude-review bot above, which I replied to earlier) and it doesn't reproduce — I verified this empirically, not just by reading.

`ListStepParticipants(ctx, stepID, taskID)` (the two-arg form used by the cross-task `resolveParticipantRoleStepScoped` path) resolves to `Repository.ListStepParticipantsForTask(ctx, stepID, taskID)` (`internal/workflow/repository/phase2_sqlite.go:219`), which already runs its result through `mergeParticipantRows` (`phase2_sqlite.go:341-363`) before returning. That merge drops the template row (`task_id = ''`) whenever a per-task row exists with the same `(role, agent_profile_id)` at the queried step — exactly the scenario in your finding.

I confirmed this with a scratch test against the real repository (not the fakes used in the unit tests you may have been reasoning from): a template row `role=reviewer agent=agent-X` at a step, plus a per-task row for `task-1` with the identical `(role, agent)`, and `ListStepParticipantsForTask` returns exactly **1** row for that agent, not 2. Test removed after confirming (not part of this PR's deliverable — this is pre-existing repository behavior, not something this diff changes).

So `resolveParticipantRoleStepScoped`'s role-filter loop never sees a duplicate row to begin with; there's nothing for `canonicalizeByTaskRoleAgent`/`collapseByRoleAgent` to collapse at that call site. No code change here.

carlosflorencio:
Thanks for the contribution. I pushed c32599aa27 with a small follow-up: shared participant-slate errors now use generic context, and the fan-out callback comment reflects workflow-scoped participant selection. This makes diagnostics and maintenance clearer without changing runtime behavior.

PR #2999, opened 2026-08-24
fix(scriptengine): replace-all the GitHub credential helper

nova28:
> [!TIP]
> **PR walkthrough:** [Open the visual walkthrough](https://walkthrough.kandev.ai/pr/2999/edd697758881.html)

## Problem

On an SSH executor, every task launch after the first fails with `ssh: prepare script failed`.

The prepare script aborts here:

```
warning: credential.https://github.com.helper has multiple values
error: cannot overwrite multiple values with a single value
       Use a regexp, --add or --replace-all to change credential.https://github.com.helper.
```

`GitHubAuthProvider` (`internal/scriptengine/providers.go`) writes the credential helper with a plain
`git config --global credential.https://github.com.helper '...'`. That form **refuses** to overwrite a
multi-valued key and exits 5. The default SSH/Docker/Sprites prepare scripts run under
`set -euo pipefail`, so exit 5 kills the launch.

## Why the key is multi-valued

The provider does it to itself. Both of its branches run `gh auth setup-git`:

- the **no-token** branch emits `gh auth setup-git 2>/dev/null || true`
- the **with-token** branch emits it too, right after the failing `git config`

`gh auth setup-git` always leaves exactly two values — an empty entry that resets the helper chain,
plus gh's own helper:

```
credential.https://github.com.helper=
credential.https://github.com.helper=!/opt/homebrew/bin/gh auth git-credential
```

So the token-less path writes the config the token path cannot overwrite.

## Why this presents as "the SSH executor is broken"

The failure needs a home directory that **persists between runs**. A Docker container gets a fresh
`~/.gitconfig` every time and always sees a clean key. An SSH host keeps the same home forever, so the
two values accumulate and stay.

Local and Worktree executors are unaffected — their default prepare scripts carry only
`{{repository.setup_script}}`, never `{{github.auth_setup}}`.

Measured on a real macOS arm64 SSH host: clear the key by hand → exactly **one** launch succeeds →
that run's own `gh auth setup-git` re-poisons it → every later launch fails.

## Fix

Add `--replace-all` to the with-token write. It collapses any number of existing values into one,
which is the semantics this line always intended.

The no-token branch is deliberately untouched: `gh auth setup-git` is legitimate there, and its
`|| true` means it cannot fail the script.

## Tests

Both added to the existing `TestGitHubAuthProvider` neighbourhood in `providers_test.go`:

1. `credential helper write uses --replace-all` — asserts the emitted command.
2. `TestGitHubAuthCredentialWriteIsIdempotent` — **behavioural**: seeds the two-valued key exactly as
   `gh auth setup-git` leaves it, then runs the emitted command twice against a real `git` binary with a
   temporary `HOME`, and asserts the key collapses to the single intended value.

The second one matters because asserting the flag string alone would not catch a future rewrite that
dropped back to a single-value write by another spelling. Verified red before the fix — it reproduces
git's exit 5 verbatim — and green after.

## Verification note for reviewers

**Test the second launch, not the first.** A cleaned-up gitconfig makes the buggy code pass.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/2999?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/scriptengine/providers.go
+++ b/apps/backend/internal/scriptengine/providers.go
@@ -357,7 +357,16 @@ gh config set git_protocol https --host github.com 2>/dev/null || true`
 		lines := []string{
 			"# GitHub token authentication",
 			"# Configure git credential helper for GitHub HTTPS authentication",
-			`git config --global credential.https://github.com.helper '!/bin/sh -c "echo username=x-access-token; echo password=${GH_TOKEN:-${GITHUB_TOKEN}}"'`,
+			// --replace-all is required, not cosmetic. `gh auth setup-git` (run
+			// below, and by the no-token branch above) leaves the key
+			// multi-valued: an empty entry that resets the helper chain plus
+			// gh's own helper. A plain `git config --global <key> <value>`
+			// refuses to overwrite a multi-valued key and exits 5, which aborts
+			// the whole prepare script under `set -euo pipefail`. That makes the
+			// no-token branch poison the with-token branch: on a host whose home
+			// directory persists between runs (SSH, unlike an ephemeral Docker
+			// container) exactly one launch succeeds and every later one fails.
+			`git config --global --replace-all credential.https://github.com.helper '!/bin/sh -c "echo username=x-access-token; echo password=${GH_TOKEN:-${GITHUB_TOKEN}}"'`,
 			"# Configure gh CLI to use HTTPS protocol",
 			"gh config set git_protocol https --host github.com 2>/dev/null || true",
 			"# Register gh as git credential helper (backup method)",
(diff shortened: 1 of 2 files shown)


carlosflorencio:
Thanks for the contribution. I pushed `edd6977` with the review follow-ups:

- The regression test now executes the complete `github.auth_setup` fragment twice.
- A controlled `gh` executable reproduces the extra helper values from `gh auth setup-git`.
- The test uses a portable `sh` lookup and skips when that shell is unavailable.

The production `--replace-all` fix is unchanged. These tests now cover the full launch path and the persistent multi-value state.

PR #3007, opened 2026-08-24
fix(backend): use the session auth token for non-Docker launch and resume

nova28:
> [!TIP]
> **PR walkthrough:** [Open the visual walkthrough](https://walkthrough.kandev.ai/pr/3007/12a1a6207cb1.html)

An SSH-executor task dies at the first workflow step that resets agent context:

```
failed to resume session: failed remote resume preflight: ssh resume: missing agentctl auth token
```

Reproduced on a real macOS arm64 SSH host on 2026-08-24, walking Triage → Build → Verify fine and
then dying on entry to Review. The same error appears after a backend restart, when startup recovery
tries to resume.

## Root cause

There are two paths that build the executor request, and they disagreed about which token to send.

`prepareWorkspaceExecution` (`internal/agent/runtime/lifecycle/manager_execution.go` ~line 734)
already gets it right — it reads the session's own handshake token and only *overrides* it with the
container-control token for the two Docker executor types:

```go
authToken := m.revealRuntimeSecret(ctx, info.Metadata, MetadataKeyAuthTokenSecret)
if info.ExecutorType == string(models.ExecutorTypeLocalDocker) || ... {
    // override with container control token
}
```

`launchBuildExecutorRequest` (`manager_launch.go` ~line 858) did not. Since #2843 it consumed the
container-control token *unconditionally*:

```go
allowLegacyContainerControlFallback := reqWithWorktree.WorkspaceReuseRequired &&
    (executorType == LocalDocker || executorType == RemoteDocker)
containerControlAuthToken, err := m.revealContainerControlAuthToken(ctx, metadata, allowLegacyContainerControlFallback)
...
AuthToken: containerControlAuthToken,
```

For SSH both branches of `revealContainerControlAuthToken` fall away: there is no
`MetadataKeyContainerControlAuthSecret` (that secret is only minted for containers), and
`allowLegacyFallback` is false because the executor is not Docker. So it returns `"", nil` and the
launch hands agentctl an empty token — every time, not intermittently.

SSH is the only executor whose resume *requires* the token (`requireSSHAgentctlAuthToken`,
`executor_ssh.go` ~line 760), which is why Local and Worktree cards are unaffected and this reads as
"SSH is broken".

The metadata plumbing was never the problem — the key is present the whole time, and only the reader
ignored it. On a resume it reaches the launch request from the session's own `executors_running` row:
`applyRunningRecordToResumeRequest` (`internal/orchestrator/executor/executor_resume.go` ~line 1074)
reads `GetExecutorRunningBySessionID` and copies every key passing `ShouldPersistMetadataKey`, and
`env_secret_id_` is in `persistentMetadataPrefixes` (`executor_backend.go` ~line 288). It survives
`buildLaunchMetadata` unfiltered. Confirmed against the database: the four `executors_running` rows
belonging to SSH sessions all carry `env_secret_id_AGENTCTL_AUTH_TOKEN`. (Zero `task_sessions` rows
do, which is correct — `session.Metadata` is deliberately never merged into the workspace info.)

## Fix

Extract `resolveLaunchAuthToken` and give `manager_launch.go` the same per-executor split
`manager_execution.go` already had: Docker keeps the container-control token (including the existing
workspace-reuse fallback), every other executor uses its own session handshake token.

The Docker behaviour is unchanged — for a Docker executor the new call is exactly the old one, since
`allowLegacyContainerControlFallback` reduced to `WorkspaceReuseRequired` once the type check was
already satisfied.

## Tests

`TestResolveLaunchAuthToken` in `manager_auth_test.go`, five cases: SSH with a session secret returns
it; SSH with nothing anywhere returns empty *and* `requireSSHAgentctlAuthToken` still rejects it (so
the fix cannot silently launch unauthenticated); and three Docker cases pinning the container-control
precedence, the legacy fallback, and the no-reuse case.

Verified RED against the previous resolver — `ssh with a session auth secret returns that token`
fails without this change and passes with it. The four Docker cases pass either way, which is the
point: they are there to pin behaviour this PR must not move.

`go build ./...`, `golangci-lint run ./...` (0 issues), and
`go test ./internal/agent/runtime/... ./internal/backendapp/...` are clean. One unrelated
pre-existing failure, `TestBuildAuthMethodsIdentityAgentOverridesEnvironment`, reproduces
identically on an unmodified `main` checkout — it binds a unix socket under `t.TempDir()` and blows
macOS's ~104-byte `sun_path` limit. It is a macOS-only local-run papercut, not a regression here.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/3007?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/agent/runtime/lifecycle/manager_auth_test.go
+++ b/apps/backend/internal/agent/runtime/lifecycle/manager_auth_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/kandev/kandev/internal/common/logger"
 	"github.com/kandev/kandev/internal/secrets"
+	"github.com/kandev/kandev/internal/task/models"
 )
 
 // inMemorySecretStore implements secrets.SecretStore for testing.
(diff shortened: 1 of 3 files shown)


carlosflorencio:
Thanks for the review. I pushed 12a1a6207c, which reuses the shared Docker executor check and reports runtime secret reveal failures during non-Docker launch and SSH resume instead of masking them as a missing auth token. There were no unresolved review threads to resolve.

PR #3032, opened 2026-08-25
fix(auth): require admin for every install-wide plugin lifecycle route

jcfs:
> [!TIP]
> **PR walkthrough:** [Open the visual walkthrough](https://walkthrough.kandev.ai/pr/3032/1e618413941e.html)

Installing a plugin required the admin role, but every sibling install-wide lifecycle mutation did not, so with `features.auth` enabled any authenticated member could uninstall, disable or reconfigure an admin-installed plugin for the whole instance. This gates them all, and stops the settings UI from offering a member controls whose requests can only come back 403.

## Important Changes

- **`internal/plugins/handlers.go`** now uses an `admin := api.Group("", authn.RequireAdmin())` split, mirroring the `g`/`admin` convention in `internal/system/system.go`, instead of one inline guard on `/install`. On `admin`: install, sync, `PUT /settings`, `GET|PATCH /:id` config, `PUT /:id/auto-update`, uninstall, enable, disable, and the marketplace mutations (refresh + source add/update/delete).
- **Reads stay open** so the plugin UI still loads for members: `GET ""`, `GET /:id`, `GET /:id/bundle`, `GET /:id/ui/*path`, the marketplace listings, and `GET /settings` (the row list renders each plugin's effective auto-update state against it).
- **The three judgement calls, decided and commented next to the routes:**
  - `GET /:id/config` -> **admin**. It does not hand out *declared* secrets in cleartext (`GetMaskedConfig` runs `maskSecrets`), but masking is driven entirely by the plugin author's `config_schema`, so a manifest that omits `secret: true` returns a live credential verbatim; the non-secret half (orgs, endpoints, channel names) is operator-owned config regardless. Nothing member-facing renders it now that the config editor is admin-only.
  - `PUT /settings` and `POST /sync` -> **admin** (instance-wide; a sync starts new code on the host for everyone). `GET /settings` stays open because the installed list renders each row's effective auto-update state against that one boolean.
  - `PUT /:id/auto-update` -> **admin**; the override decides whether the host silently replaces a plugin's code with no operator in the loop.
- **Untouched by design**: plugin actions resolve the caller's identity against the declared selector, webhooks are gated by manifest visibility, and per-user plugin storage is per-user. A test pins that none of them was swept into the admin group.
- **Behavior with authentication disabled is unchanged.** `httpmw.SyntheticIdentity()` carries `RoleAdmin`, so `RequireAdmin` is a no-op there. The pre-existing mutating-handler tests now run against a router that resolves as that exact identity, so they keep asserting that mode verbatim.
- **`apps/web`** hides rather than fails: sync, check-for-updates, install, the Browse tab, enable/disable/uninstall/update and both auto-update switches are admin-only, the per-plugin config form and danger zone are replaced by an admin-only note, and the "Setup required" probe (an admin-only read) is skipped for members. The duplicated `useIsAdmin` in `settings-tree.tsx` moved to a shared `hooks/domains/auth/use-is-admin.ts`.

Rebased onto `c3b5408f1`; all of the below was re-run after the rebase.

- `go test ./internal/plugins/... ./internal/auth/... ./internal/system/...` and `make -C apps/backend lint` (0 issues).
- `cd apps/web && pnpm run typecheck`, `pnpm lint`, `pnpm run i18n:check`, `pnpm run i18n:ratchet`, and `pnpm test`.
- **Over-gating is asserted, not assumed.** `TestPluginReadRoutesStayOpenToMembers` requires a **200** (not merely "not 403") from a member for `GET /api/plugins`, `GET /:id`, `GET /:id/bundle`, `GET /:id/ui/*path`, `GET /settings` and `GET /marketplace/sources`; `TestMarketplaceCatalogStaysOpenToMembers` covers the catalog against a live index server. Breaking the plugin UI for non-admins is the worse regression here and it is what these pin.
- **Mutation-tested.** Moving `DELETE /:id` back off the admin group turned `TestInstallWideRoutesRejectMember/uninstall` red; moving `POST /:id/disable` back turned `TestEveryPluginRouteIsClassified` red; dropping `canManage={isAdmin}` in the settings page turned a frontend test red.
- `TestEveryPluginRouteIsClassified` walks the routes gin actually registered and fails on any route that is admin-gated without being listed, or listed without being gated, so a new route cannot be added to this package without someone classifying it.
- **Composition, not just per-route checks.** No plugin lifecycle path is in `httpmw.isPublicPath` or `isDeferredPath` (only the webhook relay defers, and it self-authenticates), so with auth enabled an anonymous caller is already stopped at the global middleware. `TestInstallWideRoutesRejectAnonymous` bypasses that middleware and requires each route to fail closed with 401 on its own, so the gate does not depend on something upstream having run.
- `apps/web/e2e/tests/auth/plugin-lifecycle-authorization.spec.ts` (auth project) drives the whole stack: an admin installs the real `plugin-fixture` package, a member gets 403 on eight mutations and 200 on list/get/bundle, the plugin stays `active`, and the settings surface shows the member the admin-only notice with none of the controls.

## Possible Improvements

Low risk. The likeliest regression is over-gating, which would break the plugin UI for every non-admin; the read routes and the member-visible plugin list are asserted explicitly at both the handler and e2e levels. Marketplace source management was gated alongside the rest of the lifecycle rather than left open, which is a small scope extension beyond the reported routes but the same install-wide argument.

## Checklist

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/3032?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jcfs:
Closing as a duplicate: #2816 (open since 19 Aug) already gates the same plugin lifecycle and marketplace-mutation routes, adds the same `apps/web/hooks/domains/auth/use-is-admin.ts`, removes the same local `useIsAdmin` from `settings-tree.tsx`, threads the same `canManage` prop, and ships an equivalent auth-project e2e spec. I raised this without checking for an in-flight PR on the same surface first.

One delta is worth carrying over to #2816 rather than losing:

**`GET /api/plugins/:id/config` is still ungated there.** #2816 hides the config *form* from members in the UI but leaves the route on the open group, so a member can still read a plugin's stored config directly. `GetMaskedConfig` masks only what the plugin author declared `secret: true` / `format: password` in `config_schema`, so a manifest that under-declares returns the value verbatim, and the non-secret half (orgs, endpoints, channel names) is operator config regardless.

Two smaller things from this branch, take them or leave them:

- `admin := api.Group("", authn.RequireAdmin())` instead of repeating the guard inline on each route, mirroring the `g`/`admin` split in `internal/system/system.go:183`.
- `TestEveryPluginRouteIsClassified`, which walks the routes gin actually registered and fails on any route that is admin-gated without being listed or listed without being gated, so a new route cannot be added to the package without someone classifying it.

Also flagging one behavioural disagreement for #2816 to decide deliberately: that PR hides the `plugin-settings` `PluginSlot` from members. This branch kept it visible, on the grounds that the slot renders the plugin's *own* UI, which may be per-user rather than operator config.

Branch left in place at `feature/gate-plugin-lifecycl-4cu` if any of it is worth cherry-picking.

jcfs:
Closing again at the repo owner's explicit instruction. Same reason as before: this duplicates #2816, which has been open since 19 Aug and already gates the same plugin lifecycle and marketplace-mutation routes, adds the identical `apps/web/hooks/domains/auth/use-is-admin.ts`, makes the same `settings-tree.tsx` deletion, and threads the same `canManage` prop.

To be explicit for anyone reading the history: this close is a deliberate `gh pr close`, not the side effect of a failed `gh pr merge`. Please do not reopen without checking with @jcfs first.

The one delta worth carrying to #2816 rather than losing with this PR: **`GET /api/plugins/:id/config` is still ungated there** — #2816 hides the config form from members in the UI but leaves the route on the open group, and `GetMaskedConfig` masks only fields the plugin author declared `secret: true` / `format: password`, so an under-declaring manifest returns the credential verbatim. Details and the two smaller optional deltas are in my earlier closing comment above.

Branch `feature/gate-plugin-lifecycl-4cu` (commit `1e61841`) is left in place if any of it is worth cherry-picking.

PR #3145, opened 2026-08-29
fix(agentctl): report guarded ACP denials and allow the guarded ACP prefix

yattdev:
## Problem

Session-handoff summarization fails. The reported error progressed through two forms:

```
Summarize failed / failed to execute prompt:
  command prefix "/usr/local/bin/kandev-agent-guard" is not an allowed ACP command
```

then, once the prefix was supplied differently:

```
ACP initialize failed: {"code":-32603,"message":"Internal error",
  "data":{"error":"peer disconnected before response"}}
```

The second form was hard to diagnose because the utility ACP executor **captured the guarded child's stderr and then discarded it** when building the returned error, so the guard's own rejection never reached the caller.

## Changes

**`d430887a6` — report guarded ACP denials** (+36/-1)

Adds `withGuardDenyDiagnostic`, which appends the guard's rejection line to the response error. It matches **only** lines prefixed `ERROR: kandev-agent-guard:` — the fixed prefix the guard's `deny()` helper emits before exiting — and deliberately does not forward arbitrary child stderr, which may contain provider output.

**`a6720afe1` — allow guarded ACP inference** (+47/-1)

Adds `resolveACPCommandPrefix` to validate the optional launcher wrapping an already allow-listed ACP agent command. The deployment guard is accepted **only at its exact absolute path** `/usr/local/bin/kandev-agent-guard`; a matching basename from another directory is rejected, since that would let a caller substitute an untrusted wrapper. Existing allow-listed agent commands retain their prior prefix behaviour.

## Verification

- Fork-aware lint base resolver (`scripts/resolve-go-lint-base --comparison-base`) resolves `4d8763e4` from a clean canonical-origin clone.
- `scripts/resolve-go-lint-base.test.sh` passes all fork-aware regression cases.
- The exact `go-lint` pre-commit hook command passes with **0 issues** against canonical base.
- Branch is **2 ahead / 0 behind** `main`; no rebase required.

Both commits carry tests: `acp_executor_test.go` and `acp_command_prefix_allowlist_test.go`.

## Notes for reviewers

Opened as a **draft**. The change touches a command allow-list, so it wants a maintainer's security read rather than an automatic merge. The fix is already live in a local reviewed overlay on the deployment host; this PR is about making it persistent in the normal image path.

Related board task: `63d60af8-d1b8-48ef-a7c8-043a4488dd7a` owns the fork-aware lint base resolution (#3074) that this branch is validated against.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/3145?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/agentctl/server/utility/acp_command_prefix_allowlist_test.go
+++ b/apps/backend/internal/agentctl/server/utility/acp_command_prefix_allowlist_test.go
@@ -1,0 +1,33 @@
+package utility
+
+import "testing"
+
+func TestResolveACPCommandPrefixAllowsFixedKandevGuard(t *testing.T) {
+	t.Parallel()
+
+	if got := resolveACPCommandPrefix(kandevAgentGuardPath); got != kandevAgentGuardPath {
+		t.Fatalf("resolveACPCommandPrefix(%q) = %q, want fixed guard path", kandevAgentGuardPath, got)
+	}
+}
+
+func TestResolveACPCommandPrefixRejectsGuardLookalikes(t *testing.T) {
+	t.Parallel()
+
+	for _, name := range []string{
+		"kandev-agent-guard",
+		"/tmp/kandev-agent-guard",
+		"/usr/local/bin/kandev-agent-guard-copy",
+	} {
+		if got := resolveACPCommandPrefix(name); got != "" {
+			t.Errorf("resolveACPCommandPrefix(%q) = %q, want empty", name, got)
+		}
+	}
+}
+
+func TestResolveACPCommandPrefixRetainsExistingAllowlist(t *testing.T) {
+	t.Parallel()
+
+	if got := resolveACPCommandPrefix("npx"); got != "npx" {
+		t.Fatalf("resolveACPCommandPrefix(npx) = %q, want npx", got)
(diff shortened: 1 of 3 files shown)


yattdev:
Withdrawing this draft after deployment-scope verification.

This change targets Kandev’s locally deployed `/usr/local/bin/kandev-agent-guard` wrapper and its fixed `ERROR: kandev-agent-guard:` stderr contract. That guard/security layer is deployment infrastructure rather than behavior shipped or owned by upstream `kdlbs/kandev`, so this PR should not merge upstream as written.

The underlying local ACP failures were addressed in the deployment integration (including local deployment commit `779fcef`). The branch and commits remain preserved. A generic upstream proposal for safe, configurable wrapper diagnostics can be considered separately if it avoids coupling to this deployment’s path and stderr contract.

PR #3379, opened 2026-09-04
fix(web): stop a stale session snapshot from reverting a model switch

JnManso:
> [!TIP]
> **PR walkthrough:** [Open the visual walkthrough](https://walkthrough.kandev.ai/pr/3379/24b18e8be705.html)

Switching a session's model could silently revert, because a page reload clears the client's hydration record while the cached session metadata still names the pre-switch model, so the convergence event confirming the user's own selection was overwritten with the old value. A payload reporting the model recorded in `activeModel` — only ever written by an explicit user pick — now counts as confirmation, so the session hydrates and the provider payload wins.

The user-visible effect was that the picker label flipped to the new model and snapped back within a frame, leaving the switch looking like it never happened until a prompt refreshed the session metadata. The trigger label renders from the model config option rather than the optimistic `activeModel`, so the optimistic update alone did not keep the new value on screen.

## Validation

- `pnpm exec vitest run lib/ws/handlers components/task/model-selector*.test.* components/model-config-selector.test.tsx` — 49 files / 515 tests passed.
- `pnpm exec vitest run lib/ws lib/state/slices/session lib/state/slices/session-runtime components/task components/model-config-selector.test.tsx` — 457 files / 4357 passed, 4 skipped.
- `pnpm run typecheck` — clean.
- `pnpm exec eslint` on the changed files — clean under the repo's `--max-warnings 0`.
- `pnpm exec playwright test --config e2e/playwright.config.ts tests/chat/model-selector-consecutive-switch.spec.ts` — 2 passed.
- Manually exercised in a dev instance: switch model, reload, switch again, reload again to confirm the selection persists.

The two new tests in `apps/web/lib/ws/handlers/session-models-user-selection.test.ts` are the regression guard: they fail on the unpatched handler because the store reverts to the persisted model. The Playwright spec passes both with and without the fix — mock-agent re-emits a matching `session_models` payload on subscribe, which hydrates the store before the click, so the race does not reproduce there. It is included as flow coverage for consecutive switching, which had none, not as a strict guard.

No screenshots: the diff changes WebSocket state handling and adds tests, with no new or altered UI markup, and the defect is a transient wrong state that a static capture cannot show.

## Possible Improvements

Low risk. The new signal is gated on `!unsettledStartup`, so the resume protection that keeps persisted runtime config authoritative during startup is unchanged; `activeModel` is written only by an explicit user selection and is reverted on RPC failure, so it cannot vouch for a model the user did not choose.

## Checklist

- [ ] If I do not have repository write access and this is a large architectural change, I discussed the direction in a linked issue before opening this PR.
- [ ] This PR contains one logical change; unrelated work is split into separate PRs.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/3379?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/web/components/task/model-selector-consecutive-switch.test.tsx
+++ b/apps/web/components/task/model-selector-consecutive-switch.test.tsx
@@ -1,0 +1,187 @@
+import { act, cleanup, fireEvent, render, screen } from "@testing-library/react";
+import { TooltipProvider } from "@kandev/ui/tooltip";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { ModelSelector } from "@/components/task/model-selector";
+import { setSessionConfigOption } from "@/lib/api/domains/session-api";
+
+const mocks = vi.hoisted(() => {
+  const SESSION_ID = "session-1";
+  const OPUS = "opus[1m]";
+  const listeners = new Set<() => void>();
+  const modelOptions = [
+    { value: "opus", name: "Opus" },
+    { value: "opus[1m]", name: "Opus 1M" },
+    { value: "sonnet", name: "Sonnet" },
+    { value: "claude-fable-5", name: "Fable" },
+    { value: "claude-fable-5[1m]", name: "Fable 1M" },
+  ];
+  const appState = {
+    activeModel: { bySessionId: {} as Record<string, string> },
+    sessionModels: {
+      bySessionId: {
+        [SESSION_ID]: {
+          currentModelId: OPUS,
+          models: modelOptions.map((option) => ({
+            modelId: option.value,
+            name: option.name,
+          })),
+          configOptions: [
+            {
+              type: "select",
+              id: "mode",
+              name: "Mode",
+              currentValue: "agent",
+              options: [{ value: "agent", name: "Agent" }],
+            },
+            {
+              type: "select",
+              id: "model",
+              name: "Model",
+              currentValue: OPUS,
+              category: "model",
+              options: modelOptions,
+            },
+            {
+              type: "select",
+              id: "effort",
+              name: "Effort",
+              currentValue: "high",
+              options: [
+                { value: "high", name: "High" },
+                { value: "medium", name: "Medium" },
+              ],
+            },
+            {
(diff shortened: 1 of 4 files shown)


JnManso:
Addressed in 39f8c4ac4.

The spec no longer budgets for the label. Each switch now arms a
`session.models_updated` wait scoped to the session and the target model
before clicking the option, and awaits it after, so the UI assertions run
with the default timeout:

```ts
const converged = ws.waitForEvent("session.models_updated", {
  where: (payload) =>
    payload.session_id === sessionId && payload.current_model_id === modelId,
});
await listbox.getByRole("option", { name }).click();
await converged;
```

The startup model is read from the backend with `expect.poll` instead of a
frame wait, because that notification can be published before the page
subscribes — the case `e2e/README.md` calls out for `createTaskWithAgent`
tasks. `watchWs(testPage)` is created before the first `goto()` and survives
the `page.reload()` in the second test.

Only the two `waitForChatIdle({ timeout: 30_000 })` calls remain, matching
the sibling specs in `tests/chat/`.

Verified locally: `2 passed` against both the fixed and the pre-fix bundle.
Worth stating plainly — the causal waits made the spec deterministic but did
not turn it into a guard for this regression: mock-agent re-emits a matching
`session_models` payload on subscribe, which hydrates the store before the
click, so the race does not reproduce there. The regression guard is
`lib/ws/handlers/session-models-user-selection.test.ts`, which fails on the
unpatched handler.

carlosflorencio:
Added focused coverage in commit 24b18e8. The handler test now proves that later non-matching provider data stays authoritative after user-selection hydration, and it covers the STARTING plus unsettled-payload guard. These tests protect the stale-snapshot fix without changing runtime behavior.

PR #3442, opened 2026-09-06
fix(backend): break approver-first tie when neither quorum seat sits at the current step

nova28:
> [!TIP]
> **PR walkthrough:** [Open the visual walkthrough](https://walkthrough.kandev.ai/pr/3442/1e3ddadaf823.html)

## Summary
- `ResolveParticipantRole` always defaulted to approver-wins when a caller held both reviewer and approver seats but neither seat sat at the task's current step (e.g. both seats cast while the task sat on Backlog). In a thin Office workspace this filed the Review verdict under `approver`, which the Review step's `reviewer` guard never counts, deadlocking the task at Review forever.
- Now, when neither seat sits at the current step, resolution prefers the role named by that step's own `wait_for_quorum` guard when exactly one role is named. Falls back to approver-wins unchanged when the step names both roles, names neither, or fails to load.
- Amended `AC-TASKS-QUORUM-RECORDING-001.4` and the system-design doc to cover this third case (both roles at the same step, but not the task's current step), which the prior spec didn't address.

## Test plan
- [x] `go test ./internal/workflow/engine/... -run TestResolveParticipantRole -v` — all pass, including new regression test `TestResolveParticipantRole_BothSeatsAtSameNonCurrentStep_GuardRoleWins` reproducing the thin-workspace deadlock and asserting the Review -> Approval transition fires.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/3442?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/workflow/engine/quorum.go
+++ b/apps/backend/internal/workflow/engine/quorum.go
@@ -904,10 +904,13 @@ func (e *Engine) evaluateGuardStateReadOnly(
 // participates, per AC-57's "exactly once, in the engine" mandate.
 // When the caller holds both seats, the seat whose StepID equals stepID
 // wins, since that is the seat the current step's guard actually reads
-// decisions against. If both seats sit at stepID, or neither does,
-// approver wins, matching the precedence in office/dashboard/decisions.go's
-// resolveDeciderRole (AC-4). Returns ErrParticipantNotFound when the
-// agent occupies neither seat.
+// decisions against. When neither seat sits at stepID, the role named by
+// stepID's own wait_for_quorum guard wins, so a decision is never filed
+// under a role no guard at that step is counting. Approver wins as the
+// final tiebreak — when the step names both roles, names neither, or the
+// step cannot be loaded — matching the precedence in
+// office/dashboard/decisions.go's resolveDeciderRole (AC-4). Returns
+// ErrParticipantNotFound when the agent occupies neither seat.
 func (e *Engine) ResolveParticipantRole(
 	ctx context.Context, taskID, stepID, agentProfileID string,
 ) (role, participantID string, err error) {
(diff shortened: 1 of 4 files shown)


nova28:
Both bot-review suggestions addressed in 4e3e94959:

1. **CodeRabbit** (quorum_role_test.go:204-205): corrected the `TestResolveParticipantRole_NeitherSeatAtStep_ApproverWins` comment — the step-preference early return applies to whichever seat (reviewer or approver) sits at the queried step, not only the approver seat. Verified against `ResolveParticipantRole`'s loop (quorum.go:927-944): the `seat.StepID == stepID` early return runs for either role in iteration order.

2. **Claude review** (same file, item 2): same fix as #1.

3. **Claude review** (item 1, missing test): added `TestResolveParticipantRole_BothSeatsAtSameNonCurrentStep_StepNamesBothRoles_ApproverWins`, covering AC-4's third fallback branch — the queried step names *both* guard roles, so the single-role tiebreak can't apply and resolution falls through to approver-wins.

`go test ./internal/workflow/engine/...`, `go vet`, `gofmt`, and `golangci-lint run --new-from-rev` all clean. No production code changed — test/comment only.

carlosflorencio:
Thanks for the contribution. We pushed a follow-up that aligns role resolution with automatic transition guards, covers seats created at different earlier steps, and adds focused regression tests. It also ignores approval-only guards and corrects the misleading test comment.

PR #3509, opened 2026-09-08
fix(office): reject empty slug on skill update

nova28:
> [!TIP]
> **PR walkthrough:** [Open the visual walkthrough](https://walkthrough.kandev.ai/pr/3509/e4d458f8b091.html)

fix(office): reject empty slug on skill update

PATCH /api/v1/skills/:id with {"slug":""} previously short-circuited
past the well-formedness check and persisted an empty slug, producing a
skill row that skillslug delivery silently drops. Empty is now treated
as not well-formed like any other invalid slug and rejected with 400;
omitting the field is still the only way to leave the slug unchanged.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/3509?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/agentctl/server/process/workspace_poll_mode_grace_test.go
+++ b/apps/backend/internal/agentctl/server/process/workspace_poll_mode_grace_test.go
@@ -187,6 +187,11 @@ func TestPollModeGrace_StopDisarmsTimer(t *testing.T) {
 // it to return, and prevent the callback from changing the mode afterward.
 func TestPollModeGrace_StopJoinsFinalScan(t *testing.T) {
 	wt := newGraceTestTracker(t, graceFiresQuickly)
+	// The test covers the grace callback's cancellation and wait-group
+	// ownership. Disable the independent polling loops so their real Git work
+	// cannot change poll mode while this lifecycle boundary is under test (the
+	// Windows race runner can otherwise report an unrelated paused transition).
+	wt.gitIndexPath = ""
 	finalScanStarted := make(chan struct{})
 	finalScanFinished := make(chan struct{})
 	wt.gitStatusObserver = func(ctx context.Context) (types.GitStatusUpdate, error) {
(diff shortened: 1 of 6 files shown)


carlosflorencio:
Addressed in commit a20ece63f: moved the empty-slug service regression into a focused test file and added handler coverage for another malformed slug. The Office skills package passes 113 tests. Thanks for the contribution. Fresh CI is now running on the pushed head.

nova28:
Thanks for the review and for tightening up the test coverage in a20ece63f — I've built on top of it (not replaced it) with two more commits that close a related gap:

**The gap:** `ValidateSkillUpdate`'s empty-slug rejection is correct for an explicit `{"slug":""}`, but a stored row can *already* have an empty slug through no fault of the update request — e.g. a row created outside `ValidateAndPrepareSkill` (the config-import path bypasses it; tracked separately). Before this addition, any *content-only* PATCH to such a row (no `slug` field at all) would hit the same `WellFormed("")` check and 400 — so an already-broken row became permanently un-editable through the normal edit flow, even for a request that never touched the slug.

**The fix (`ValidateSkillUpdate(ctx, skill, slugRequested bool)`):**
- Caller sends `slug` → unchanged from your version: strict `WellFormed` → `Normalize` → uniqueness check, `""` still 400s.
- Caller omits `slug` and the stored value is already `""` → self-heals to a free name-derived slug (`healEmptySlug`) so the request succeeds instead of failing on a field it never mentioned; if every candidate collides, the stored empty value is left as-is and the request still succeeds (never fails on slug grounds for a request that didn't ask about the slug).

Added `TestValidateSkillUpdate_HealsEmptyStoredSlugWhenNotRequested`, `TestValidateSkillUpdate_LeavesEmptyStoredSlugOnCollisionWhenNotRequested` (service_slug_update_test.go), and the matching handler-level tests. Full `internal/office/...` suite is green, `golangci-lint --new-from-rev` is clean, `gofmt -l` is clean.

Happy to drop this if you'd rather ship the simpler reject-only version and handle the stored-empty-slug case separately — let me know.

PR #3526, opened 2026-09-09
docs(mcp): note that agent_profile_id cannot set an Office task assignee

nova28:
> [!TIP]
> **PR walkthrough:** [Open the visual walkthrough](https://walkthrough.kandev.ai/pr/3526/420d2fa2b617.html)

**Today:** `create_task_kandev`'s `agent_profile_id` description doesn't say what it does and doesn't do. An agent caller reading it can reasonably (and wrongly) assume setting `agent_profile_id` on an Office task also makes that agent the task's assignee. It doesn't — and the MCP tool schema sets `additionalProperties: false`, so the handler's existing, correct rejection message ("assignee_agent_profile_id is office-only and cannot be set via create_task_kandev") is unreachable from MCP. There is no discovery path except the description itself.
**After this:** The description states plainly that `agent_profile_id` is the launch profile only, that it does not set an Office assignee, that setting one is `can_assign_tasks`-gated, and points at the real mechanism: the Office UI or `PATCH /api/v1/office/tasks/:id`.
**Who hits this:** Any MCP agent caller creating or reassigning Office-workspace tasks through `create_task_kandev` — which today has no way to discover the limitation short of trying it, getting silent non-assignment, and debugging why.
**Scope:** Standalone doc-string fix. No siblings.
**Not here:** Exposing `AssigneeAgentProfileID` via MCP is a deliberate, permission-gated design decision (#1315) — reversing it is a feature, not this chore. Also out of scope: the auto-start-on-unassigned-task path, which already fails visibly (`auto_start_failed` marker, landed in #2847) rather than silently.

`create_task_kandev`'s `agent_profile_id` sets the launch profile only and cannot set an Office task's assignee, but nothing told a caller that. This appends one clarifying sentence to the tool description, in both task-mode and external-mode variants, naming the correct mechanism instead.

## Validation

- `go test ./internal/mcp/... -count=1` — all packages pass, including the new `TestCreateTask_ExternalMode_AgentProfileDescNotesOfficeAssigneeLimitation`
- `gofmt -l internal/mcp/server/server.go internal/mcp/server/handlers_test.go` — clean
- `make lint` (backend + web) — 0 issues
- `make lint-format` — clean
- `make typecheck` — clean
- `cd apps/web && pnpm run i18n:ratchet` — clean (no UI source touched)
- `internal/worktree` test failures on this branch reproduce identically on the merge base (`18d38c11a`) in a separate scratch worktree — pre-existing sandbox/temp-dir environment issue on this runner, not caused by this diff
- `.github/scripts/lint-harness-files.py` fails under this runner's Python 3.9 (`str | None` needs 3.10+) — pre-existing environment mismatch, file untouched by this diff
- E2E not required: diff touches only `apps/backend/internal/mcp/server/{server.go,handlers_test.go}`, zero `apps/web/` paths

## Possible Improvements

Low risk: pure doc-string change, no behavior change. The two new/updated test assertions check the load-bearing claim (non-assignment) and mechanism (PATCH path) as substrings rather than every clause of the sentence — accepted as test-rigor-only per review.

## Checklist

- [x] If I do not have repository write access and this is a large architectural change, I discussed the direction in a linked issue before opening this PR.
- [x] This PR contains one logical change; unrelated work is split into separate PRs.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/3526?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


apps/backend/internal/mcp/server/server.go
```diff
@@ -1419,12 +1419,12 @@ func (s *Server) registerMRAutomationTools() {
...
+	agentProfileDesc := "Agent profile ID to use. On a workflow step, the step's launch profile (its pinned profile, or the workflow default when unpinned) outranks it; otherwise an explicit agent_profile_id wins. When both are absent, current_task uses the verified creating session's profile and effective model, mode, and dynamic options for a session-bound call, then falls back to the current/source or parent profile without verified session context. workspace_default skips those task profiles and the creating session, then uses workflow profiles before the target workspace default. Explicit profiles do not copy creator-session runtime values. start_agent=false still needs a resolvable profile for later manual start. agent_profile_id is the launch profile only and does not set an Office task's assignee; the Office assignee is can_assign_tasks-gated and set via the Office UI or PATCH /api/v1/office/tasks/:id."
```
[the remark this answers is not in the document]
nova28: Valid — confirmed `UpdateTaskRequest.hasAnyField()` (`internal/office/dashboard/handler.go:598`) checks `AssigneeAgentProfileID`, and the JSON tag on that field is `assignee_agent_profile_id` (not `agent_profile_id`), so a caller who copies the MCP tool's own parameter name into the PATCH body would get a 400. Fixed in 5d3452969: both descriptions now name the required field explicitly, and `TestCreateTask_ExternalMode_AgentProfileDescNotesOfficeAssigneeLimitation` asserts on it.

carlosflorencio:
I pushed a small follow-up in 420d2fa2b: the MCP descriptions now clarify that can_assign_tasks applies to agent callers on the Office PATCH path, and the tests cover assignee_agent_profile_id in both task and external modes.

apps/backend/internal/mcp/server/server.go
```diff
@@ -1419,12 +1419,12 @@ func (s *Server) registerMRAutomationTools() {
...
+	agentProfileDesc := "Agent profile ID to use. On a workflow step, the step's launch profile (its pinned profile, or the workflow default when unpinned) outranks it; otherwise an explicit agent_profile_id wins. When both are absent, current_task uses the verified creating session's profile and effective model, mode, and dynamic options for a session-bound call, then falls back to the current/source or parent profile without verified session context. workspace_default skips those task profiles and the creating session, then uses workflow profiles before the target workspace default. Explicit profiles do not copy creator-session runtime values. start_agent=false still needs a resolvable profile for later manual start. agent_profile_id is the launch profile only and does not set an Office task's assignee; set the Office assignee through the Office UI or PATCH /api/v1/office/tasks/:id with an assignee_agent_profile_id field in the request body. Agent callers need can_assign_tasks for this PATCH operation (agent_profile_id there is not recognized)."
```
carlosflorencio, line 1422:
```suggestion
	agentProfileDesc := "Agent profile ID to use. On a workflow step, the step's launch profile (its pinned profile, or the workflow default when unpinned) outranks it; otherwise an explicit agent_profile_id wins. When both are absent, current_task uses the verified creating session's profile and effective model, mode, and dynamic options for a session-bound call, then falls back to the current/source or parent profile without verified session context. workspace_default skips those task profiles and the creating session, then uses workflow profiles before the target workspace default. Explicit profiles do not copy creator-session runtime values. start_agent=false still needs a resolvable profile for later manual start."
```

We should not need to add the office info

> agent_profile_id is the launch profile only and does not set an Office task's assignee; set the Office assignee through the Office UI or PATCH /api/v1/office/tasks/:id with an assignee_agent_profile_id field in the request body. Agent callers need can_assign_tasks for this PATCH operation (agent_profile_id there is not recognized).

to the  create_task_kandev tool when its used in kanban workspaces, kanban workspaces shouldn't be allowed to create tasks in the office workspace for now.

carlosflorencio (changes requested):

apps/backend/internal/mcp/server/server.go
nova28: Checked — that gate doesn't exist yet. `handleCreateTask` (`internal/mcp/handlers/handlers.go`) takes `workspace_id` straight from the caller with no check that it matches the caller's own (kanban) workspace; nothing else in the create-task path restricts it either. So a kanban-scoped in-session agent can target the office workspace today — that's literally how the original bug this PR documents was found (task 241149c6, reproduced by creating a task in the live Office workspace via MCP from outside it).

Given that, I'd rather not strip the sentence: it's the only place a caller finds out `agent_profile_id` silently doesn't set the Office assignee, and removing it reopens that exact gap for everyone who can reach this path right now, not a hypothetical.

If you'd rather fix this at the source — gate `create_task_kandev` so a kanban-workspace caller can't target an office workspace's `workspace_id` at all — that's a real (if small) contract change, out of scope for this doc-only card. Happy to file it as a tracked follow-up if you want it pursued; just say the word and I'll open it referencing this thread. Leaving this open for you either way.

carlosflorencio (approved):

PR #3618, opened 2026-09-12
fix(task): stop plugin pending-interaction API from picking the wrong current turn

nova28:
> [!TIP]
> **PR walkthrough:** [Open the visual walkthrough](https://walkthrough.kandev.ai/pr/3618/5a9d0c519a33.html)

**Today:** A plugin calling the pending-interactions host API can get the wrong "current turn" for a session: a stale completed turn can outrank the actual open one, or a lifecycle-only turn (e.g. an automatic status update) can stand in for an unanswered question and hide it from the plugin.

**After this:** The pending-interactions query now excludes lifecycle-only turns and ranks an open turn ahead of any completed one, the same rule the other ten current-turn resolution sites in this package already follow. All eleven sites now agree on which turn is "current."

**Who hits this:** Any plugin reading pending interactions through `Host.ListPendingInteractions` (`pkg/pluginsdk`). This is a plugin host API surface, not the web UI.

**Scope:** standalone fix, no sibling PRs.

**Not here:** This does not move Threads' `needs_action` badges or the sidebar pending-action counts. Those already read `GetPendingActionsBySessionIDs`, which already used the shared `currentTurnAuthority` logic. Only the plugin-facing listing changes.

One of eleven current-turn resolution sites in `internal/task/repository/sqlite` hand-wrote its own predicate and `ORDER BY` instead of reusing the shared `currentTurnAuthority` helper, so it disagreed with the other ten on which turn counts as "current" for a session. `buildPendingInteractionQuery` now takes both the predicate and the ordering from `currentTurnAuthority`, closing the two points of divergence: lifecycle-only turns are excluded, and an open turn now outranks a newer completed one. `turnAuthorityPredicate` has exactly one non-test caller left, so all eleven sites are structurally guaranteed to agree.

## Validation

- `go test -count=1 -tags fts5 ./internal/task/repository/sqlite/...` — PASS (full package, 72.6s)
- Postgres suite actually run (dialect-sensitive raw SQL), verbatim:
  ```
  --- PASS: TestPostgresPendingInteractionsMatchSQLiteAuthority (0.20s)
  --- PASS: TestPostgresListPendingInteractionsSkipsLifecycleOnlyTurn (0.16s)
  --- PASS: TestPostgresListPendingInteractionsRanksOpenTurnAheadOfNewerCompletedTurn (0.16s)
  ok  github.com/kandev/kandev/internal/task/repository/sqlite  0.829s
  ```
- `gofmt -l` — clean
- `go vet ./...` — clean
- `make lint` — 0 issues
- `make lint-format` — clean
- `make typecheck` — clean
- `cd apps/web && pnpm run i18n:ratchet` — clean (diff touches no UI files)
- E2E not required: `git diff --name-only origin/main...HEAD` touches only `apps/backend/internal/task/repository/sqlite/`, zero `apps/web/` paths.

Two classes of pre-existing, unrelated test failures were hit while running the full suite and confirmed present at merge base `20efe4855` in a detached scratch worktree (not caused by this branch, which never touches these packages):
- `internal/worktree` (multiple tests): `unsafe worktree path ... not a directory` — environment-specific worktree/filesystem failures, reproduced identically at the merge base.
- `internal/task/service` (11 tests, e.g. `TestArchiveTaskCleanupPreservesTaskEnvironmentIdentity`, `TestTaskLifecycleCleanup_MissingWorktree`): same class of failure, reproduced identically at the merge base.
- `TestPostgresWorkspaceSourceParentGuard` fails identically at merge base `20efe4855` (`subquery in FROM must have an alias`, SQLSTATE 42601, in `workspace_folder_postgres_test.go`, a file this branch never touches).

## Possible Improvements

Low risk: a read-only query change scoped to one CTE's predicate/ordering, covered by dialect-matched tests on both SQLite and Postgres plus mutation testing (4/4 mutants killed across both dialects).

## Checklist

- [ ] If I do not have repository write access and this is a large architectural change, I discussed the direction in a linked issue before opening this PR.
- [ ] This PR contains one logical change; unrelated work is split into separate PRs.

<a href="https://cubic.dev/pr/kdlbs/kandev/pull/3618?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/internal/task/repository/sqlite/pending_interactions.go
+++ b/apps/backend/internal/task/repository/sqlite/pending_interactions.go
@@ -62,11 +62,13 @@ func buildPendingInteractionQuery(
 	pendingIDExpr := dialect.JSONExtract(driverName, "m.metadata", "pending_id")
 	scopedSessions, args := pendingInteractionSessionScope(filter)
 	kindConditions, kindArgs := pendingInteractionKindClauses(filter.Kinds)
+	predicate, orderBy := currentTurnAuthority(driverName, "turn_row")
 
 	query := fmt.Sprintf(
 		pendingInteractionQueryTemplate,
 		scopedSessions,
-		turnAuthorityPredicate(driverName, "turn_row"),
+		orderBy,
+		predicate,
 		nonTerminalSessionPredicate("turn_row"),
 		pendingIDExpr,
 		models.MessageTypeClarificationRequest,
@@ -90,10 +92,11 @@ func buildPendingInteractionQuery(
 }
 
 // pendingInteractionQueryTemplate is the format string behind
-// buildPendingInteractionQuery. current_turn duplicates the ranking in
-// pendingActionsBySessionQuery on purpose: both must agree, and a shared
-// builder would have to thread the scoping subquery's bind arguments through
-// the projection's placeholder list. The pinning test
+// buildPendingInteractionQuery. current_turn resolves its turn through
+// currentTurnAuthority, the same predicate and ordering every other
+// current-turn resolution site in this package uses, so this query and the
+// compact pending-action projection (GetPendingActionsBySessionIDs) always
+// agree on which turn is current. The pinning test
 // (TestListPendingInteractionsAgreesWithPendingActionProjection) fails if the
 // two ever disagree on a session.
 const pendingInteractionQueryTemplate = `
@@ -105,7 +108,7 @@ const pendingInteractionQueryTemplate = `
 				       id AS turn_id,
 				       ROW_NUMBER() OVER (
 				         PARTITION BY task_session_id
-				         ORDER BY started_at DESC, created_at DESC, id DESC
+				         ORDER BY %s
 				       ) AS rn
 				FROM task_session_turns turn_row
 				WHERE turn_row.task_session_id IN (SELECT id FROM scoped_sessions)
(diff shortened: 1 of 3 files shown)


nova28:
Thanks for the review. Re: "Format-string arg alignment is fragile at scale" (pending_interactions.go:67-85) — confirmed valid, but out of scope for this PR:

- The `fmt.Sprintf` template pattern with 16 positional `%s` placeholders predates this change; it was already present at merge base `20efe4855` (originally introduced in #2922). This PR only adds two more args to an existing template of the same shape, it doesn't introduce the pattern.
- The suggested mitigation (naming template holes via a struct or named format function) is, as you note, a larger refactor spanning the whole package's query-building convention, not a fix scoped to this defect.

Filing a follow-up card to track the refactor separately rather than growing this PR.

carlosflorencio:
Thanks for the contribution. I pushed 5a9d0c519 to replace the pending-interaction query’s positional SQL substitutions with named fields. This keeps the behavior unchanged while making future template edits safer and easier to review; the renderer regression and full SQLite repository test package pass.

